PR #6, opened 2020-07-06
Sketch for some command line tools in the toolbox.

kousu:
This demonstrates how to make a CLI that thinly wraps some library code in your package.

The CLIs I made are based on sections of the flowchart at https://www.shimming-toolbox.org/1_overview/2_flowchart/. I think it's likely that they will be soon/eventually deleted and replaced by scripts that perform a larger block of the MRI shimming process in one step -- calling the (currently empty) library functions in the process. But having a ready-made CLI is really good for experimentation on different datasets as you develop the library functions.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,8 +26,8 @@ setup(
     keywords="",
     entry_points={
         'console_scripts': [
-            "shim-referencemaps=shimmingtoolbox.cli.referencemaps:main",
-            "shim-b0maps=shimmingtoolbox.cli.b0maps:main"
+            "shim-referencemaps=shimmingtoolbox.cli.referencemap:main",
+            "shim-b0maps=shimmingtoolbox.cli.b0map:main"
         ]
     },
     packages=find_packages(exclude=["contrib", "docs", "tests"]),
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,7 @@ setup(
     keywords="",
     entry_points={
         'console_scripts': [
-            "shim-referencemaps=shimmingtoolbox.cli.referencemap:main",
+            "shim-referencemaps=shimmingtoolbox.cli.referencemaps:main",
             "shim-b0maps=shimmingtoolbox.cli.b0map:main"
         ]
     },
--- a/setup.py
+++ b/setup.py
@@ -24,6 +24,12 @@ setup(
     author="NeuroPoly Lab, Polytechnique Montreal",
     author_email="neuropoly@googlegroups.com",
     keywords="",
+    entry_points={
+        'console_scripts': [
+            "shim-referencemaps=shimmingtoolbox.cli.referencemaps:main",
+            "shim-b0maps=shimmingtoolbox.cli.b0maps:main"
+        ]
+    },
     packages=find_packages(exclude=["contrib", "docs", "tests"]),
     install_requires=install_reqs,
     package_dir={"shimmingtoolbox": "shimmingtoolbox"},
--- a/shimmingtoolbox/b0map.py
+++ b/shimmingtoolbox/b0map.py
@@ -1,0 +1,4 @@
+
+
+def b0map(*args, **kwargs):
+    raise NotImplementedError()
--- a/shimmingtoolbox/cli/b0map.py
+++ b/shimmingtoolbox/cli/b0map.py
@@ -1,0 +1,27 @@
+#!/usr/bin/env python3
+
+import logging
+
+import click
+
+from .. import b0map
+
+@click.command()
+@click.option("--verbose", is_flag=True, help="Be more verbose.")
+@click.option("--coilshims", help="Output hardware-shim calibration.")
+@click.option("--pulseseq", help="Output pulse-sequence calibration.")
+@click.argument("input", type=click.Path(exists=True, file_okay=False))
+@click.argument("output", type=click.Path(exists=False))
+def main(verbose, coilshims, pulseseq, input, output):
+    """
+    Compute b0 fieldmap OUTPUT from a folder of dicom files INPUT.
+    """
+    if verbose:
+        logging.getLogger().setLevel(logging.INFO)
+    logging.info(f'{coilshims}, {pulseseq}, {input}, {output}')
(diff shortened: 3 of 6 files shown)


setup.py
```diff
@@ -16,7 +16,7 @@
 
 setup(
     name="shimmingtoolbox",
-    python_requires=">=3.7",
+    python_requires=">=3.6",
```
kousu, line 19: I changed this to support Ubuntu Xenial (which is one of the machines I'm on). It's not strictly necessary; relying on 3.7+ has some benefits afterall.

setup.py
```diff
@@ -25,6 +25,12 @@
     author_email="neuropoly@googlegroups.com",
     keywords="",
     packages=find_packages(exclude=["contrib", "docs", "tests"]),
+    entry_points={
+        'console_scripts': [
+            "shim-referencemaps=shimmingtoolbox.cli.referencemaps:main",
+            "shim-b0maps=shimmingtoolbox.cli.b0maps:main"
```
kousu, line 31: Maybe the scripts should just be "referencemaps" and "b0maps"? :shrug: 

I wish `scripts=` worked still.

shimmingtoolbox/b0map.py
```diff
@@ -0,0 +1,4 @@
+
```
kousu, line 1: This should have a docstring
```suggestion
"""
Compute b0 maps from ...

See: arxiv.org/whatever, plosone.org/whatever, etc
```

setup.py
po09i: Any reason why `shim-` is added before the function name?

jcohenadad: we should probably settle on a "shimming-toolbox" prefix. How about `st_`? or do we want something more explicit? like `shimtool_`?
also: separation with _ or - ?

po09i: I like `st-` better because it is faster but I'm open to other opinions

jcohenadad: traditionally in the lab we go with snake_case for CLI commands, but i'm happy to reconsider "-" for word separators. it is definitely faster

kousu: It's just to keep the collection of related tools together. There isn't a namespace system for command line commands, so lots of projects that are made of small related tools do this.

Sticking with snake_case makes the python names and the unix names compatible, so my vote is `st_`.

jcohenadad: > Sticking with snake_case makes the python names and the unix names compatible, so my vote is st_

out of curiosity, why wouldn't `st-` be python and unix compatible?  is it because "-" could be interpreted as "minus" in python? (it shouldn't because we are not calling CLI names inside code, are we?)

po09i: Snake_case also works

gab-berest: I also like better the "_" by habits. Also a convention for C like programming languages like Python is to use the underscore for exactly the idea of minus symbol since minus operation doesn't need to have spaces before or after the symbol (From what I know)

shimmingtoolbox/cli/b0map.py
```diff
@@ -0,0 +1,27 @@
...
+
+from .. import b0map
+
+@click.command()
+@click.option("--verbose", is_flag=True, help="Be more verbose.")
+@click.option("--coilshims", help="Output hardware-shim calibration.")
+@click.option("--pulseseq", help="Output pulse-sequence calibration.")
+@click.argument("input", type=click.Path(exists=True, file_okay=False))
+@click.argument("output", type=click.Path(exists=False))
+def main(verbose, coilshims, pulseseq, input, output):
```
gab-berest, line 15: Shouldn't it be called another name since main is the normally called automatically. Shouldn't it be a specific function name and then called in the if __name == '__main__'?

shimmingtoolbox/cli/referencemaps.py
```diff
@@ -0,0 +1,27 @@
...
+@click.command()
+@click.option("--verbose", is_flag=True, help="Be more verbose.")
+@click.argument("inputs", required=True, nargs=-1, type=click.File('rb'))
+@click.argument("output", type=click.Path(exists=False))
+def main(verbose, inputs, output):
+    """
+    Compute a reference field b0 fieldmap OUTPUT from a folder of dicom files INPUT.
+    """
+    if verbose:
+        logging.getLogger().setLevel(logging.INFO)
```
gab-berest, line 18: should we decide the logger levels to use since there's a lot. ERROR and INFO are straight forward but like what would be a CRITICAL or a WARNING?

shimmingtoolbox/cli/b0map.py
kousu: @gab-berest main() is never called automatically. It's only called automatically if you write:

```
if __name__ == '__main__':
    main()
```

Aside, that unfortunately suffers under python3's [import rule](https://www.python.org/dev/peps/pep-0328/#relative-imports-and-name), so it's sort of fallen out of favour, being replaced by setup.py's `console_scripts`:

```
entry_points={
  'console_scripts':
    {'b0map': 'shimmingtoolbox.cli.b0map:main'}
}
```

but again that's not automatic, that's explicitly telling python to call it.

shimmingtoolbox/cli/referencemaps.py
kousu: My thinking about logging is that Golang is right: [log levels are unhelpful](https://dave.cheney.net/2015/11/05/lets-talk-about-logging).

This gives three levels of output: print()s, to show the normal course of output, verbose logging, to show details that might help a user debug why it's not working for them, and exceptions, which immediately exit with an error. Any more is unhelpfully splitting hairs, I think.

po09i (approved):
I looked at the code, as you mentioned most scripts will eventually be removed. This helps a lot to get started with CLIs so I think we can merge and base our future command line scripts to this template.

jcohenadad:
> I looked at the code, as you mentioned most scripts will eventually be removed. This helps a lot to get started with CLIs so I think we can merge and base our future command line scripts to this template.

maybe address the "easy" comments so they won't have to be done again in the future, e.g. https://github.com/shimming-toolbox/shimming-toolbox-py/pull/6/files#r450144812

jcohenadad:
good to merge @kousu ?

jcohenadad (approved):

PR #7, opened 2020-07-06
packaging: only install tests when asked

kousu:
This uses "extras_require" to only install tests for testing, and not for end-user usage.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,7 @@ with open(path.join(here, "README.md"), encoding="utf-8") as f:
 
 setup(
     name="shimmingtoolbox",
-    python_requires=">=3.7",
+    python_requires=">=3.6",
     version=shimmingtoolbox.__version__,
     description="Code for performing real-time shimming using external MRI shim coils",
     long_description=long_description,
--- a/setup.py
+++ b/setup.py
@@ -9,11 +9,6 @@ here = path.abspath(path.dirname(__file__))
 with open(path.join(here, "README.md"), encoding="utf-8") as f:
     long_description = f.read()
 
-req_path = path.join(here, "requirements.txt")
-with open(req_path, "r") as f:
-    install_reqs = f.read().strip()
-    install_reqs = install_reqs.split("\n")
-
 setup(
     name="shimmingtoolbox",
     python_requires=">=3.7",
@@ -25,6 +20,10 @@ setup(
     author_email="neuropoly@googlegroups.com",
     keywords="",
     packages=find_packages(exclude=["contrib", "docs", "tests"]),
-    install_requires=install_reqs,
-    package_dir={"shimmingtoolbox": "shimmingtoolbox"},
+    install_requires=[
+        #"numpy~=1.16.0",
+    ],
+    extras_require={
+        'testing': ["pytest~=4.6.3", "pytest-cov~=2.5.1"]
+    },
 )
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 .pytest_cache/
 .ipynb_checkpoints/
 .vscode/
+*.pyc
 
 # OS-related files/folders
 __MACOSX/
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ python:
 # command to install dependencies
 install:
   - echo "Installation of the repository"
-  - pip install -e .
+  - pip install -e .[testing]
   - pip install python-coveralls  # Add me to install python-coveralls
 # command to run tests
 script:
--- a/README.md
+++ b/README.md
@@ -2,3 +2,30 @@
 [![Build Status](https://travis-ci.com/shimming-toolbox/shimming-toolbox-py.svg?branch=master)](https://travis-ci.com/shimming-toolbox/shimming-toolbox-py)
 
 Code for performing real-time shimming using external MRI shim coils
+
+## Installation
+
+```
+$ pip install shimmingtoolbox@git+https://github.com/shimming-toolbox/shimming-toolbox-py
+```
+
+Depending on your system, you may need to prefix this command with `sudo` (for a global install),
+or write it with `install --user` (for a user-local install) in order to be allowed to install it.
+
+## Development
+
+To set up a development environment, make sure python is installed, then run
+
+```
+$ pip install -e ".[testing]"
+```
+
+This will add the project folder to your python path and install the necessary dependencies to run tests.
+
+To run tests manually, do
+
+```
+$ pytest
+```
+
+See https://docs.pytest.org/ for more options.
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,0 @@
-pytest==4.6.3
-pytest-cov==2.5.1
-


setup.py
```diff
@@ -9,14 +9,9 @@
...
 
-req_path = path.join(here, "requirements.txt")
-with open(req_path, "r") as f:
-    install_reqs = f.read().strip()
-    install_reqs = install_reqs.split("\n")
-
 setup(
     name="shimmingtoolbox",
-    python_requires=">=3.7",
+    python_requires=">=3.6",
```
mathieuboudreau, line 14: Have you folks internally discussed which python version you want to start developing this package for? In AxonDeepSeg we've recently just transitioned from Python 3.6 to 3.7, and if you recommend users to use conda virtual environments then it's really not an issue to move ahead in version number.

mathieuboudreau:
I tried the command `pip install -e .[testing]`, however got an error because I'm running on a ZSH shell (and not bash). FYI: All new Mac OS versions are using ZSH, so this'll be a recurring issue moving forward.

```bash
(st_7) mathieuboudreau@mathieu shimming-toolbox-py % pip install -e .[testing]
zsh: no matches found: .[testing]
```

The command required to work was `pip install -e ."[testing]"

mathieuboudreau (changes requested):
Personally for projects where most users are the developers (e.g. axondeepseg, and this project for the start), I prefer having the packages required for testing installed by default as 1) it encourage devs to write & run tests and 2) it's quicker to make users run the tests to debug when they encounter issues. However I'm not supper attached to it, all I'd recommend here is you at least write in the readme or wiki some instructions for developers vs users on how to install the software (i.e. `pip install -e ."[testing]"` vs `pip install -e `. Also see my two other comments. Once I get responses/confirmations I'll approve.

mathieuboudreau (approved):

setup.py
gab-berest: During the little formation, we tried to avoid conda and made them install 3.8, but I don't think we discussed it. It should be discussed since it will impact (a little bit) on portability. However, I know that 3.6 is used a lot in the Python community

kousu:
> > The other(?) blessed method is to use tox's `deps` option: [pypa/setuptools#931](https://github.com/pypa/setuptools/issues/931)
> 
> tox maintainer here. deps is not per se blessed for this. In the modern days should probably only be used to specify application dependencies, IMHO should be left empty for libraries shrug

I appreciate the attention, @gaborbernat. Thank you for dropping into our project in your spare time to help us out. It's hard to start a new package with the best practices in the process of shifting. It's sort of confusing because my original patch had

```
setup(
...
    tests_require=[
       "pytest",
    ],
...
```

but that warned me

```
$ python3 setup.py test
running test
WARNING: Testing via this command is deprecated and will be removed in a future version. Users looking for a generic test entry point independent of test runner are encouraged to use tox.
```

which led me to https://github.com/pypa/setuptools/pull/1878/files (which [added that](https://github.com/jdufresne/setuptools/blob/cd84510713ada48bf33d4efa749c2952e3fc1a49/setuptools/command/test.py) warning) which [led](https://github.com/pypa/setuptools/pull/1878#issuecomment-544987531) me [to](https://github.com/pypa/setuptools/issues/1684) find https://github.com/pypa/setuptools/issues/931, a long thread where pganssle and jaraco repeat multiple times that `tox` is the thing to use, that you can and should use it to ["do what `setup.py test` _should_ do"](https://github.com/pypa/setuptools/issues/931#issuecomment-508312175).

I concluded that `tox` is the blessed solution because it's recommended on the command line and by some of PyPA's core devs. Do you think it is always a mistake to use `tox` as a shorthand for handling test-only dependencies? Are your PyPA team-members confused?

But [this comment](https://github.com/pypa/setuptools/issues/1684#issuecomment-648441338) from jaraco also recommends the `extras_requires`, which seems to work for me (except as @mathieuboudreau points out, that method has a stumbling block on the most recent macOS :face_with_head_bandage: ).

kousu:
> I tried the command `pip install -e .[testing]`, however got an error because I'm running on a ZSH shell (and not bash). FYI: All new Mac OS versions are using ZSH, so this'll be a recurring issue moving forward.
> 
> ```shell
> (st_7) mathieuboudreau@mathieu shimming-toolbox-py % pip install -e .[testing]
> zsh: no matches found: .[testing]
> ```
> 
> The command required to work was `pip install -e ."[testing]"

That's an unfortunate surprise. I admit that gives me pause since so many people are on macOS. I wish computers would just get along with each other! I still want to do this because tests are optional dependencies and this is how python knows to express them.

setup.py
kousu: I am against developing for conda on the principle that it's a very specialized environment. Developing for stock cpython will make code that's portable to conda, but I don't think the reverse would be true, and I don't like building deep dependency trees. It's also not that hard to cover 3.6..3.9 in one codebase, there's no major syntax changes since 3.6, and we have [Travis](https://docs.travis-ci.com/user/build-matrix/) and/or [tox](https://tox.readthedocs.io/) to handle validating multiple versions.

But to be clear, I only bumped it backwards for my own convenience, because of the machine I happened to be working on had python3=python3.6.  We can scratch this change.

kousu:
I added some documentation @mathieuboudreau, how does it look to you? If you approve a second time, would you merge it?

mathieuboudreau:
@kousu the documentation looks great! I would actually recommend against suggesting the user use the `install --user` flag though - the behaviour of this flag is not always straightforward, and has caused me a gigantic headache in the past. (If I recall, anything installed there will override what you try to install in your active virtual environment). @jcohenadad have you encountered this behaviour too in the past? 

@kousu I know you said that you didn't cover venvs with the new devs yet, but I'd recommend using venvs over the user installation to get around sudo permissions. We have venv instructions in AxonDeepSeg that even completely inexperienced users were able to follow without any issue, and it works without needing sudo anywhere: 

* Installation of Miniconda: https://axondeepseg.readthedocs.io/en/latest/documentation.html#miniconda
* Creating and activate a conda venv for a specific Python version: https://axondeepseg.readthedocs.io/en/latest/documentation.html#virtual-environment

jcohenadad:
> @kousu the documentation looks great! Only minor point is that I would actually recommend against suggesting the user use the install --user flag though - the behaviour of this flag is not always straightforward, and has caused me a gigantic headache in the past. (If I recall, anything installed there will override what you try to install in your active virtual environment). @jcohenadad have you encountered this behaviour too in the past?

agreed about the headaches-- i just ran into this issue (again) 2h ago 😅 

also agreed about offering an "installation scenario" for minimally-experienced users. This is one of the  strength of miniconda.

setup.py
```diff
@@ -25,6 +20,10 @@
     author_email="neuropoly@googlegroups.com",
     keywords="",
     packages=find_packages(exclude=["contrib", "docs", "tests"]),
-    install_requires=install_reqs,
-    package_dir={"shimmingtoolbox": "shimmingtoolbox"},
+    install_requires=[
+        #"numpy~=1.16.0",
```
kousu, line 24: To be uncommented as soon as needed. I left this in as an example to get us started.

setup.py
gab-berest: I also think that working with Python 3.6 isn't a problem for the same reasons. I would keep it that way.

mathieuboudreau: Ok re:conda, it was just a suggestion! We've had better experience in AxonDeepSeg managing user experiences by requiring conda venvs, so I was just sharing our experience 🙂

mathieuboudreau: But as for making the code compatible for the 3.6 to 3.9 range - you should modify the Travis config file to test on all those versions if that's yoyr goal, otherwise some bugs might slip for users running Python 3.6 if some devs are using features only available for higher python versions.

mathieuboudreau:
The recent discussions, while important, I think are deviating from the original aim of this PR. So I'll merge and we can continue discussing in new issues if needed.

PR #9, opened 2020-07-06
Add a shimming scenario script under /examples

### LINKED ISSUE #8: Add the different shimming scenarios as scripts under /examples
## Context 
The `shimming-toolbox` is designed to be used in different shimming scenarios (Slice-wise or Volume-wise, realtime or dynamic, with or without respiratory probe, with or without RF shimming...) .

## Goal
We want to provide scripts corresponding to the different scenarios with the call to the right functions/methods in order to make our workflow easier to understand and adapt to other specific applications.

## How to do this
We plan to start by adding non-functional scripts that contain the different steps of each scenario and then implement the corresponding functions bits-by-bits to progressively build a functional toolbox.

gaspardcereza:
## Context
The 'shimming-toolbox' is designed to be used in different shimming scenarios (Slice-wise or Volume-wise, realtime or dynamic, with or without respiratory probe, with or without RF shimming...). 

These scenarios must describe all the different steps and functions required in each specific case. That will provide us a general overview of the project, help us organize our work and make the workflow easier to understand and to adapt. 

## Goal
Here we want to provide a [fully commented script](https://github.com/shimming-toolbox/shimming-toolbox-py/blob/gc/scenarios/examples/volume_RT_shim_respiratory.py) containing all the  different steps of a scenario. That file will allow us to keep track of the different functions we need to create/update to complete that scenario. 

Once these different functions will be implemented, we will be able to just uncomment them in the file (thus turning it into a functional script). This will be covered in subsequent pull requests.

Fixes #8

### DIFF
--- a/examples/volume_RT_shim_respiratory.py
+++ b/examples/volume_RT_shim_respiratory.py
@@ -1,18 +1,4 @@
-# Context: this is blablabla
-# Things to consider:
-# - with this large refactoring, we should try to accommodate Jason's current code
-# - log file should be created for each action.
-#   - should we use Matlab's diary()?
-#   - if so, is there a way to configure an output log file at the beginning of a process, and then pass the log handler across functions?
-# - when useful, plottings, etc. will appear.
-# - conventions:
-#   - variable: mixedCamelCase
-#     - images: start with "img", example: imgFieldmapMag
-#   - functions: snake_convention
-#     - use "action", then "what"
-# - use nifti-based software
-#   - ANTs: great for registration, interpolation (fast!)
-#
+# Context: That scripts covers a volume-wise realtime shimming with respiratory probe scenario 
 #
 # Sort unorganized images from DICOM socket transfer
 # dcm2bids is used to create folder structure and nifti conversion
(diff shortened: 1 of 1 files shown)


jcohenadad:
I would suggest to start with Eva’s script, as this is less cumbersome (don’t have to deal with setup of external custom probe and no need to deal with real-time)

po09i:
That makes sense. We can add it as another example. The steps up to B0 mapping should be the same for most example scripts so that should be the starting point.

examples/volume_RT_shim_respiratory.py
```diff
@@ -1,18 +1,4 @@
...
-# - when useful, plottings, etc. will appear.
-# - conventions:
-#   - variable: mixedCamelCase
-#     - images: start with "img", example: imgFieldmapMag
-#   - functions: snake_convention
-#     - use "action", then "what"
-# - use nifti-based software
-#   - ANTs: great for registration, interpolation (fast!)
-#
+# Context: That scripts covers a volume-wise realtime shimming with respiratory probe scenario
```
jcohenadad, line 1: good!

examples/volume_RT_shim_respiratory.py
```diff
@@ -27,32 +13,32 @@
 # > phase = load_niftis( niftiPath ) 5D array (x, y, z, nEcho, nAcq)
```
jcohenadad, line 13: > `Using niftiPath will prompt the user to select the appropriate acquisition. `

this is inconsistent with the variable niftiPath introduced above

jcohenadad, line 13: i would comment the "5D array (x, y, z, nEcho, nAcq)", not putting it at the same line as the command

examples/volume_RT_shim_respiratory.py
```diff
@@ -27,32 +13,32 @@
...
 #
 # Convert to complex
-# complexArray = mag.*exp( 1i.*phase ); 5D array (x, y, z, nEcho, nAcq)
+# > complexArray = mag.*exp( 1i.*phase ); 5D array (x, y, z, nEcho, nAcq)
```
jcohenadad, line 16: what's "1i"?

examples/volume_RT_shim_respiratory.py
```diff
@@ -0,0 +1,56 @@
...
+#
+# Sort unorganized images from DICOM socket transfer
+# dcm2bids is used to create folder structure and nifti conversion
+# unsortedDicomDir: path to the folder containing the dicoms
+# niftiPath: path where we want dcm2bids to store the nifti files
+# 
+# > dicom_to_nifti( unsortedDicomDir, niftiPath )
+#
+# Load images from the path
+# Using niftiPath will prompt the user to select the appropriate acquisition. Alternatively, if the path to the acquisition is known, the path can be changed and the function won't prompt the user
```
jcohenadad, line 11: > `Using niftiPath will prompt the user to select the appropriate acquisition. `

this is inconsistent with the variable niftiPath introduced above

examples/volume_RT_shim_respiratory.py
```diff
@@ -0,0 +1,56 @@
...
+# > mag = load_niftis( niftiPath ) 5D array (x, y, z, nEcho, nAcq)
+# > phase = load_niftis( niftiPath ) 5D array (x, y, z, nEcho, nAcq)
+#
+# Convert to complex
+# > complexArray = mag.*exp( 1i.*phase ); 5D array (x, y, z, nEcho, nAcq)
+#
+# Compute B0 fieldmap
+# b0FieldMaps could be 2d, 3d or 4d. 4th dimension is the time in [s] (i.e. one B0 per timestamp).
```
jcohenadad, line 19: update to camel_case

examples/volume_RT_shim_respiratory.py
gaspardcereza: When creating complex on Matlab, it is better to use '1i' rather than 'i' to avoid confusion with eventual variables called 'i'. I don't know how we will do this in python though.

examples/volume_RT_shim_respiratory.py
gaspardcereza: You mean snake_case right ?

jcohenadad: yes sorry

examples/volume_RT_shim_respiratory.py
gaspardcereza: I don't think it is because we defined niftiPath as the path were the files are stored (so it corresponds to the folder). When we use that folder as input, the user must precise which acquisition corresponds to the magnitudes or the phases. But if the direct path to the acquisition is given, the user's intervention won't be needed. @po09i can you confirm ?

jcohenadad: clarification is needed here: what is meant by "path", what is meant by "acquisition"?
example of folder structure should easily clarify

po09i: Yes this is how it works @gaspardcereza, maybe add some clarification as Julien suggested to make things clearer

examples/volume_RT_shim_respiratory.py
gab-berest: In python is is j that is import from the cmath package. Also, in Python element wise operation doesn't exist. You'll have to use a separate function, a list comprehension or (easier) numpy.

jcohenadad (approved):

PR #15, opened 2020-07-07
Add unwrap phase functionality

### LINKED ISSUE #10: Add unwrap phase functionality
Convert previous implementation of unwrap phase from shimming-toolbox to this python repo.

### LINKED ISSUE #17: Implement prelude wrapper
Allow prelude to be easily called from python. Based off this [shimming-toolbox file](https://github.com/shimming-toolbox/shimming-toolbox/blob/master/misc/prelude.m).

### LINKED ISSUE #23: Bundle prelude as a dependency to run on Travis
To test prelude on Travis, prelude needs to be bundled as a dependency so that Travis can run prelude.

po09i:
This PR is to set up the initial unwrapping structure as well as an unwrapping algorithm (prelude). 

The main purposes of this PRs:
- Add example script: examples/unwrap_script.py
- Implemented one of the uwrapping algorithm: prelude: shimmingtoolbox/unwrap/prelude.py (wrapper)
- Implement tests

Fixes #10 
Fixes #17 
Fixes #23


shimmingtoolbox/unwrap/__init__.py
```diff
@@ -0,0 +1,3 @@
+__version__ = "0.1"
```
jcohenadad, line 1: we only need to mention the version in the main `shimmingtoolbox/__init__.py`, right? otherwise there is duplication

examples/unwrap_script.py
```diff
@@ -0,0 +1,57 @@
...
+import nibabel as nib
+import numpy as np
+import matplotlib.pyplot as plt
+import cmath
+
+
+def main():
+
+    # Import phase
+    phasePath = ""
```
jcohenadad, line 18: we use snake_case for variables, as per [our conventions](https://github.com/shimming-toolbox/shimming-toolbox-py/wiki)

shimmingtoolbox/unwrap/prelude.py
```diff
@@ -0,0 +1,74 @@
...
+    phaseNii = nib.Nifti1Image(np.angle(complexArray[:-1, :-1, 0, 0]), affine)
+    magNii = nib.Nifti1Image(np.abs(complexArray[:-1, :-1, 0, 0]), affine)
+    nib.save(phaseNii, os.path.join(dataSaveDirectory, "rawPhase.nii"))
+    nib.save(magNii, os.path.join(dataSaveDirectory, "mag.nii"))
+
+    # Fill options
+    options = " "
+
+    if isUnwrappingIn2D:
+        options = options + "-s "
```
gab-berest, line 42: You could use the += operator in Python. It is optimized for this kind of operation.

shimmingtoolbox/unwrap/prelude.py
```diff
@@ -0,0 +1,74 @@
...
+    if not mask.all() == -1:
+        assert mask.shape == complexArray.shape, "Mask must be the same shape as the array"
+        maskNii = nib.Nifti1Image(mask, affine)
+
+        options = options + "-m "
+        options = options + os.path.join(dataSaveDirectory, "mask.nii")
+        nib.save(maskNii, os.path.join(dataSaveDirectory, "mask.nii"))
+
+    # Unwrap
+    unwrapCommand = "prelude -p " + os.path.join(dataSaveDirectory, "rawPhase") + " -a " + \
```
gab-berest, line 54: For this type of concatenation, I prefer to use the .format() function. It helps to see what are the types that you need to put in your string (like %d for integer or %f for float) and control how it is presented to make sure that no new version of Python can implicitly change how your string is formatted.

examples/unwrap_script.py
```diff
@@ -0,0 +1,79 @@
...
+from shimmingtoolbox.unwrap import unwrap_phase
+
+
+def main():
+
+    # Download example data
+    url = 'https://github.com/shimming-toolbox/data-testing/archive/r20200709.zip'
+    filename = 'data-testing.zip'
+    # TODO: replace with download function
+    os.system('curl -o {} -L {}'.format(filename, url))
```
gab-berest, line 26: For a replace download function, there is the url.request.urlopen(url) function that will do that.

examples/unwrap_script.py
```diff
@@ -0,0 +1,84 @@
...
+    with ZipFile(filename, 'r') as zipObj:
+        # Extract all the contents of zip file in current directory
+        zipObj.extractall()
+    os.remove(filename)
+    # TODO: use systematic name for data-testing (could be in metadata of shimmingtoolbox
+    path_data = glob.glob('data-test*')[0]
+
+    # Open phase data
+    fname_phases = glob.glob(os.path.join(path_data, 'sub-fieldmap', 'fmap', '*phase*.nii.gz'))
+    nii_phase_e1 = nib.load(fname_phases[0])
```
gab-berest, line 37: When hardcoding indexes in an array, try to validate the length or number to check if it correctly took 2 spaces. If there's one missing, it will raise an error so it should be fine, but if it parses wrongly and creates 3 wrong fname_phases, you might have weird behaviors.

```suggestion
    if len(fname_phases) > 2:
        raise IndexError('Phase data parsing is wrongly parsed or something like this')
    nii_phase_e1 = nib.load(fname_phases[0])
```

examples/unwrap_script.py
```diff
@@ -0,0 +1,84 @@
...
+    nii_mag_e2 = nib.load(fname_phases[1])
+
+    # TODO: Convert to a function b0_map
+    # phasediff: (matlab code for 2 echoes)
+    # Z1(:,:,:) = mag_data(:,:,:,1).*exp(1i*ph_data(:,:,:,1));
+    # Z2(:,:,:) = mag_data(:,:,:,2).*exp(1i*ph_data(:,:,:,2));
+    # atan2(imag(Z1(:,:,:).*conj(Z2(:,:,:))),real(Z1(:,:,:).*conj(Z2(:,:,:))));
+    # Convert to radians (Assumes there are wraps)
+    phase_diff = - np.min(phase_diff) + phase_diff
+    phase_diff = phase_diff / 4096 * 2 * np.pi - np.pi
```
gab-berest, line 55: There's the built-in and optimised numpy.interp funciton in numpy that does this directly:
```suggestion
    MAX_PHASE = 4096
    MIN_PHASE = 0
    phase_diff = np.interp(phase_diff, [MIN_PHASE, MAX_PHASE], [-np.pi, np.pi] )
```

Like this, it scales between -np.pi and np.pi, assigning MIN_PHASE to -np.pi and MAX_PHASE to np.pi. I'm not sure why you substract the minimum before hand so I'll let you rearrange the function with what you need.

examples/unwrap_script.py
```diff
@@ -0,0 +1,84 @@
...
+    plt.subplot(1, 2, 1)
+    plt.imshow(np.angle(complex_array[:-1, :-1, 0]))
+    plt.title("Wrapped")
+    plt.colorbar()
+    plt.subplot(1, 2, 2)
+    plt.imshow(unwrapped_phase[:-1, :-1, 0])
+    plt.title("Unwrapped")
+    plt.colorbar()
+    # plt.show()
+    plt.savefig("myplot.png")
```
gab-berest, line 80: Maybe give it a more significant name? You could eventually name them unwrap_phase_plot_ii.png and check the last ii that was written and increment it by one. Or better, suffix the date of the creation of the plot.

examples/unwrap_script.py
```diff
@@ -0,0 +1,84 @@
...
+    phase_diff = - np.min(phase_diff) + phase_diff
+    phase_diff = phase_diff / 4096 * 2 * np.pi - np.pi
+
+    # TODO: create mask (probably in script)
+    # Call SCT or user defined mask
+    # mask = np.ones(phase_diff.shape)
+
+    complex_array = nii_mag_e1.get_fdata() * np.exp(1j * phase_diff)
+    affine = nii_phase_e1.affine
+    unwrap_function = "prelude"
```
gab-berest, line 63: Why save prelude in a variable and not just call it directly in the function?

shimmingtoolbox/unwrap/prelude.py
```diff
@@ -0,0 +1,81 @@
...
+    # Fill options
+    options = " "
+
+    if is_unwrapping_in_2d:
+        options += "-s "
+
+    # Add mask data and options if there is a mask provided
+    if not np.any(mask == -1):
+        # TODO: Make sure values are either 1 or 0
+        assert mask.shape == complex_array.shape, "Mask must be the same shape as the array"
```
gab-berest, line 55: `assert` is mostly for debugging. You should try to raise instead or find a way to work arround this problem instead.

shimmingtoolbox/unwrap/prelude.py
```diff
@@ -0,0 +1,81 @@
...
+
+        options += "-m "
+        options += os.path.join(data_save_directory, "mask.nii")
+        nib.save(mask_nii, os.path.join(data_save_directory, "mask.nii"))
+
+    # Unwrap
+    unwrap_command = "prelude -p {} -a {} -o {} {}".format(os.path.join(data_save_directory, "rawPhase"),
+                                                           os.path.join(data_save_directory, "mag"),
+                                                           path_2_unwrapped_phase, options)
+    os.system(unwrap_command)
```
gab-berest, line 66: As @kousu mentioned, you should try to avoid using os.system command:
```suggestion
    subprocess.run(['prelude', '-p ',os.path.join(data_save_directory, "rawPhase") , '-a', os.path.join(data_save_directory, "mag"), '-o',path_2_unwrapped_phase, options ], check=True)
```
Sorry for going against what I said previously.

shimmingtoolbox/unwrap/unwrap_phase.py
```diff
@@ -0,0 +1,53 @@
...
+
+    Returns:
+        numpy.ndarray: 3D array with the shape of `complex_array` of the unwrapped phase output from prelude
+
+    Raises:
+        Exception: complex array does not have 2,3,4,5 dimensions
+        Exception: unwrapping algorithm in `unwrap_function` is not supported
+    """
+    unwrapped_phase = np.empty(complex_array.shape, np.complex)
+    if unwrap_function == "prelude":
```
gab-berest, line 32: If this part is meant to have multiple unwrap_function, a better practice would be to create separate functions for each unwrap_function and save them in a dictionary with the unwrap_function as a key. Then in the function unwrap_phase you would just need to call it with the correct arguments: https://github.com/shimming-toolbox/shimming-toolbox-py/pull/18#discussion_r453138872
```suggestion
unwrap_functions = {
    'prelude' = unwrap_prelude,
    'other unwrap' = unwrap_other,
}

unwrapped_phase = unwrap_functions[unwrap_function](complex_array, affine, mask)
```
Also the dimension checks should be in the prelude function instead of here, the user function should not be aware of the different ways to give arguments. It should only give the inputs necessary and if the format is not adequate, then return an exception. Also, if someone wants only to use the prelude function for XY reason, he would have to redo all these checks and it results in duplicated code.

gab-berest: If we add another unwrapper then, we would have to rewritte these checks for the other unwrapper function. So another solution is to reshape the complexe_array beforehand and pass it to the function and adapt the function so it can take any of these shapes. I'll think of how later on.

examples/unwrap_script.py
po09i: Used request.get since `urllib.request.urlopen(url)` threw an error, "certificate error"

examples/unwrap_script.py
po09i: Not sure why but using `phase_diff = np.interp(phase_diff, [0, 4096], [-np.pi, np.pi])` does not work, as I already have it working another way i'll leave it be.

shimmingtoolbox/unwrap/unwrap_phase.py
po09i: From what I remember, previous discussions settled on having unwrappers that have as inputs (x,y,z) and dealing with echoes and acqs in the parent function. II'm open to revisit the question though.
[Comment](https://github.com/shimming-toolbox/shimming-toolbox/pull/131#discussion_r443025788) I'm referring to

examples/unwrap_script.py
jcohenadad: no problem on my end:
~~~
Python 3.7.3 (default, Mar 27 2019, 16:54:48) 
>>> import urllib.request
>>> urllib.request.urlopen(url)
<http.client.HTTPResponse object at 0x102242b70>
~~~
@po09i could you output the full syntax + output you got?

po09i: Input:
```
import urllib.request
url = 'https://github.com/shimming-toolbox/data-testing/archive/r20200709.zip'
urllib.request.urlopen(url)
```
Output:
```
Traceback (most recent call last):
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.7/lib/python3.7/urllib/request.py", line 1317, in do_open
    encode_chunked=req.has_header('Transfer-encoding'))
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.7/lib/python3.7/http/client.py", line 1229, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.7/lib/python3.7/http/client.py", line 1275, in _send_request
    self.endheaders(body, encode_chunked=encode_chunked)
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.7/lib/python3.7/http/client.py", line 1224, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.7/lib/python3.7/http/client.py", line 1016, in _send_output
    self.send(msg)
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.7/lib/python3.7/http/client.py", line 956, in send
    self.connect()
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.7/lib/python3.7/http/client.py", line 1392, in connect
    server_hostname=server_hostname)
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.7/lib/python3.7/ssl.py", line 412, in wrap_socket
    session=session
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.7/lib/python3.7/ssl.py", line 853, in _create
    self.do_handshake()
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.7/lib/python3.7/ssl.py", line 1117, in do_handshake
    self._sslobj.do_handshake()
ssl.SSLCertVerificationError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1056)
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/Users/alex/Library/Python/3.7/lib/python/site-packages/IPython/core/interactiveshell.py", line 3331, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-20-e5f25a21e2a4>", line 3, in <module>
    urllib.request.urlopen(url)
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.7/lib/python3.7/urllib/request.py", line 222, in urlopen
    return opener.open(url, data, timeout)
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.7/lib/python3.7/urllib/request.py", line 525, in open
    response = self._open(req, data)
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.7/lib/python3.7/urllib/request.py", line 543, in _open
    '_open', req)
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.7/lib/python3.7/urllib/request.py", line 503, in _call_chain
    result = func(*args)
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.7/lib/python3.7/urllib/request.py", line 1360, in https_open
    context=self._context, check_hostname=self._check_hostname)
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.7/lib/python3.7/urllib/request.py", line 1319, in do_open
    raise URLError(err)
urllib.error.URLError: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1056)>
```

jcohenadad: hum, weird, what's your output of
~~~
>>> urllib.request.__version__
'3.7'
~~~

po09i: I have the same output : '3.7'. I looked a bit into it on google and most of the errors refer to post install scripts that need to be run on OSX to activate certificates or quick fixes that deactivate the certificate check but that does not seem very secure.

po09i:
@jcohenadad Not sure how to set up prelude to work with travis, I looked at axondeepseg but they seem to download fsleyes and conda to do so which seems overkill for prelude only. If I only take the executable, I'm not sure how to link travis to the executable either.

jcohenadad:
> @jcohenadad Not sure how to set up prelude to work with travis, I looked at axondeepseg but they seem to download fsleyes and conda to do so which seems overkill for prelude only. If I only take the executable, I'm not sure how to link travis to the executable either.

this one is a bit tricky, i guess we would need to bundle the dependencies to run them on Travis-- could you open an issue for this?

po09i:
> this one is a bit tricky, i guess we would need to bundle the dependencies to run them on Travis-- could you open an issue for this?

https://github.com/shimming-toolbox/shimming-toolbox-py/issues/23

shimmingtoolbox/unwrap/unwrap_phase.py
```diff
@@ -0,0 +1,53 @@
...
+import numpy as np
+from shimmingtoolbox.unwrap.prelude import prelude
+
+
+def unwrap_phase(complex_array, affine, unwrap_function="prelude", mask=np.array([-1])):
+    """ Calls different unwrapping algorithms according to the specified `unwrap_function` parameter
+
+    Args:
+        complex_array (numpy.ndarray): 2D,3D,4D or 5D complex values numpy array to perform phase unwrapping
+        affine (numpy.ndarray): 2D array (4x4) containing the transformation coefficients. Can be acquired by :
```
jcohenadad, line 17: looks strange to ask for affine transfo (supposed to be for prelude, but unclear here)

shimmingtoolbox/unwrap/unwrap_phase.py
```diff
@@ -0,0 +1,53 @@
...
+            nii = nib.load("nii_path")
+            affine = nii.affine
+        unwrap_function (string, optional): unwrapper algorithm name. Possible values : prelude
+        mask: numpy array of booleans with shape of `complex_array` to mask during phase unwrapping. Calling `mask` wth
```
jcohenadad, line 21: needs to be tested here

shimmingtoolbox/unwrap/unwrap_phase.py
```diff
@@ -0,0 +1,53 @@
...
+
+import numpy as np
+from shimmingtoolbox.unwrap.prelude import prelude
+
+
+def unwrap_phase(complex_array, affine, unwrap_function="prelude", mask=np.array([-1])):
+    """ Calls different unwrapping algorithms according to the specified `unwrap_function` parameter
+
+    Args:
+        complex_array (numpy.ndarray): 2D,3D,4D or 5D complex values numpy array to perform phase unwrapping
```
jcohenadad, line 16: i suggest to input mag and phase as 2 separate arguments. we could make mag optional

shimmingtoolbox/unwrap/unwrap_phase.py
```diff
@@ -0,0 +1,53 @@
...
+""" Wrapper to different unwrapping algorithms
+
+"""
+
+
+import numpy as np
+from shimmingtoolbox.unwrap.prelude import prelude
+
+
+def unwrap_phase(complex_array, affine, unwrap_function="prelude", mask=np.array([-1])):
```
jcohenadad, line 12: at this point, i would not have a wrapper, but instead call `unwrap.prelude()` directly. why?
- it is easier to deal about dimensions requirements on a function-by-function basis
- add other checks (type of input: mag/phase vs. complex, etc.)

if we do that, there is not much that the wrapper does, so i would remove it for simplicity

shimmingtoolbox/unwrap/prelude.py
```diff
@@ -0,0 +1,82 @@
...
+
+"""
+
+import numpy as np
+import os
+import nibabel as nib
+import subprocess
+
+
+def prelude(complex_array, affine, mask=np.array([-1]), path_2_unwrapped_phase='./unwrapped_phase.nii',
```
jcohenadad, line 14: i would input phase and optionally mag, and have the input dims as they are expected by the actual FSL prelude.

if we need to loop across time or echoes, we can use python decorators

shimmingtoolbox/unwrap/unwrap_phase.py
gab-berest: I see. Then, we could still do the dimension check here, but still, my first comment on the function caller could be easier to avoid having other ifs for the different call to unwrappers ('prelude', etc.). [https://github.com/shimming-toolbox/shimming-toolbox-py/pull/15/files/d50e95663d36e378ffd12b30c2532a6e87ee303b#r454710944](url)

examples/unwrap_script.py
gab-berest: Yes, an error on Python seems to cause post install Python scripts to sometime not install correctly. This is an issue with Mac and not the program itself.

examples/unwrap_script.py
gab-berest: Strange on my Linux and my Windows it works. I'll investigate.

gab-berest: What does it give you?

shimmingtoolbox/unwrap/prelude.py
gab-berest, line 14: Normally, verbs like is are not in variable or argument names. They are reserved for function names (significance of action) and should not be used in these.

shimmingtoolbox/unwrap/prelude.py
```diff
@@ -0,0 +1,82 @@
...
+    subprocess.run(unwrap_command, shell=True, check=True)
+
+    # Uncompress
+    subprocess.run(['gunzip', path_2_unwrapped_phase + '.gz', '-df'], check=True)
+    unwrapped_phase = nib.load(path_2_unwrapped_phase)
+    unwrapped_phase = unwrapped_phase.get_fdata()
+
+    # Delete temporary files according to options
+    if not is_saving_nii:
+        os.remove(os.path.join(data_save_directory, 'mag.nii'))
```
gab-berest, line 76: When removing files, you have to be carefull. If the file is used by another process, in Windows, it will crash (you'd have to manage this error), but in Linux, it will delete it without freeing the memory allocated until the file is closed (wich could result in memory loss if not well coordinated). This is why it is important to think about the memory allocated during your program. A simple variable, will be deleted at the end by the garbage collector. However, dynamic entities like pointers or file descriptor can leak memory. A way to remove these ambiguities is to check if the file is opened in another process and refuse to delete if it is (with a warning) or wait to close it.
psutil can help you to manage these sort of mistakes. Something like this for one file (I'm not 100% sure how it is, but should be something like this):
```suggestion
        for proc in psutil.process_iter():
             flist = proc.open_files()
             if os.path.join(data_save_directory, 'mag.nii') not in flist:
                     os.remove(os.path.join(data_save_directory, 'mag.nii'))
```

It begins to be a deep understanding of programming language and is a good practice for packages and programs to not leak memory. However, it is your desicion to choose if you want to implement it or not.

test/test_prelude.py
```diff
@@ -0,0 +1,90 @@
...
+import glob
+import nibabel as nib
+import numpy as np
+
+import pytest
+
+from shimmingtoolbox.unwrap import prelude
+
+
+class TestCore(object):
```
gab-berest, line 17: Just to be sure, this only test the prelude funciton and not the wrapper and dimension checks that you have done previously. Would it be posisble to test these to?

test/test_prelude.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def setup(self):
```
gab-berest, line 19: A good practive for unit tests is to have every test be completly independent. Then, you have to make sure that your setup and teardown is called before each test that you will have in this class (if you have more than one)

shimmingtoolbox/unwrap/unwrap_phase.py
po09i: unwrap_phase function removed for now

shimmingtoolbox/unwrap/prelude.py
po09i: Not sure I understand your comment.

test/test_prelude.py
po09i: This is only for prelude, we are removing the unwrap_phase for now as we only have one unwrapper so we only need to test for prelude.

shimmingtoolbox/unwrap/unwrap_phase.py
po09i: unwrap_phase is deprecated, prelude will be run directly from the script so we don't need to bother.

shimmingtoolbox/unwrap/unwrap_phase.py
po09i: Removed unwrap_phase function, prelude is called directly now in the script and the check is done in prelude.

shimmingtoolbox/unwrap/prelude.py
```diff
@@ -0,0 +1,92 @@
...
+
+        options += '-m '
+        options += os.path.join(data_save_directory, 'mask.nii')
+        nib.save(mask_nii, os.path.join(data_save_directory, 'mask.nii'))
+
+    # Unwrap
+    unwrap_command = 'prelude -p {} -a {} -o {} {}'.format(os.path.join(data_save_directory, 'rawPhase'),
+                                                           os.path.join(data_save_directory, 'mag'),
+                                                           path_2_unwrapped_phase, options)
+    subprocess.run(unwrap_command, shell=True, check=True)
```
gab-berest, line 77: Subprocess is normally donne so you can pass him a list of options instead of formating a string by hand:
```suggestion
    subprocess.run(["prelude","-p",os.path.join(data_save_directory, "rawPhase"), os.path.join(data_save_directory, "mag"), path_2_unwrapped_phase, "-m",os.path.join(data_save_directory, "mask.nii")], shell=True, check=True)
```

jcohenadad: we could also introduce the following API (for convenient printout and more intuitive cmd input):
````python
def run_subprocess(cmd):
    """
    Wrapper for subprocess.run() that enables to input cmd as a full string (easier for debugging).
    :param cmd:
    :return:
    """
    print("{}".format(cmd))
    subprocess.run(cmd.split(' '), stdout=subprocess.PIPE, text=True)
````

Comment (jcohenadad) [reply shimmingtoolbox/unwrap/prelude.py:77]: of course print() would be replaced by logging in this project

Comment (po09i) [reply shimmingtoolbox/unwrap/prelude.py:77]: A problem occurs when a mask isn't provided and if the option "is_unwrapping_in_2d" is provided. Since both arguments are optional, then `ifs` are required and 4 different calls to prelude would be necessary to accomplish the same thing. ("options" can't be by itself because there are spaces in the string). Do you have a way to deal with that @gab-berest ?

Comment (po09i) [reply shimmingtoolbox/unwrap/prelude.py:77]: I guess Julien's function could help.

File: examples/unwrap_script.py [hunk 22]
```diff
@@ -0,0 +1,84 @@
...
+
+    nii_mag_e1 = nib.load(fname_mags[0])
+    nii_mag_e2 = nib.load(fname_mags[1])
+
+    # TODO: Convert to a function b0_map
+    # phasediff: (matlab code for 2 echoes)
+    # Z1(:,:,:) = mag_data(:,:,:,1).*exp(1i*ph_data(:,:,:,1));
+    # Z2(:,:,:) = mag_data(:,:,:,2).*exp(1i*ph_data(:,:,:,2));
+    # atan2(imag(Z1(:,:,:).*conj(Z2(:,:,:))),real(Z1(:,:,:).*conj(Z2(:,:,:))));
+    # Convert to radians (Assumes there are wraps)
```
jcohenadad, line 53: remove comment and create issue

examples/unwrap_script.py
```diff
@@ -0,0 +1,84 @@
...
+    phase_e1 = nii_phase_e1.get_fdata() / 4096 * 2 * np.pi - np.pi
+    phase_e2 = nii_phase_e2.get_fdata() / 4096 * 2 * np.pi - np.pi
+
+    # Open mag data
+    fname_mags = glob.glob(os.path.join(path_data, 'sub-fieldmap', 'fmap', '*magnitude*.nii.gz'))
+
+    if len(fname_mags) > 2:
+        raise IndexError('Mag data parsing is wrongly parsed')
+
+    nii_mag_e1 = nib.load(fname_mags[0])
```
jcohenadad, line 45: i would remove this test (we are in an example script-- so things are assumed the way they are: file names, number of files, etc.)

shimmingtoolbox/unwrap/prelude.py
```diff
@@ -0,0 +1,92 @@
...
+        is_unwrapping_in_2d (bool, optional): prelude parameter to unwrap in 2d
+        is_saving_nii (bool, optional): specify whether `wrapped_phase`, `mag`, `affine`, `mask` and `unwrapped_phase`
+                                        nii files will be saved
+
+    Returns:
+        numpy.ndarray: 3D array with the shape of `complex_array` of the unwrapped phase output from prelude
+    """
+    # Get absolute path
+    abs_path = os.path.abspath(path_2_unwrapped_phase)
+    data_save_directory = os.path.dirname(abs_path)
```
jcohenadad, line 39: please follow conventions: https://github.com/neuropoly/spinalcordtoolbox/wiki/variables#variables

shimmingtoolbox/unwrap/prelude.py
```diff
@@ -0,0 +1,92 @@
...
+
+    # Make sure directory exists, if not create it
+    if not os.path.exists(data_save_directory):
+        print('\nCreating directory for unwrapped phase at {}'.format(data_save_directory))
```
jcohenadad, line 43: use logging in API

shimmingtoolbox/unwrap/prelude.py
```diff
@@ -0,0 +1,92 @@
...
+# -*- coding: utf-8 -*-
+"""
+Wrapper to FSL Prelude (https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/FUGUE/Guide#PRELUDE_.28phase_unwrapping.29)
+"""
+
+import numpy as np
+import os
+import nibabel as nib
+import subprocess
+
```
jcohenadad, line 11: use logging

shimmingtoolbox/unwrap/prelude.py
```diff
@@ -0,0 +1,92 @@
...
+
+def prelude(wrapped_phase, mag, affine, mask=np.array([-1]), path_2_unwrapped_phase='./unwrapped_phase.nii',
+            is_unwrapping_in_2d=True, is_saving_nii=False):
```
jcohenadad, line 14: i would not input `is_unwrapping_in_2d` but check input data dim: if 3d --> use 3d option.

shimmingtoolbox/unwrap/prelude.py
```diff
@@ -0,0 +1,92 @@
...
+    abs_path = os.path.abspath(path_2_unwrapped_phase)
+    data_save_directory = os.path.dirname(abs_path)
+
+    # Make sure directory exists, if not create it
+    if not os.path.exists(data_save_directory):
+        print('\nCreating directory for unwrapped phase at {}'.format(data_save_directory))
+        os.mkdir(data_save_directory)
+
+    # Make sure phase and mag are the right shape
+    if wrapped_phase.ndim != 3:
```
jcohenadad, line 47: we should be able to use prelude in 2d mode

shimmingtoolbox/unwrap/prelude.py
jcohenadad, line 14: i would remove `is_saving_nii ` and check if logger is in DEBUG mode. If it is, save those outputs.

shimmingtoolbox/unwrap/prelude.py
```diff
@@ -0,0 +1,74 @@
...
+    Returns:
+        numpy.ndarray: 3D array with the shape of `complex_array` of the unwrapped phase output from prelude
+    """
+    # Make sure phase and mag are the right shape
+    if wrapped_phase.ndim != 3:
+        raise RuntimeError('wrapped_phase must be 3d')
+    if wrapped_phase.shape != mag.shape:
+        raise RuntimeError('The magnitude image (mag) must be the same shape as wrapped_phase')
+
+    tmp = tempfile.TemporaryDirectory(prefix='st_'+pathlib.Path(__file__).stem)
```
gab-berest, line 43: I know that temporary variables are common in programming, but you should still give them meenigfull names in relation with their function.

shimmingtoolbox/unwrap/prelude.py
gab-berest: >  both arguments are optional,

I would suggest tosave options in a list instead so you can add them accrodingly for their presence:
```suggestion
    options = []
    if is_unwrapping_2d:
        options.append("-s")
    ....

    subprocess.run(unwrap_command, shell=True, check=True)
```
In the end, you'll only need to pass the list of string options.

gab-berest: Also, Julien's function could be a good wrapper to normalize the use of processes through the code.

test/test_prelude.py
```diff
@@ -0,0 +1,182 @@
...
+        # Get the directory where this current file is saved
+        self.tmp_path = self.test_path / '__tmp__'
+        if not self.tmp_path.exists():
+            self.tmp_path.mkdir()
+        print(self.tmp_path)
+        self.toolbox_path = self.test_path.parent
+
+        self.get_phases_mags_affines()
+
+    def teardown(self):
```
gab-berest, line 33: We could check to integrate the unittest library. It helps to create unit test and includes classes, tear down functions, mock classes and setup functions. It's another library to learn, but it is quite powerfull.

jcohenadad: i'd rather stick with pytest, just because the rest of the lab is using it (when new students come in, they get help from other students)

shimmingtoolbox/unwrap/prelude.py
po09i: The section of code calling print has been removed. No need to use logging in this case

shimmingtoolbox/unwrap/prelude.py
po09i: https://github.com/shimming-toolbox/shimming-toolbox-py/pull/15#discussion_r461301641

shimmingtoolbox/unwrap/prelude.py
po09i: I'll remove the feature entirely, if the niis are required, they can directly be saved in the parent function. Using DEBUG requires resaving the niis since they are in a temporary directory.

shimmingtoolbox/unwrap/prelude.py
po09i: `tmp` is our [convention](https://github.com/neuropoly/spinalcordtoolbox/wiki/variables#variables). unfortunatly tempfile uses `temp`. `path_temp` is also explicit and the name of the function TemporaryDirectory is also very explicit. I think this is fine.

shimmingtoolbox/unwrap/prelude.py
po09i: Defaulted to False, it is significantly slower when not using a mask however. Also added ability to input 2d.

shimmingtoolbox/unwrap/prelude.py
po09i: Added the ability to do so.

shimmingtoolbox/unwrap/prelude.py
jcohenadad: > I'll remove the feature entirely, if the niis are required, they can directly be saved in the parent function. Using DEBUG requires resaving the niis since they are in a temporary directory.

and in fact, the niftis are _already_ saved, so no need to re-save them. The temp folders usually stay there for several hours (or weeks, depending on the OS config), which is enough for debugging

shimmingtoolbox/unwrap/prelude.py
```diff
@@ -0,0 +1,82 @@
...
+        mask (numpy.ndarray, optional): numpy array of booleans with shape of `complex_array` to mask during phase
+                                        unwrapping
+        threshold: Threshold value for automatic mask generation (Use either mask or threshold, not both)
+        is_unwrapping_in_2d (bool, optional): prelude parameter to unwrap slice by slice
+
+    Returns:
+        numpy.ndarray: 3D array with the shape of `complex_array` of the unwrapped phase output from prelude
+    """
+    # Make sure phase and mag are the right shape
+    if not((wrapped_phase.ndim == 2) or (wrapped_phase.ndim == 3)):
```
jcohenadad, line 38: shorter version:
```suggestion
    if not wrapped_phase.ndim in [2, 3]:
```

shimmingtoolbox/unwrap/prelude.py
```diff
@@ -0,0 +1,82 @@
...
+        raise RuntimeError('wrapped_phase must be 2 or 3d')
```
jcohenadad, line 39:
```suggestion
        raise RuntimeError('wrapped_phase must be 2d or 3d')
```

shimmingtoolbox/unwrap/prelude.py
```diff
@@ -0,0 +1,82 @@
...
+            logging.warning('Specifying both a mask and a threshold is not recommended, results might not be what is '
+                            'expected')
+
+    # Unwrap
+    unwrap_command = 'prelude -p {} -a {} -o {} {}'.format(os.path.join(path_tmp, 'rawPhase'),
+                                                           os.path.join(path_tmp, 'mag'),
+                                                           os.path.join(path_tmp, 'rawPhase_unwrapped'), options)
+    logging.info('Unwrap with prelude')
+    logging.debug('prelude command: %s', unwrap_command)
+    subprocess.run(unwrap_command, shell=True, check=True)
```
jcohenadad, line 79: with my previous suggestion of creating a `run_subprocess()` routine, you could add the logging.debug() inside the routine, avoiding this additional line of code every time you run subprocess.

shimmingtoolbox/unwrap/prelude.py
po09i: Implemented run_subprocess() function

test/test_prelude.py
gab-berest: Of course pytest is also an option. I was just saying this because I didn't see the required import.

test/test_prelude.py
gab-berest: The actual name to run a setup and a teardown before each test method is `setup_method` and `teardown_method` for pytest

examples/unwrap_script.py
```diff
@@ -0,0 +1,80 @@
...
+
+def main():
+
+    logging.basicConfig(level='INFO')
+
+    # Download example data
+    run_subprocess('st_download_data testing_data')
+
+    # Open phase data
+    path_data = glob.glob(os.path.join('.', 'testing_data*'))[0]
```
jcohenadad, line 30: in the future it could be replaced with `__dir_testing_data__`

examples/unwrap_script.py
```diff
@@ -0,0 +1,80 @@
...
+    # TODO: create mask
+    # Call SCT or user defined mask
+    # mask = np.ones(phase_e1.shape)
+
+    # Call prelude to unwrap the phase
+    unwrapped_phase_e1 = prelude(phase_e1, nii_mag_e1.get_fdata(), nii_phase_e1.affine)
+    unwrapped_phase_e2 = prelude(phase_e2, nii_mag_e2.get_fdata(), nii_phase_e2.affine, threshold=200)
+
+    # Plot results
+    plt.figure(figsize=(10, 10))
```
jcohenadad, line 57: please use OO of matplotlib to avoid crashing for headless systems. Context: https://github.com/neuropoly/spinalcordtoolbox/issues/1784#issuecomment-392424622

useful info: https://github.com/neuropoly/spinalcordtoolbox/wiki/Moving-away-from-pyplot

shimmingtoolbox/unwrap/prelude.py
```diff
@@ -0,0 +1,83 @@
...
+                                        unwrapping
+        threshold: Threshold value for automatic mask generation (Use either mask or threshold, not both)
+        is_unwrapping_in_2d (bool, optional): prelude parameter to unwrap slice by slice
+
+    Returns:
+        numpy.ndarray: 3D array with the shape of `complex_array` of the unwrapped phase output from prelude
+    """
+    # Make sure phase and mag are the right shape
+    if wrapped_phase.ndim not in [2, 3]:
+        raise RuntimeError('wrapped_phase must be 2d or 3d')
```
jcohenadad, line 40: convention " vs. ' --> https://github.com/neuropoly/spinalcordtoolbox/wiki/Single-or-double-quotes
```suggestion
        raise RuntimeError("Wrapped_phase must be 2d or 3d")
```

shimmingtoolbox/unwrap/unwrap_phase.py
```diff
@@ -0,0 +1,53 @@
+# #!/usr/bin/python3
```
jcohenadad, line 1: remove, and open issue with title "add other unwrapper" (or something like that), and include this code as potentially helpful to solve the issue.

jcohenadad (approved):

PR #5, opened 2020-07-04
Create numerical model data for multi-echo field maps

mathieuboudreau:
Resolves #1 

Completed conversion of the code & tests from the MATLAB implementation to Python.

Left to-do
* [x] Document & cleanup the code
* [x] Write up a demo in a Jupyter Notebook & add a Binder button.

PR #19, opened 2020-07-14
Source package version from setup.py.

kousu:
This avoids a tricky chicken-or-egg problem:
as soon as `shimmingtoolbox/__init__.py` starts loading up
submodules, which are going to load up libraries,
installation will fail because setup.py was doing

> import shimmingtoolbox

at the top, just to get at

> shimmingtoolbox.__version__

and of course when setup.py is run, none of the
libraries we need will be installed yet.

This is mentioned on https://packaging.python.org/guides/single-sourcing-package-version/

> Warning
>
> Although this technique is common, beware that it will fail if sample/__init__.py
> imports packages from install_requires dependencies, which will very likely not
> be installed yet when setup.py is run.

and after looking through that page, I decided that using importlib.metadata
is the future-proofed option.

(Editorializing: I really wish PyPA would stick to the Zen of Python's admonishment
to "There should be one obvious way to do it."; but the first thing this page says is
"There are many techniques" and the best is buried as #5)

### DIFF
--- a/shimmingtoolbox/__init__.py
+++ b/shimmingtoolbox/__init__.py
@@ -1,1 +1,9 @@
-__version__ = "0.1"
+
+# https://packaging.python.org/guides/single-sourcing-package-version/
+try:
+    from importlib import metadata
+except ImportError:
+    # Running on pre-3.8 Python; use importlib-metadata package
+    import importlib_metadata as metadata
+__version__ = metadata.version(__name__)
+del metadata
--- a/setup.py
+++ b/setup.py
@@ -1,8 +1,6 @@
 from setuptools import setup, find_packages
 from os import path
 
-import shimmingtoolbox
-
 # Get the directory where this current file is saved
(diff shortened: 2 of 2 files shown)


shimmingtoolbox/__init__.py
```diff
@@ -1 +1,9 @@
-__version__ = "0.1"
+
+# https://packaging.python.org/guides/single-sourcing-package-version/
+try:
+    from importlib import metadata
+except ImportError:
+    # Running on pre-3.8 Python; use importlib-metadata package
+    import importlib_metadata as metadata
+__version__ = metadata.version(__name__)
```
jcohenadad, line 8: wow, that's cool. First time i see this strategy

po09i (approved):

gaspardcereza (approved):

jcohenadad:
good to merge? @kousu

PR #28, opened 2020-07-20
Install dcm2niix in CI

### LINKED ISSUE #27: Install dcm2niix for testing suite
Testing suite will required [dcm2niix](https://github.com/rordenlab/dcm2niix), which is a dependency of [dcm2bids](https://github.com/cbedetti/Dcm2Bids).

According to the doc, there are various ways to install it, including:
- conda: `conda install -c conda-forge dcm2niix`  (for Linux, MacOS or Windows).
- [pre-built packages](https://github.com/rordenlab/dcm2niix/releases)

Question: which one should we go for? i vote for conda

kousu:
Fixes #27

This upgrades the CI environment from Ubuntu 16.04 LTS (which is Very Old now) to 18.04 (which is Current Boring Mainline), because 16.04 is too old to have dcm2niix available, but 18.04 does and is probably closer to what most people are running anyway!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+os: linux
+dist: bionic
 language: python
 python:
   - "3.7"
@@ -5,6 +7,8 @@ python:
 # command to install dependencies
 install:
   - echo "Installation of the repository"
+  - sudo apt-get update
+  - sudo apt-get install -y dcm2niix
   - pip install -e .[testing]
   - pip install python-coveralls  # Add me to install python-coveralls
 # command to run tests
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ install:
   - sudo apt-get update
   - sudo apt-get install -y dcm2niix
   - pip install -e .[testing]
-  - pip install python-coveralls  # Add me to install python-coveralls
+  - pip install python-coveralls
 # command to run tests
 script:
   - echo "Launch general integrity test"


kousu:
Tagging both Julien and Alex; but you can sort out which of you reviews this first.

jcohenadad (approved):
awesome! much more elegant approach than what i suggested. when we will be testing other OSs we might want to find an alternative solution

kousu:
Yeah. Well. In my opinion conda counts as a separate OS, just like Java or docker does. Except there's always some [lhttps://www.joelonsoftware.com/2002/11/11/the-law-of-leaky-abstractions/eakiness](https://www.joelonsoftware.com/2002/11/11/the-law-of-leaky-abstractions/) from the underlying platform, so I would rather just adapt to that platform directly than trying to paper over it. This is the approach advocated by [Drew Devault](https://drewdevault.com/2019/12/09/Developers-shouldnt-distribute.html).

That would means on macOS running `brew install dcm2niix` and on Windows `choco install dcm2niix` or using `curl`'ing the prebuilt binaries for Windows. Those are the most canonical ways to install out-of-language dependencies for those platforms. In broader context, eventually, someone could package shimming-toolbox into brew or apt and add the dependency in their brew/apt build script; if we used conda in our code then packagers for those platforms would need to do surgery to remove it and fit the package in to their system. It's not impossible, but it's somewhat unfriendly.

Issue #13, opened 2020-07-06
Add the ability to load niftis


po09i:
Convert the current load_niftis function from shimming-toolbox repo to this repo.

The load_niftis function allows to load an entire acquisition instead of only single nifti files at a time. (The original function also allowed to select the acquisition from a list depending on the input)

For example:
The output of dcm2bids for the [dicom_unsorted](https://github.com/shimming-toolbox/data-testing/tree/master/dicom_unsorted) is 6 different phase files organized in `./fmap` called `./fmap/sub-test_phaseX.nii`. The goal of the function is to call `phase = load_niftis/load_acquistion('./fmap')` instead of calling 6 times `nibabel.load()` then stitching them in the required format for our API which we agreed was (x, y, z, echoes, volumes). This example is for different echoes but the function should accommodate different volumes through time as well.

jcohenadad:
@gab-berest what working branch are you for this issue?

gab-berest:
> 
> 
> @gab-berest what working branch are you for this issue?

I'm working in the branch gb/load_nifti. I have to talk to @gaspardcereza to check for the function read_nii since I use it and there was some issue with how it was loaded.

jcohenadad:
> I have to talk to @gaspardcereza to check for the function read_nii since I use it and there was some issue with how it was loaded.

can we rather have the discussion on this github thread (or another more appropriate thread)-- so that the whole team is on the same page-- thanks

gab-berest:
@jcohenadad Is there available Nifti data to download for unittesting? I created dummy data with nibabel for now, but it could be easier if we could download Nifti data directly.

po09i:
Yes, you can use the [download data](https://github.com/shimming-toolbox/shimming-toolbox-py/blob/master/shimmingtoolbox/cli/download_data.py) CLI on master.

jcohenadad:
> @jcohenadad Is there available Nifti data to download for unittesting? I created dummy data with nibabel for now, but it could be easier if we could download Nifti data directly.

https://github.com/shimming-toolbox/data-testing
also see https://github.com/shimming-toolbox/shimming-toolbox-py/pull/15 which includes data download in CI

jcohenadad:
@gab-berest @po09i i'm not convinced about the relevance of this issue-- the body says: "Convert the current load_niftis function from shimming-toolbox repo to this repo"

[...]

PR #18, opened 2020-07-10
Implement dicom_to_nifti: wrapper for dcm2bids to convert DICOM data

gaspardcereza:
## Context
The conversion from dicom to nifti is one of the first step of our workflow. We thus need to have a functional code to allow us to organize our data in the format we want. The former Matlab functions we use must be adapted to the python standards.

This PR introduces the following changes:
- shimmingtoolbox/dicom_to_nifti.py → API
- config/dcm2bids.json --> config file that lists the sequence names used for organizing data under BIDS directory. This file should be edited. https://github.com/shimming-toolbox/shimming-toolbox-py/issues/32

BIDS convention for file naming: https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#task-including-resting-state-imaging-data

dcm2bids documentation about conversion and config file:
- https://github.com/cbedetti/Dcm2Bids/issues/26
- config file for phasediff data: https://github.com/cbedetti/Dcm2Bids/issues/21#issuecomment-369943143

TODO in subsequent PRs:
- cover realistic scenario of experimental workflow. #31
- A function `read_nii()` initially developed here is no more used, but code could be recycled: https://github.com/shimming-toolbox/shimming-toolbox-py/issues/33

Fixes #12.

PR #35, opened 2020-07-27
Add the ability to load niftis

gab-berest:
Integration of the load_nifti function that uses the read_nii functionality to load Nifti type data folders. Linked ot issue #13.


.idea/misc.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jcohenadad, line 1: all these files shouldn't go into the repository

shimmingtoolbox/cli/load_nifti.py
```diff
@@ -0,0 +1,28 @@
+#!usr/bin/env python3
```
jcohenadad, line 1: not sure what is the purpose of this CLI?

shimmingtoolbox/dicom_to_nifti.py
```diff
@@ -0,0 +1,122 @@
+from distutils.dir_util import copy_tree
```
jcohenadad, line 1: that's a duplicate of https://github.com/shimming-toolbox/shimming-toolbox-py/pull/18

having several unmerged branches that share the same development makes it difficult to keep track and to review (we end up making the same comments on both PRs...). is this PR necessary? or can it be re-focused?

gab-berest: It's just that I'm using the content in this functionality. I most say, I would probably put it with the other PR since it is closely related. It was a separate task so I made a separate PR, but I think it would be better to put it together with the other one.

.idea/misc.xml
gab-berest: These were pulled from #18 . I'll check to remove them and add them to .gitignore on the pulled content.

shimmingtoolbox/load_nifti.py
```diff
@@ -0,0 +1,92 @@
...
+import nibabel
+
+from shimmingtoolbox.read_nii import read_nii
+
+logger = logging.getLogger(__name__)
+
+def load_nifti(file_path):
+    """
+    Load data from a NIFTI type file with dcm2bids.
+    :param file_path: absolute or relative path to the directory the acquisition data
```
po09i, line 16: Variable convention : https://github.com/neuropoly/spinalcordtoolbox/wiki/variables#variables
`path_acquisition` or `path_data`

shimmingtoolbox/read_nii.py
```diff
@@ -0,0 +1,73 @@
+import json
+import math
+import nibabel as nib
+import numpy as np
+import os
+
+
+def read_nii(nii_path):
```
jcohenadad, line 8: is this function needed? `nibabel.load()` outputs the image + header (which contains the metadata)

shimmingtoolbox/read_nii.py
```diff
@@ -0,0 +1,73 @@
...
+    # More info in the "Data scaling" section in https://nipy.org/nibabel/nifti_images.html
+    if ('Manufacturer' in json_data) and (json_data['Manufacturer'] == 'Siemens') \
+            and (image_type(json_data) == 'phase'):
+        PHASE_SCALING_SIEMENS = 4096
+        image = image * (2 * math.pi / PHASE_SCALING_SIEMENS)
+
+    return info, json_data, image
+
+
+def image_type(json_data):
```
jcohenadad, line 40: this should be already solved by the BIDS convention in #18

test/test_load_nifti.py
```diff
@@ -0,0 +1,214 @@
+#!usr/bin/env python3
```
jcohenadad, line 1: i would suggest to wait for the dev core team to agree on the proposed developments before implementing the unit tests (otherwise, the unit tests might become stale/irrelevant)

shimmingtoolbox/read_nii.py
gab-berest: I am going to remove it. It is just convenient since it was already programmed with the verifications.

test/test_load_nifti.py
gab-berest: Sure, I implemented them for now so I can validate that the core functionality of the function works correctly, but I will wait for further information.

shimmingtoolbox/load_nifti.py
```diff
@@ -91,7 +91,7 @@ def load_nifti(path_data):
     return niftis, info, json_info
 
 
-def read_nii(nii_path, auto_scale=True):
+def read_nii(nii_path, auto_scale = True):
```
jcohenadad, line 94: i think pep8 asks for no space between equality terms in input arguments

shimmingtoolbox/load_nifti.py
```diff
@@ -0,0 +1,165 @@
...
+
+            if select_acquisition in range(len(acquisitions)):
+                break
+            else:
+                logging.error("Input must be linked to an acquisition folder. {} is out of range".format(input_resp))
+
+        nifti_path = os.path.abspath(file_list[select_acquisition])
+
+    nifti_list = [os.path.join(nifti_path, f) for f in os.listdir(nifti_path) if f.endswith((".nii", ".nii.gz"))]
+    n_echos = len(nifti_list)
```
po09i, line 66: This correctly implements the matlab function but it should be revisited as the number of files is not always the number of echoes. What if there are multiple runs, multiple time points, etc. 

The header should probably be read to figure out the appropriate way to "stitch" the files together.

shimmingtoolbox/load_nifti.py
```diff
@@ -0,0 +1,165 @@
...
+    image = np.asarray(info.dataobj)
+    if auto_scale:
+        if ('Manufacturer' in json_data) and (json_data['Manufacturer'] == 'Siemens') \
+                and (image_type(json_data) == 'phase'):
+            image = image * (2 * math.pi / PHASE_SCALING_SIEMENS)
+
+    return info, json_data, image
+
+
+def image_type(json_data):
```
po09i, line 128: Can nibabel output this for us instead?

shimmingtoolbox/load_nifti.py
```diff
@@ -0,0 +1,165 @@
...
+    elif all([os.path.isfile(f) for f in file_list]):
+        logging.info("Acqusition directory given. Using acquisitions.")
+        nifti_path = path_data
+    else:
+        raise RuntimeError("Directories and files in input path")
+
+    if not nifti_path:
+        for i in range(len(acquisitions)):
+            logging.info("{}:{}\n".format(i, os.path.basename(file_list[i])))
+            print("{}:{}\n".format(i, os.path.basename(file_list[i])))
```
jidicula, line 48: Copied from: https://github.com/shimming-toolbox/shimming-toolbox-py/pull/46#discussion_r465966701

Since our `setup.py` explicitly requires Python 3.6 or greater, we have access to f-strings for string formatting, which is a bit tidier for future readers.

See these resources for details:
- https://realpython.com/python-f-strings/#f-strings-a-new-and-improved-way-to-format-strings-in-python
- https://docs.python.org/3.6/reference/lexical_analysis.html#f-strings

shimmingtoolbox/load_nifti.py
```diff
@@ -0,0 +1,165 @@
...
+            input_resp = input("Enter the number for the appropriate acquisition folder, (type 'q' to quit) : ")
+            if input_resp == 'q':
+                return 0
+
+            select_acquisition = int(input_resp)
+
+            if select_acquisition in range(len(acquisitions)):
+                break
+            else:
+                logging.error("Input must be linked to an acquisition folder. {} is out of range".format(input_resp))
```
jidicula, line 61: See my comment above about f-strings

shimmingtoolbox/load_nifti.py
```diff
@@ -0,0 +1,165 @@
...
+
+        nifti_path = os.path.abspath(file_list[select_acquisition])
+
+    nifti_list = [os.path.join(nifti_path, f) for f in os.listdir(nifti_path) if f.endswith((".nii", ".nii.gz"))]
+    n_echos = len(nifti_list)
+
+    if n_echos <= 0:
+        raise RuntimeError("No acquisition images in selected path {}".format(nifti_path))
```
jidicula, line 69: f-string

jidicula (changes requested):
This is a quick review to catch instances where `str.format()` could be refactored into the cleaner f-string method of string formatting.

shimmingtoolbox/load_nifti.py
gab-berest: True. I should add the ability to detect the correct way to parse through data. From what I've seen for multiple runs, the files are in separate folders which you can choose before hand. But, what do you mean by time points?

shimmingtoolbox/load_nifti.py
gab-berest: I didn't see a image type field in the nibabel header and nibabel cannot open json type files. So the information has to be loaded with another way than nibabel for what I've seen. Maybe I've just missed the field to do it.

shimmingtoolbox/load_nifti.py
gab-berest: If you mean that we should parse the description of the header file, is it really reliable as information?

shimmingtoolbox/load_nifti.py
jcohenadad: @po09i what do we want from this `image_type()` function? what would be example of output "type"?

gab-berest: It only tells us if it's phase or magnitude from the 'ImageType' section in the json file thta is linked to the nifti file. The read_nii function uses it to know if the data needs to be scaled in radian or not. I did not find a good way in the header of the nibabel file to tell us this information.

jcohenadad: don't we already have that information after [BIDS conversion](https://github.com/shimming-toolbox/shimming-toolbox-py/blob/master/config/dcm2bids.json#L105)?

gab-berest: Yes, but dcm2bids automatically creates the json file that contain this information with the data. So I was trying to make this function independent form dcm2bids so when you only have the original nifti files and json without dcm2bids, you could also know this information.

jcohenadad: ah i see. I think at this point in the development i would suggest to not bother covering all possible scenario. It is fair to assume input data is either DICOM or BIDS folder, in which case no need for the `image_type()` function.

shimmingtoolbox/load_nifti.py
gab-berest: @po09i  I think that the NIFTI files generated by dcm2bids will be always as different echoes for each file and the files will contain the rest of the data. Tell me if I'm wrong. I have not seen another example on the BIDS website.

po09i: I found this link to the [bids convention](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#the-run-entity) really useful. It describes what runs are and other specific bids format information. 

We used to separate scans in separate acquisitions however now that we are trying to deal with only having one config file, it is not a guaranty that one acquisition will be in the folder (If we run 2 fieldmaps for example). I think counting the number of files in a folder is not a rigorous check and the only way to make sure is indeed reading the header info.

jcohenadad: let's not spend too much time on this, as it is still unclear how the typical experimental workflow will look like (e.g. will we impose convention for naming runs during acquisition? will we only select the lastest fieldmap? etc.). For now, let's keep it simple and as experimental scenario are clearer (meaning: post-covid19: we can scan again) we will revisit this

gaspardcereza (commented):
Everything seems to be working with the B0 niftis. 
It doesn't work with my B1 maps because the dimensions of my NIfTI aren't the same  (their dimension are [x, y, slice, n_coils]). I'll raise the issue once this PR is merged but I think the easiest solution will be to have a separate fonction dedicated to B1 niftis

shimmingtoolbox/load_nifti.py
```diff
@@ -0,0 +1,161 @@
...
+    elif all([os.path.isfile(f) for f in file_list]):
+        logging.info("Acqusition directory given. Using acquisitions.")
+        nifti_path = path_data
+    else:
+        raise RuntimeError("Directories and files in input path")
+
+    if not nifti_path:
+        for i in range(len(acquisitions)):
+            logging.info("{}:{}\n".format(i, os.path.basename(file_list[i])))
+            print(f"{i}:{os.path.basename(file_list[i])}\n")
```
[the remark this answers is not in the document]
gab-berest: Oups. Forgot to remove it.

gab-berest:
> 
> 
> Everything seems to be working with the B0 niftis.
> It doesn't work with my B1 maps because the dimensions of my NIfTI aren't the same (their dimension are [x, y, slice, n_coils]). I'll raise the issue once this PR is merged but I think the easiest solution will be to have a separate fonction dedicated to B1 niftis

What are your data files so I can test. Normally, it would, with these dimensions, create a 5D array of [x,y,slices(as z),echoes,n_choil]. Maybe an argument could be given to load B0 or B1 maps or just by header information.

gaspardcereza:
> > Everything seems to be working with the B0 niftis.
> > It doesn't work with my B1 maps because the dimensions of my NIfTI aren't the same (their dimension are [x, y, slice, n_coils]). I'll raise the issue once this PR is merged but I think the easiest solution will be to have a separate fonction dedicated to B1 niftis
> 
> What are your data files so I can test. Normally, it would, with these dimensions, create a 5D array of [x,y,slices(as z),echoes,n_choil]. Maybe an argument could be given to load B0 or B1 maps or just by header information.

I used the data that are in the `b1_maps` folder in the testing-data repository. But I convert them using the `dicom_to_nifti` function for B1 maps that is not merged yet (still some work to do on them).
Even when I try `load_nifti` after a simple call to dcm2niix on my `b1_maps` folder as follows: 
```
loaded_nifti = load_nifti.load_nifti(r'C:\Users\gaspa\Documents\GitHub\shimming-toolbox-py\testing_data\b1_maps')
``` 
It returns a 5d array that is a 64x2x8x2x64 array instead of (64x64x2x16 in the original nifti). 
```
loaded_nifti[0].shape
(64, 2, 16, 1, 64)
```

gab-berest:
> 
> 
> > What are your data files so I can test. Normally, it would, with these dimensions, create a 5D array of [x,y,slices(as z),echoes,n_choil]. Maybe an argument could be given to load B0 or B1 maps or just by header information.
> 
> I used the data that are in the `b1_maps` folder in the testing-data repository. But I convert them using the `dicom_to_nifti` function for B1 maps that is not merged yet (still some work to do on them).
> Even when I try `load_nifti` after a simple call to dcm2niix on my `b1_maps` folder as follows:
> 
> ```
> loaded_nifti = load_nifti.load_nifti(r'C:\Users\gaspa\Documents\GitHub\shimming-toolbox-py\testing_data\b1_maps')
> ```
> 
> It returns a 5d array that is a 64x2x8x2x64 array instead of (64x64x2x16 in the original nifti).
> 
> ```
> loaded_nifti[0].shape
> (64, 2, 16, 1, 64)
> ```

True enough, the answer should be a 64x64x2x16x1. I'll check it later.

shimmingtoolbox/load_nifti.py
gab-berest: I changed the function so it will detect if there are multiple runs in the same folder and ask for which one to load. It should be clarified for later on.

gab-berest:
I added a case that checks the run number and asks for the one to analyze and an argument that tells if phase or magnitude needs to be loaded (checked from ImageComments like dicom_to_nifti.

jidicula (changes requested):
Could you also add automodule stubs for `load_nifti` in the API documentation? It'll be a one-liner in `docs/source/6_api_reference/api.rst`, and you can follow the format of existing stubs for modules like `download` or `utils`.

gab-berest:
> 
> 
> Could you also add automodule stubs for `load_nifti` in the API documentation? It'll be a one-liner in `docs/source/6_api_reference/api.rst`, and you can follow the format of existing stubs for modules like `download` or `utils`.

Done

gab-berest:
Is there something to change? @jcohenadad @po09i @jidicula

shimmingtoolbox/load_nifti.py
```diff
@@ -111,14 +120,14 @@ def load_nifti(path_data, modality = 'phase'):
...
+            nifti_init[0][0].shape[0], nifti_init[0][0].shape[1], nifti_init[0][0].shape[2],
+            echo_shape, nifti_init[0][0].shape[3]], dtype=float)
         for i_echo in range(len(run_list[select_run])):
             if modality in run_list[select_run][i_echo][1]:
                 tmp_nii = run_list[select_run][i_echo][0]
                 info.append(tmp_nii[0].header)
                 json_info.append(tmp_nii[1])
-                for i_volume in range(info_init[0][0].shape[3]):
+                for i_volume in range(nifti_init[0][0].shape[3]):
                     niftis[:, :, :, nifti_pos, i_volume] = tmp_nii[2][:, :, :, i_volume]
```
jidicula: Other than this inner for-loop, the enclosing for loop is a duplicate of line 114-120. This is out of scope for the current PR, but is a prime candidate for refactoring in the future. The duplicated code could be factored out into a function taking a boolean param that only runs the inner for-loop for the appropriate case. Also, depending on the size of `range(len(run_list[select_run]))` and `range(nifti_init[0][0].shape[3])`, the `map` function could be used to achieve better performance.

jidicula (approved):
Doc stubs look ok to me!

shimmingtoolbox/load_nifti.py
```diff
@@ -0,0 +1,168 @@
...
+    Args:
+        path_data (str): Absolute or relative path to the directory the acquisition data
+        modality (str): Modality to read nifti (can be phase or magnitude)
+    Returns:
+        nibabel.Nifti1Image.Header: List containing all information from every Nifti image
+        dict: List containing all information in JSON format from every Nifti image
+        numpy.ndarray: 5D array of all acquisition in time (x, y, z, echo, volume)
+
+    Note:
+        If 'path' is a folder containing niftis, directly output niftis. It 'path' is a folder containing acquisitions,
```
jcohenadad, line 27: "folder containing acquisitions" --> what do you mean by "acquisitions"? do you mean DICOM files?

shimmingtoolbox/load_nifti.py
```diff
@@ -0,0 +1,168 @@
...
+
+logger = logging.getLogger(__name__)
+PHASE_SCALING_SIEMENS = 4096
+
+
+def load_nifti(path_data, modality='phase'):
+    """
+    Load data from a NIFTI type file with nibabel.
+    Args:
+        path_data (str): Absolute or relative path to the directory the acquisition data
```
jcohenadad, line 19:
```suggestion
        path_data (str): Path to the directory containing the file(s) to load
```

shimmingtoolbox/load_nifti.py
```diff
@@ -0,0 +1,168 @@
...
+import json
+import math
+
+logger = logging.getLogger(__name__)
+PHASE_SCALING_SIEMENS = 4096
+
+
+def load_nifti(path_data, modality='phase'):
+    """
+    Load data from a NIFTI type file with nibabel.
```
jcohenadad, line 17: i find it confusing: here it says the function loads from a nifti file, but `path_data` is expecting a folder (not a file)

shimmingtoolbox/load_nifti.py
```diff
@@ -0,0 +1,168 @@
...
+    Args:
+        path_data (str): Absolute or relative path to the directory the acquisition data
+        modality (str): Modality to read nifti (can be phase or magnitude)
+    Returns:
+        nibabel.Nifti1Image.Header: List containing all information from every Nifti image
```
jcohenadad, line 22:
```suggestion
        nibabel.Nifti1Image.Header: List containing headers for every Nifti file
```

shimmingtoolbox/load_nifti.py
```diff
@@ -0,0 +1,168 @@
...
+
+    if os.path.isfile(json_path):
+        json_data = json.load(open(json_path))
+    else:
+        raise ValueError('Missing json file')
+
+    image = np.asarray(info.dataobj)
+    if auto_scale:
+        if ('Manufacturer' in json_data) and (json_data['Manufacturer'] == 'Siemens') \
+                and ("phase" in json_data['ImageComments']):
```
po09i, line 161: This seems like an edge case. "ImageComments" won't always have in the description if it is phase data or not. For example, our testing data does not even have the field ImageComments and uses the more useful ImageType to differentiate between phase, mag and complex.

shimmingtoolbox/load_nifti.py
```diff
@@ -0,0 +1,168 @@
...
+            image = image * (2 * math.pi / PHASE_SCALING_SIEMENS)
+
+    return info, json_data, image
+
+
+if __name__ == "__main__":
+    load_nifti("C:\\Users\\Gabriel\\Documents\\share\\008_a_gre_DYNshim1")
```
po09i, line 168: This will obviously need to be removed before merging

shimmingtoolbox/load_nifti.py
gaspardcereza: It would also maybe be better to check for "*phase*" instead of just "phase" in case the user decides to write some other comments in the "ImageComments" tag.

shimmingtoolbox/load_nifti.py
gab-berest: Yes it was while I was testing

shimmingtoolbox/load_nifti.py
gab-berest: I was checking the ImageType before, but this field is also not always present in the json file outputted by dcm2bids. I read the other issues to find that @gaspardcereza was adding the phase, mag, complex tag in the ImageComments. That's why I used it.

shimmingtoolbox/load_nifti.py
gab-berest: You can specify a directory wich contains multiple other directories with the data files. It just tells you that you need to choose one of the folders contained in the given path. Should I rename it? It was named like this in the MATLAB version.

shimmingtoolbox/load_nifti.py
gab-berest: Yes it should say: "from a directory containing NIFTI type files"

shimmingtoolbox/load_nifti.py
gab-berest: It's true that I doubled my code. I will try to remove it. I don't think a separate function will be usefull since it will be used only these two times and that we can merge the duplicated code into a more condensed code.

gab-berest:
@po09i @jcohenadad Are there any other questions to merge since a subsequent PR might change some feature for PyBIDS?

jcohenadad:
> @po09i @jcohenadad Are there any other questions to merge since a subsequent PR might change some feature for PyBIDS?

all good for me

gaspardcereza (approved):

po09i (approved):

PR #43, opened 2020-07-30
Upgrade dcm2niix to latest.

### LINKED ISSUE #42: Controlling which version of dcm2niix is installed
# Context
Has mentionned [here](https://github.com/shimming-toolbox/shimming-toolbox-py/pull/18#issuecomment-665818569), Travis downloads dcm2niix `v1.0.20171215` instead of `v1.0.20200331` which is the last release.
It might be the cause of the test failure when pushing #18. 

Currently, dmc2niix is installed by running `sudo apt-get install -y dcm2niix` in [.travis.yml](https://github.com/shimming-toolbox/shimming-toolbox-py/blob/master/.travis.yml).

kousu:
Fixes #42, and fixes the broken tests in #18 so it will need to be rebased under #18 for that one to work.


README.md
```diff
@@ -5,6 +5,16 @@ Code for performing real-time shimming using external MRI shim coils
 
 ## Installation
 
+The first dependency you need is [`dcm2niix`](https://github.com/rordenlab/dcm2niix) >= v1.0.20200331.
+You can download precompiled binaries from https://github.com/rordenlab/dcm2niix/releases/latest -- make
+sure to add them to your PATH.
+It is also [in the AUR](https://aur.archlinux.org/packages/dcm2niix/) (`pikaur -Sy dcm2niix`),
+[`brew`](https://github.com/Homebrew/homebrew-core/blob/master/Formula/dcm2niix.rb) (`brew install dcm2niix`),
+and [`conda`](https://anaconda.org/conda-forge/dcm2niix) (`conda install -c conda-forge/label/cf202003 dcm2niix`).
+Unfortunately, the version [currently in Debian/Ubuntu](https://packages.ubuntu.com/eoan/dcm2niix) (`apt install dcm2niix`) is too old to work reliably.
```
jcohenadad, line 14: the problem with that, is we won't be testing those other installation methods. 

to prevent any issue in the future like https://github.com/shimming-toolbox/shimming-toolbox-py/pull/18#issuecomment-665818569, i would suggest to instruct users to only fetch the latest binaries provided on github. Alternatively, we could implement an installer which will fetch the appropriate binary depending on the platform.

jcohenadad:
@kousu sorry i screwed up-- i was working with Gaspard and Alexandre on the dcm2bids branch, which we wanted to merge so we could work on the data management thingy-- i forgot to merge this one before the other one as you asked-- sorry about that---

jcohenadad (approved):

README.md
kousu: An integrated installer won't integrate well with external systems or distros. It would mean running everything in an expensively isolated box.

jcohenadad: ok, in that case, we should have a routine in shimmingtoolbox that verifies which version of dcm2niix is run by default by the user, and if this is an incorrect version, it errors and quit-- how about that?

Issue #20, opened 2020-07-14
Implement optimization algorithms for shimming


jcohenadad:
It would be nice to have a testing script that generates a B0 fieldmap from test data, in order to start implementing different algorithms for optimizing B0 shimming. 

Once https://github.com/shimming-toolbox/shimming-toolbox-py/pull/15 is merged, the same script could be used as a basis for these investigations.

Things to consider:
- We will need coil profiles #21 ;

lcbMGH:
Our current model, to be discussed at the next toolbox meeting:

Two layers for each package -- sequencer and optimizer.

The optimizer takes:
- X*Y*Z integer valued mask array, where integers represent different materials (eg. 1 for brain, 2 for skull, etc.)
- X*Y*Z float valued unshimmed array
- X*Y*Z*N float valued field map of all N coils
- Physical constraints (extendable data structure)
- Iteration/optimization constraints (extendable data structure)

All of these should be pre-aligned

The sequencer allows for more complicated shim sequences, such as oblique slices, volumetric shimming, or different structures on different materials. Different sequencers might need different parameters, and we could discuss what those might look like. The basic idea is that the sequencer deals with time variation, and sends in a single mask to the optimizer to iteratively output the coil channel currents for each time value. A predictive "shimmed" field map could be generated from these and the field maps, for visualization.

jcohenadad:
To move forward, we should start by implementing testing suites that cover different scenario:
- realtime shimming: one optimization (of slice or slice group) per time point, driven by respiratory trace.
- dynamic shimming: one optimization per slice (or volume/slab or odd/even, or SMS group of slices): done by the "optimizer", and orchestrated by the parent "sequencer". The sequencer would notably pass regularization parameters (e.g. spatial/temporal smoothness between adjacent slices/slice_groups). In this case, the "trigger" is coming from the pulse sequence, sent to the microcontroller that is directly connected to the amplifier board (ie bypasses the laptop to avoid lags-- so the shim currents have already been loaded into the microcontroller).
- multi-region shimming: sub-case of dynamic shimming where two or more regions will be shimmed (e.g. fat during fat saturation pulse; and tissue during tissue excitation and readout). In that case, two (or more) masks will be input.

Suggestion for testing suite:
[...]

PR #46, opened 2020-07-30
Updated general demo script: general_demo.py

jcohenadad:
This script will include most functionalities.

TODO:
- [x] Add download of test data
- [x] Add `dicom_to_nifti` functionality
- [x] Add usage instructions in an example section under getting_sarted for the website

TODO in subsequent PRs:
- include masking (function to segment, BET or SCT or else) #47 
   - Update dataset to include anatomical data, multiple echoes
- include optimizer for finding shim currents in the static case #20 
- output txt file or whatever output is needed


jcohenadad:
something to consider: is this testing_data appropriate for this general_demo script (e.g. are there anat data to segment from, etc.)

jidicula (changes requested):
Could the usage instructions for this demo script be added to the documentation?

po09i:
Testing_data is indeed not appropriate, we should include anatomical data, multiple echoes and volumes through time in the same dataset. These should also not be sorted and in Dicom format. This data will also be useful for testing `load_nii` from this [PR](https://github.com/shimming-toolbox/shimming-toolbox-py/pull/35)

po09i:
> Could the usage instructions for this demo script be added to the documentation?

Yes that's a good idea. @jidicula What do you think of putting this in a subsection of `getting_started` or just creating a new `example` section?

jidicula:
> > Could the usage instructions for this demo script be added to the documentation?
> 
> Yes that's a good idea. @jidicula What do you think of putting this in a subsection of `getting_started` or just creating a new `example` section?

@po09i I think `2_getting_started.rst` is an ideal place to include it, perhaps under a heading called "Example".

gab-berest:
> 
> 
> > > Could the usage instructions for this demo script be added to the documentation?
> > 
> > 
> > Yes that's a good idea. @jidicula What do you think of putting this in a subsection of `getting_started` or just creating a new `example` section?
> 
> @po09i I think `2_getting_started.rst` is an ideal place to include it, perhaps under a heading called "Example".

For a general demo showing the use of the package, getting started would be a good place. Then when we will have more complex example for advanced functionality, we could create another section since it will not be linked to getting started information

examples/general_demo.py
```diff
@@ -41,7 +44,7 @@ def main():
     phase_e2 = nii_phase_e2.get_fdata() / 4096.0 * 2.0 * np.pi - np.pi
```
gab-berest, line 44: I've seen this scaling in some other files and functionality. It is a detail, but I think it would be good to make an utility function that scales. This way, it will be easier to update our code in the futur. Also, I don't think it is a good idea to explicitly scale phases in the example since it can be prone to errors from the user part:
```suggestion
    phase_e2 = scale_phase(nii_phase_e2.get_fdata(), [-np.pi, np.pi])
    
def scale_phase(phase_data, range)
    return np.substract(np.multiply(np.divide(phase_data, 4096.0), 2.0*np.pi), np.pi)
    #or return np.interp(-np.pi, np.pi, phase_data)
```
The use of numpy functions is generally faster for large arrays and implements a true division which adapts the output format for better precision instead of the floor division in Python.

po09i: Used `np.interp(nii_phase_e2.get_fdata(), [0, 4096], [-np.pi, np.pi])` but removed the scale_phase function since it added a layer of function call without removing any line of code in the script. `interp` is also self descriptive.

examples/general_demo.py
```diff
@@ -0,0 +1,104 @@
...
+
+    # Open mag data
+    fname_mags = glob.glob(os.path.join(path_nifti, 'sub-01', 'fmap', '*magnitude*.nii.gz'))
+
+    nii_mag_e1 = nib.load(fname_mags[0])
+    nii_mag_e2 = nib.load(fname_mags[1])
+
+    # TODO: create mask
+    # Call SCT or user defined mask
+    # mask = np.ones(phase_e1.shape)
```
jidicula, line 62: Ideally TODOs shouldn't make it into production – would this belong in an issue ticket instead?

examples/general_demo.py
```diff
@@ -0,0 +1,104 @@
...
+
+    # Open phase data
+    fname_phases = glob.glob(os.path.join(path_nifti, 'sub-01', 'fmap', '*phase*.nii.gz'))
+
+    nii_phase_e1 = nib.load(fname_phases[0])
+    nii_phase_e2 = nib.load(fname_phases[1])
+
+    # Scale phase to radians
+    phase_e1 = np.interp(nii_phase_e1.get_fdata(), [0, 4096], [-np.pi, np.pi])
+    phase_e2 = np.interp(nii_phase_e2.get_fdata(), [0, 4096], [-np.pi, np.pi])
```
jidicula, line 52: When scaling to radians, are each of these args always the same except for the first? If so, it might benefit maintainability to abstract this scaling operation into a helper function that can be imported since it's repeated here and in `test_prelude.py`. Perhaps it could go in `utils.py` and be imported from there.

test/test_prelude.py
```diff
@@ -49,8 +49,8 @@ def get_phases_mags_affines(self):
         nii_phase_e2 = nib.load(fname_phases[1])
 
         # Scale to phase to radians
-        phase_e1 = nii_phase_e1.get_fdata() / 4096 * 2 * np.pi - np.pi
-        phase_e2 = nii_phase_e2.get_fdata() / 4096 * 2 * np.pi - np.pi
+        phase_e1 = np.interp(nii_phase_e1.get_fdata(), [0, 4096], [-np.pi, np.pi])
+        phase_e2 = np.interp(nii_phase_e2.get_fdata(), [0, 4096], [-np.pi, np.pi])
```
jidicula, line 53: see my comment above about abstracting this phase-scaling into a function.

jidicula (changes requested):
My review includes a suggestion for including an automodule stub in the documentation, a refactoring of a format string into an f-string literal, and a suggestion for abstracting phase scaling into a separate function.

If these changes are out of scope, let me know and I can approve the PR.

examples/general_demo.py
po09i: The issue #47 is already created, I could remove the TODOs in the script.

examples/general_demo.py
po09i: @gab-berest also had this idea, my initial thought is that `[0, 4096]` might change for non siemens data. We only deal with siemens so it might not be relevant but I still like the idea of being flexible and not introduce a source of error. So now with 2 input arguments I'm not sure it is worth it to introduce a new `scale_phase(data, input_range)` if people are already familiar with interp. What do you think?

examples/general_demo.py
```diff
@@ -0,0 +1,104 @@
...
+
+    Returns:
+        str: File name and path of output figure
+    """
+
+    logging.basicConfig(level='INFO')
+
+    # Download example data
+    path_testing_data = os.path.join(path_output, __dir_testing__)
+    run_subprocess('st_download_data {} --output {}'.format(__dir_testing__, path_testing_data))
```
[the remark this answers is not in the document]
po09i: Did not know this was a thing. I really like it thanks!

docs/source/2_getting_started/2_getting_started.rst
```diff
@@ -1,2 +1,30 @@
...
+plot. More precisely, it will:
+
+* Download unsorted dicoms
+* Run dcm2bids to convert to nifti with bids structure
+* Unwrap phase
+* Save wrapped and unwrapped plot of first X,Y volume as unwrap_phase_plot.png in the output directory
+
+To use the script, simply provide an output directory where the processing and output figure will be generated, if no
+directory is provided, a directory *output_dir* will be created in the current directory.
+
```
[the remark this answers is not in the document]
po09i: I'm having trouble to run sphinx locally, could you describe the procedure above in a wiki page on GH and i'll fill it in with my findings.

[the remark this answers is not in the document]
po09i: Issue was that I installed sphinx with pip3 instead of pip so the command `python -msphinx` did not work. Could we go for `python3 -msphinx` instead since the project requires python > 3.6

examples/general_demo.py
jcohenadad: >  I'm not sure it is worth it to introduce a new scale_phase(data, input_range)

sounds like the best option to me: isolating the "scaling" problem, with unknown data provenance (e.g. from siemens, philips, ge, toshiba, different sequences, software versions, etc.). And you could create an exhaustive sample of data and a specific unit test.

docs/source/2_getting_started/2_getting_started.rst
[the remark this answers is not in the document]
jidicula: Oops, I neglected to add this installation step in the docs: in order to build the package locally with sphinx as a dependency, you can install the package (inside your activated miniconda virtualenv) with `shimming-toolbox-py: $ pip install -e .[testing,docs]`. I'll add this as an issue.

This *should* add `sphinx` to your path, allowing the `docs/Makefile` to function correctly.

examples/general_demo.py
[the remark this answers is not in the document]
jidicula: My pleasure, always happy to spread the word about f-strings 😁

docs/source/2_getting_started/2_getting_started.rst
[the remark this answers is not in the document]
jidicula: Alternatively, once you push the commit to this branch remote, you can see the resulting build on RTD here: https://shimming-toolbox-py--46.org.readthedocs.build/en/46/. The link is available in the same list of checks where the Travis branch build link is located:

Screen Shot 2020-08-06 at 16 07 22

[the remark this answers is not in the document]
po09i: Thanks for the info. I really like that we can build locally so we don't over commit. I'll take a look at your PR and it should fix the issues.

examples/general_demo.py
po09i: Opened an [issue](https://github.com/shimming-toolbox/shimming-toolbox-py/issues/64) to implement a scale_phase function. It is out of scope of this PR so it will be implemented in the future.

jidicula (approved):
LGTM!

PR #51, opened 2020-07-31
Add FSL license for prelude

po09i:
Prelude is an external dependency and should have its own license. This PR adds FSL's license.

License taken from: https://fsl.fmrib.ox.ac.uk/fsl/fslwiki

### DIFF
--- a/shimmingtoolbox/unwrap/LICENSE_FSL_PRELUDE.md
+++ b/shimmingtoolbox/unwrap/LICENSE_FSL_PRELUDE.md
@@ -1,0 +1,93 @@
+FMRIB Software Library, Release 6.0 (c) 2018, The University of Oxford
+(the "Software")
+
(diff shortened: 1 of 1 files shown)


gaspardcereza (approved):

jcohenadad (changes requested):
we don't actually bundle FSL's prelude in this repository but in this [other repository](https://github.com/shimming-toolbox/binaries).

what is the reason you'd like to add the license here?

po09i:
I thought it should be bundled where we use it but I can do the same PR in this [binaries repo](https://github.com/shimming-toolbox/binaries) instead.

jcohenadad:
so i guess we can close after you create the PR in the other repos and cross-ref here--

po09i:
PR moved to [binaries repo](https://github.com/shimming-toolbox/binaries/pull/1) 
This can be closed

Issue #55, opened 2020-07-31
Fix docstring formatting in `utils.py`


jidicula:
There are some issues with docstring formatting in `shimmingtoolbox/utils.py` that are throwing warnings from Sphinx-autodoc when generating the function documentation stubs for that file.

```
/Users/johanan/neuropoly/shimming-toolbox-py/shimmingtoolbox/utils.py:docstring of shimmingtoolbox.utils.run_subprocess:3: WARNING: Unexpected indentation.
```

jidicula:
https://github.com/shimming-toolbox/shimming-toolbox-py/pull/56#issuecomment-667320630

Issue #54, opened 2020-07-31
Fix docstrings in `download.py`


jidicula:
There are some issues with docstring formatting in `shimmingtoolbox/download.py` that are throwing warnings from Sphinx-autodoc when generating the function documentation stubs for that file.

```
/Users/johanan/neuropoly/shimming-toolbox-py/shimmingtoolbox/download.py:docstring of shimmingtoolbox.download.install_data:7: WARNING: Unexpected indentation.
```

jidicula:
https://github.com/shimming-toolbox/shimming-toolbox-py/pull/56#issuecomment-667320630

Issue #53, opened 2020-07-31
Fix docstrings in `NumericalModel`


jidicula:
There are some issues with docstring formatting in `shimmingtoolbox/simulate/numerical_model.py` that are throwing warnings from Sphinx-autodoc when generating the documentation stubs for the NumericalModel class.

Edit: partially resolved by #57, remaining warning is below:

```
/Users/johanan/neuropoly/shimming-toolbox-py/shimmingtoolbox/simulate/numerical_model.py:docstring of shimmingtoolbox.simulate.numerical_model.NumericalModel.generate_deltaB0:9: WARNING: Unexpected indentation.
```

jidicula:
https://github.com/shimming-toolbox/shimming-toolbox-py/pull/56#issuecomment-667320630

PR #56, opened 2020-07-31
[docs][API reference] Add unwrap module to API reference

### LINKED ISSUE #52: Add unwrap module to API reference in documentation
The unwrap module currently contains a docstring that can be picked up by Sphinx-autodoc if the directive is added to the API reference.

jidicula:
Resolves: #52

### DIFF
--- a/docs/source/6_api_reference/api.rst
+++ b/docs/source/6_api_reference/api.rst
@@ -21,3 +21,6 @@ The following section outlines the API of shimming-toolbox.
 
 .. automodule:: shimmingtoolbox.simulate.numerical_model
    :members:
+
+.. automodule:: shimmingtoolbox.unwrap.prelude
+   :members:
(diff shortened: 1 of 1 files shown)


po09i:
We are using the google style docstring which is not the default way that sphinx parses the docstring. We might need to add extensions to fix the output. 

This might be of interest: https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html

jidicula:
> We are using the google style docstring which is not the default way that sphinx parses the docstring. We might need to add extensions to fix the output.
> 
> This might be of interest: https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html

Thanks @po09i, I've updated #53, #54, and #55 with links to your comment. Should be able to knock them all out by adding that extension.

po09i (changes requested):
Can you rabase on master so we can see the changes from the previous merge that fixed the google style docstring

po09i (approved):

PR #60, opened 2020-08-06
Added masking functionality

### LINKED ISSUE #47: Implement masking function
## Context

A mask is a binary image, with voxels=1 in regions that are anatomically or geometrically relevant (e.g. skull, brain, etc.).

In the context of this project, it is useful for:
- unwrapping
- RF shimming
- B0 shimming

There are a few algorithms/software that exist to perform masking:
- SCT: for spinal cord + simple shapes (license MIT)
- FSL BET  (non-commercial): for brain
- Others?

For simple shape masking (e.g. square), code is easy and could potentially be included inside shimming-toolbox.

## Suggestion

We could implement a wrapper similar to `shimmingtoolbox.unwrap.prelude()` for masking. E.g.:
- `shimmingtoolbox.masking.shape(shape={'square', 'cube', 'disk', 'ball'})` --> input is a np.array (the image to mask)
- `shimmingtoolbox.masking.threshold(thr=DEFAULT_THR)` --> input is a np.array; scikit-image has a lot of thr algo
  - we could consider algo that works with phase images
- `shimmingtoolbox.masking.bet()` --> input is file name
- `shimmingtoolbox.masking.sct()` --> input could be file name or an np.array (if we use SCT's API)
output: mask as nd.array (of file in the case of bet)

## Related to
Originally as a TODO in [general_demo script]([url](https://github.com/shimming-toolbox/shimming-toolbox-py/pull/46/commits/57126cbcef79abddf09d1927d489c7775a134953#diff-f849ab6d978a45df1c493a686d102cb7L60))

jcohenadad:
## Context

A mask is a binary image, with voxels=1 in regions that are anatomically or geometrically relevant (e.g. skull, brain, etc.).

In the context of this project, it is useful for:
- unwrapping
- RF shimming
- B0 shimming

This PR implements a new module `shimmingtoolbox.masking` with a few masking capabilities:
- `shimmingtoolbox.masking.shape(shape={'square', 'cube', 'disk', 'ball'})`
  - Input is a numpy.ndarray (the image to mask)
  - output: mask as numpy.ndarray
- `shimmingtoolbox.masking.threshold(thr=DEFAULT_THR)`
  - Input is a numpy.ndarray (the image to mask)
  - output: mask as numpy.ndarray
- Added as an automodule to be documented on the website

output: mask as numpy.ndarray

## In subsequent PR's
- We could consider algo that works with phase images for the threshold algorithm
- `shimmingtoolbox.masking.bet()` --> input is file name (or np.array?)
- `shimmingtoolbox.masking.sct()` --> input could be file name or an np.array (if we use SCT's API)

Fixes #47


po09i:
Any feedback on the `masking.shape` part of the PR, I'm thinking to structure the `shape` function as a wrapper to the different shapes to support ('square', 'cube', 'disk', 'ball'). It allows to call only one function for shape masking. That being said I'm not a big fan of the number of argument that ends up in the "wrapper" (`masking.shape`). It also makes it unclear which argument is necessary for which shape.

shimmingtoolbox/masking/shape.py
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
+"""
+Image mask with shape API
+"""
+
+import numpy as np
+
+
+def shape(data, shape, center_x=None, center_y=None, center_z=None, len_x=None, len_y=None, len_z=None, radius=None):
```
gab-berest, line 10: I think a wrapper function would be a good way to go. Normally when using wrapper function like the one you have, you should use *args and **kargs and then for each call, check if the correct arguments are passed down to the wrapper. This way, it can be used with any type of function. And could even be implemented as a decorator, but it is not necessary:
```suggestion
def shape(data, shape, **kargs):
    mask_info = {}
    for key, value in kargs.items():
        mask_info[key] = value
```
Then you would have a dictionary with all the information provided and you can parse through it to check if the required fields for your shape are present.

shimmingtoolbox/masking/shape.py
```diff
@@ -0,0 +1,70 @@
...
+        # TODO
+        pass
+    elif shape == 'disk':
+        # TODO
+        pass
+    elif shape == 'ball':
+        # TODO
+        pass
+    else:
+        raise RuntimeError('Input shapes not part of the allowed shapes for masking')
```
gab-berest, line 43: A good way to avoid using if elif is to make a global dictionary with all the functions for the shapes keyed with the shape name and only call this function:
```suggestion
   shape_mask[shape](data, mask_info)
```
And at the beginning of the document:
```
shape_mask = {'square': shape_square, 'cube': shape_cube, 'disk': shape_disk, 'ball': shape_ball}
```
This way, it is easier to call. You can also unwrap the information in the mask_info to use with you're current function call.

shimmingtoolbox/masking/shape.py
```diff
@@ -0,0 +1,70 @@
...
+    Args:
+
+    Returns:
+        nd.array: mask
+    """
+
+    # allowed_shapes = {'square', 'cube', 'disk', 'ball'}
+
+    # Make sure input params are valid
+    if data.ndim not in [2, 3]:
```
gab-berest, line 23: If you check dimensions in your masking function afterward, this line becomes redundant and prone to mistake afterward.

shimmingtoolbox/masking/threshold.py
```diff
@@ -0,0 +1,28 @@
...
+import os
+import nibabel as nib
+import pathlib
+import tempfile
+import logging
+
+from shimmingtoolbox.utils import run_subprocess
+
+
+def threshold(data, thr=30):
```
gab-berest, line 17: Is it necessary to have a function that only does this? It adds another layer of complexity and linking for the executer. If it will be used at multiple places or do other things then it is alright, but this function seems quite short.

jcohenadad: probably not, but at some point we might have "smarter" thresholding methods à la Otsu, patch-based thresholding, etc. so the idea is just to get started with something trivial so we can move forward with our example scenario

shimmingtoolbox/masking/shape.py
po09i: That's an awesome tip, thanks. I had to modify the `shape_mask` function call to `shape_mask[shape](data, **mask_info)` to keep the original function declaration.

shimmingtoolbox/masking/threshold.py
gab-berest: Good, just wanted to be sure.

po09i:
Opened issue #66 for future work on masking

jcohenadad:
@po09i i approve this PR (i opened it, so i cannot officially review 😉 )

po09i (approved):
Approved for @jcohenadad

Issue #58, opened 2020-08-03
dcm2niix doesn't separate magnitude and phase with B1 mapping dicoms


gaspardcereza:
## Context
(Issue originally opened in [shimming-toolbox](https://github.com/shimming-toolbox/shimming-toolbox/issues/121))

When I run [dcm2niix](https://github.com/rordenlab/dcm2niix) with RF mapping `DICOM` (acquired with a turbo flash sequence on a 7T Siemens Magneton Terra), it only outputs one `NIfTI` file (and its corresponding `json` file) that gather magnitude and phase. 

I think that might be related to the fact that these `DICOM` files doesn't have the exact same fields as the others we play with (e.g. no `ImageType` field). By default, the phase and magnitudes are thus stored in the same `NIfTI` file. 

Here the field that could be used to separate phase and magnitude images is `ImageComments` that has the value `'flip angle map, TraRefAmpl: 400.0 V'` for the magnitude and `'relative phase map'` for the phase but it seems to be filled by the user during the scan --> that might not be a good option as it is sensitive to human errors.

## Possible solutions
- Contact the `dcm2niix` team to find a way to separate the phases and magnitudes based on the `ImageComments` field,
- Find a way to obtain RF mapping `DICOM` files with the 'classic' field names 

## Data
The DICOM files I use are standard RF maps from the MNI measurements:
[RF_dicoms.zip](https://github.com/shimming-toolbox/shimming-toolbox-py/files/5018048/RF_dicoms.zip)

The easiest way to call the function from the command prompt is to go in the folder where the data have been extracted and type: `dcm2niix RF_dicoms`

My test was done on Windows 10 64-bits with dcm2niix v1.0.20200331 (latest version)

jcohenadad:
@gaspardcereza i suggest you open an issue on dcm2niix, and if this issue cannot be solved on their end, we can always add a checker for 4d data in our `dicom_to_bids` function and split across the time dimension if needed

gaspardcereza:
@jcohenadad  I got an [answer](https://github.com/rordenlab/dcm2niix/issues/414#issuecomment-668509501) from Chris Rorden: 

>This is a limitation of your images, not my software.

[...]

PR #63, opened 2020-08-06
Read B1 maps

gaspardcereza:
## Context
The RF maps are here directly measured by the scanner that returns them as `DICOM` files. We then want to convert these `DICOM` files to the `NIfTI` format with respect of the BIDS conventions. 

That conversion is thus the first step of our B1 mapping process.

## How to test this PR

The test `test_dicom_to_nifti.py` downloads data and has a test sensitive to the related issue.

More details about the data: The DICOM files I use are standard RF maps from the MNI measurements:
[RF_dicoms.zip](https://github.com/shimming-toolbox/shimming-toolbox-py/files/5018048/RF_dicoms.zip)

The easiest way to call the function from the command prompt is to go in the folder where the data have been extracted and type: `dcm2niix RF_dicoms`

## Problem encountered with our dataset
See #58 for more details.
When running [dcm2niix](https://github.com/rordenlab/dcm2niix) with B1 mapping `DICOM` files, it only outputs one `NIfTI` file (and its corresponding `json` file) that gather magnitude and phase. 

That is be related to the fact that these `DICOM` files don't have the exact same fields as the others we play with (e.g. no `ImageType` field). By default, the phase and magnitudes are thus stored in the same `NIfTI` file. 

Here the field that could be used to separate phase and magnitude images is `ImageComments` that has the value `'flip angle map, TraRefAmpl: 400.0 V'` for the magnitude and `'relative phase map'` for the phase but it seems to be filled by the user during the scan --> that might not be a good option as it is sensitive to human errors.

## Proposed changes

- After calling `dcm2bids`, separate the phase and magnitude files into 2 different nifti files, if modality is listed in `special_dicom` argument.
- Delete original nifti once phase and magnitude niftis are created
- Updated URL in download_data to include `b1_maps` in the downloaded release
- Updated `config/dcm2bids.json` to add the T1map case, as per [BIDS convention](https://bep001.readthedocs.io/en/dev/04-modality-specific-files/01-magnetic-resonance-imaging-data.html)

### DIFF
--- a/config/dcm2bids.json
+++ b/config/dcm2bids.json
@@ -156,18 +156,10 @@
         },
         {
             "dataType": "TB1map",
-            "modalityLabel": "TB1map_mag",
+            "modalityLabel": "TB1map",
             "criteria": {
                 "SeriesDescription": "standard",
                 "ImageComments": "flip_angle_map*"
-            }
-        },
-        {
-            "dataType": "TB1map",
-            "modalityLabel": "TB1map_phase",
-            "criteria": {
-                "SeriesDescription": "standard",
-                "ImageComments": "*phase*"
             }
         }
     ]
--- a/config/dcm2bids.json
+++ b/config/dcm2bids.json
@@ -155,16 +155,16 @@
             }
         },
         {
-            "dataType": "rfmap",
-            "modalityLabel": "rfmap_mag",
+            "dataType": "TB1map",
+            "modalityLabel": "TB1map_mag",
             "criteria": {
                 "SeriesDescription": "standard",
                 "ImageComments": "flip_angle_map*"
             }
         },
         {
-            "dataType": "rfmap",
-            "modalityLabel": "rfmap_phase",
+            "dataType": "TB1map",
+            "modalityLabel": "TB1map_phase",
             "criteria": {
                 "SeriesDescription": "standard",
                 "ImageComments": "*phase*"
--- a/config/dcm2bids.json
+++ b/config/dcm2bids.json
@@ -153,6 +153,22 @@
                 "ImageType": ["*", "*", "P", "*", "*"],
                 "EchoNumber": "6"
             }
+        },
+        {
+            "dataType": "rfmap",
+            "modalityLabel": "rfmap_mag",
+            "criteria": {
+                "SeriesDescription": "standard",
+                "ImageComments": "flip_angle_map*"
+            }
+        },
+        {
+            "dataType": "rfmap",
+            "modalityLabel": "rfmap_phase",
+            "criteria": {
+                "SeriesDescription": "standard",
+                "ImageComments": "*phase*"
+            }
         }
     ]
 }
--- a/shimmingtoolbox/dicom_to_nifti.py
+++ b/shimmingtoolbox/dicom_to_nifti.py
@@ -9,32 +9,41 @@ from shimmingtoolbox import __dir_config_dcm2bids__
 
 
 def dicom_to_nifti(path_dicom, path_nifti, subject_id='sub-01', path_config_dcm2bids=__dir_config_dcm2bids__,
-                   special_dicom = False):
+                   special_dicom=[]):
     """ Converts dicom files into nifti files by calling dcm2bids
 
     Args:
         path_dicom (str): path to the input dicom folder
         path_nifti (str): path to the output nifti folder
-
+        subject_id (str): ID of subject
+        path_config_dcm2bids (str): Path to dcm2bids.json file
+        special_dicom (:obj:`list` of :obj:`str`): List of modalities for which the output NIFTI from dcm2niix should
+        be half-split along the last dimension.
     """
-    # Case where the DICOM files have
-    if special_dicom:
-        with tempfile.TemporaryDirectory(prefix='st_mag_') as tmp_mag:
-            dicom_number = len(os.listdir(path_dicom))
-            dicom_list = os.listdir(path_dicom)
-            for i in range(int(dicom_number/2)):
-                shutil.copy(os.path.join(path_dicom, dicom_list[i]), tmp_mag)
-            cmd = f'dcm2bids -d {tmp_mag} -p {subject_id} -o {path_nifti} -l DEBUG -c {path_config_dcm2bids}'
-            subprocess.run(cmd.split(' '))
+    # # Case where the DICOM files have
+    # if special_dicom:
+    #     with tempfile.TemporaryDirectory(prefix='st_mag_') as tmp_mag:
+    #         dicom_number = len(os.listdir(path_dicom))
+    #         dicom_list = os.listdir(path_dicom)
+    #         for i in range(int(dicom_number/2)):
+    #             shutil.copy(os.path.join(path_dicom, dicom_list[i]), tmp_mag)
+    #         cmd = f'dcm2bids -d {tmp_mag} -p {subject_id} -o {path_nifti} -l DEBUG -c {path_config_dcm2bids}'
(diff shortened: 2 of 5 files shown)


config/dcm2bids.json
```diff
@@ -153,6 +153,22 @@
                 "ImageType": ["*", "*", "P", "*", "*"],
                 "EchoNumber": "6"
             }
+        },
+        {
+            "dataType": "rfmap",
+            "modalityLabel": "rfmap_mag",
```
jcohenadad, line 159: can you check what is the current BIDS convention for naming B1 maps?

config/dcm2bids.json
```diff
@@ -153,6 +153,22 @@
...
+                "SeriesDescription": "standard",
+                "ImageComments": "flip_angle_map*"
+            }
+        },
+        {
+            "dataType": "rfmap",
+            "modalityLabel": "rfmap_phase",
+            "criteria": {
+                "SeriesDescription": "standard",
+                "ImageComments": "*phase*"
```
jcohenadad, line 170: that seems too inclusive, but i agree until we have other data to play with (after imposing conventions on the console), that is probably our only option to filter via dcm2bids. 

alternatively, you could do the filtering _after_ the conversion https://github.com/shimming-toolbox/shimming-toolbox-py/issues/58#issuecomment-670171854. I would lean towards the latter approach, given that it provides more flexibility. E.g., what if a researcher uses an RF map sequence with `"SeriesDescription": "myAwesomeRFsequence"`? the current approach would require updating this json file *and* assume the mag/phase data are half-split.

shimmingtoolbox/dicom_to_nifti.py
```diff
@@ -1,20 +1,42 @@
...
     Args:
         path_dicom (str): path to the input dicom folder
         path_nifti (str): path to the output nifti folder
 
     """
-    cmd = 'dcm2bids -d {} -p {} -o {} -l DEBUG -c {}'.format(
-        path_dicom, subject_id, path_nifti, path_config_dcm2bids)
-    subprocess.run(cmd.split(' '))
+    if special_dicom:
+        with tempfile.TemporaryDirectory(prefix='st_mag_') as tmp_mag:
```
gab-berest, line 21: As @jcohenadad  mentionned, copying a temporary file is not really optimized and should be rethink so it can be avoided. Also, for big files, copying can crash the OS if the computer is not powerfull enough. So it is not recommended.

shimmingtoolbox/dicom_to_nifti.py
```diff
@@ -1,20 +1,42 @@
...
+            dicom_number = len(os.listdir(path_dicom))
+            dicom_list = os.listdir(path_dicom)
+            for i in range(int(dicom_number/2)):
+                shutil.copy(os.path.join(path_dicom, dicom_list[i]), tmp_mag)
+            cmd = 'dcm2bids -d {} -p {} -o {} -l DEBUG -c {}'.format(
```
po09i, line 26: You should try f-strings, they are awesome. See this [comment](https://github.com/shimming-toolbox/shimming-toolbox-py/pull/46#discussion_r465966701) for details.

config/dcm2bids.json
gaspardcereza: @jcohenadad there's nothing on the [official specification](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/01-magnetic-resonance-imaging-data.html) right now but [here](https://bep001.readthedocs.io/en/dev/04-modality-specific-files/01-magnetic-resonance-imaging-data.html)'s a link that Agah shared with me some weeks ago where it is specified that RF field mapping maps should be named `TB1map`. I guess I can go with that for now event if it is not a definitive edition.

config/dcm2bids.json
gaspardcereza: I like the idea of separating them after the conversion but will I be able to modify the metadata of the 2 `.nii.gz` files to specify which one is the magnitude and which one is the phase ? Or is it only possible to specify it in the `json` sidecar ?

shimmingtoolbox/dicom_to_nifti.py
gaspardcereza: Indeed that's a really cool feature. Thanks for the tip !

config/dcm2bids.json
jcohenadad: > will I be able to modify the metadata of the 2 .nii.gz files to specify which one is the magnitude and which one is the phase

with your approach (splitting the dicoms and then convert), are the metadata of the niftis different? what field in the nifti metadata are you referring to?

gaspardcereza: I'm referring to the field that correspond to "ImageComments". For example, when I load my `NIfTI` files with `nibabel.load()`, I get a `object_name.header['aux_file']` field where the content is different for the phase ('relative_phase_map') and magnitude ('flip_angle_map_TraRefAm'). That's because `dcm2niix` seems to take the metadata of the first dicom file in the folder and applies it to the final nifti so when running it on a single folder everything is merged under one type of metadata.

jcohenadad: > I'm referring to the field that correspond to "ImageComments". For example, when I load my NIfTI files with nibabel.load(), I get a object_name.header['aux_file'] field where the content is different for the phase ('relative_phase_map') and magnitude ('flip_angle_map_TraRefAm'). That's because dcm2niix seems to take the metadata of the first dicom file in the folder and applies it to the final nifti so when running it on a single folder everything is merged under one type of metadata.

i see. that's a problem of the "convert everything to nifti and then split" approach. However, given that it only concerns isolated cases and that this will (hopefully) be solved by proper dicom output from the scanner, maybe we should still go ahead with the splitting of nifti _after_ the dicom conversion, and edition of the json file to update the fields that will be useful for RF shimming down the line. What do you think?

gaspardcereza: Yes as you said we might probably find a way to get proper dicoms on the scanner so we can go with that solution for now.

test/test_dicom_to_nifti.py
```diff
@@ -8,16 +8,35 @@
...
         subject_id = 'sub-test'
-        dicom_to_nifti(__dir_testing__, path_nifti, subject_id=subject_id)
+        # Conversion of the B0 mapping dicoms
+        dicom_to_nifti(os.path.join(__dir_testing__, 'dicom_unsorted'), path_nifti, subject_id=subject_id)
+
         # Check that all the files (.nii.gz and .json) are created with the expected names. The test data has 6
         # magnitude and phase data.
         for i in range(1, 7):
             for modality in ['phase', 'magnitude']:
                 for ext in ['nii.gz', 'json']:
```
gab-berest, line 23: It is not really a good way to have more than two nested for loops. You should use itertools that helps to have only one for loop:
```suggestion
import itertools

        # Check that all the files (.nii.gz and .json) are created with the expected names. The test data has 6
        # magnitude and phase data.
        for i, modality, ext in itertools.product(range(1,7), ['pahse','magnitude'], ['nii.gz','json']:
```

gab-berest:
@gaspardcereza Is this PR still alive?

gaspardcereza:
> @gaspardcereza Is this PR still alive?

Yes I've just been focusing on other things lately but I need to focus back on that PR, especially now that you modified the load_nifti function.

gaspardcereza:
I close this PR because a lot of things have evolved since I opened it and it will be easier to restart it from scratch.

po09i:
Do you need the branch for future work or we can delete the branch? @gaspardcereza

PR #74, opened 2020-08-24
Create synthetic coil-profiles

po09i:
## Context

Coil profiles are required for shimming to know how the field will change when current is applied to a channel. Siemens coils are designed to generate SH fields. Using spherical harmonics generation can be used to create synthetic coil profiles as they simulate the Siemens' shimming coils.

## Proposed changes

This PR introduces a function that creates spherical harmonics and reorders/scales them to siemens specifications (X, Y, Z, Z2, ZX, ZY, X2-Y2, XY). 

- [x] siemens_basis(x, y, z) (1st to 2nd order)
  - [x] spherical_harmonics(orders, x, y, z)

Future PRs
siemens_basis(orders, x, y, z) (3rd order) #76 

## Rendered documentation

➤ https://shimming-toolbox.org/en/ad-coil-profiles/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ testing_data
 # Pycharm stuff
 *.pyc
 .idea
+venv
 
 # Sphinx stuff
 docs/build/
--- a/docs/source/6_api_reference/api.rst
+++ b/docs/source/6_api_reference/api.rst
@@ -1,8 +1,38 @@
-
 API Reference
 =============
 
 The following section outlines the API of shimming-toolbox.
+
+
+unwrap
+------
+
+.. automodule:: shimmingtoolbox.unwrap.prelude
+   :members:
+
+masking
+-------
+
+.. automodule:: shimmingtoolbox.masking.shapes
+    :members:
+
+coils
+_____
+
+.. automodule:: shimmingtoolbox.coils.siemens_basis
+    :members:
+
+.. automodule:: shimmingtoolbox.coils.spherical_harmonics
+    :members:
+
+numerical_model
+---------------
+
+.. automodule:: shimmingtoolbox.simulate.numerical_model
+   :members:
+
+misc
+____
 
 .. automodule:: shimmingtoolbox.download
    :members:
@@ -18,21 +48,3 @@ The following section outlines the API of shimming-toolbox.
 
 .. automodule:: shimmingtoolbox.cli.referencemaps
    :members:
-
-numerical_model
----------------
-
-.. automodule:: shimmingtoolbox.simulate.numerical_model
-   :members:
-
-unwrap
-------
-
-.. automodule:: shimmingtoolbox.unwrap.prelude
-   :members:
-
-masking
--------
-
-.. automodule:: shimmingtoolbox.masking.shapes
-    :members:
--- a/docs/source/6_api_reference/api.rst
+++ b/docs/source/6_api_reference/api.rst
@@ -19,9 +19,13 @@ masking
 coils
 -----
 
+siemens_basis
+_____________
 .. automodule:: shimmingtoolbox.coils.siemens_basis
     :members:
 
+spherical_harmonics
+___________________
 .. automodule:: shimmingtoolbox.coils.spherical_harmonics
     :members:
 
--- a/docs/source/6_api_reference/api.rst
+++ b/docs/source/6_api_reference/api.rst
@@ -17,7 +17,7 @@ masking
     :members:
 
 coils
-_____
+-----
 
 .. automodule:: shimmingtoolbox.coils.siemens_basis
     :members:
@@ -32,7 +32,7 @@ numerical_model
    :members:
 
 misc
-____
+----
 
 .. automodule:: shimmingtoolbox.download
    :members:
--- a/shimmingtoolbox/coils/siemens_basis.py
+++ b/shimmingtoolbox/coils/siemens_basis.py
@@ -125,7 +125,8 @@ def siemens_basis(x, y, z):
 
         for i_ch in range(0, n_channels):
             field = sh[:, :, :, i_ch]
-            scaling_factors[i_ch] = 42.576 * ((r[i_ch] * 0.001) ** orders[i_ch]) / field[i_ref[i_ch]]
+            gyro_ratio = 42.576
+            scaling_factors[i_ch] = gyro_ratio * ((r[i_ch] * 0.001) ** orders[i_ch]) / field[i_ref[i_ch]]
 
         return scaling_factors
 
--- a/shimmingtoolbox/coils/siemens_basis.py
+++ b/shimmingtoolbox/coils/siemens_basis.py
@@ -88,8 +88,7 @@ def siemens_basis(x, y, z):
 
         NOTE: The method has been worked out empirically and has only been tested for 2 Siemens Prisma systems.
         E.g.re: Y, Z, ZX, and XY terms, their polarity had to be flipped relative to the form given by
-        shimmingtoolbox.coils.spherical_harmonics. To adapt the code to other systems, arbitrary(?) changes along these
-        lines will likely be needed.
+        shimmingtoolbox.coils.spherical_harmonics.
         """
 
         [x_iso, y_iso, z_iso] = np.meshgrid(np.array(range(-1, 2)), np.array(range(-1, 2)), np.array(range(-1, 2)),
--- a/shimmingtoolbox/coils/siemens_basis.py
+++ b/shimmingtoolbox/coils/siemens_basis.py
@@ -7,8 +7,8 @@ from shimmingtoolbox.coils.spherical_harmonics import spherical_harmonics
 
 def siemens_basis(x, y, z):
     """
-    The function first wraps `shimmingtoolbox.coils.spherical_harmonics` to generate 1st and 2nd order spherical
-    harmonic `basis` fields at the grid positions given by arrays `X,Y,Z`. *Following Siemens convention*, `basis` is
+    The function first wraps ``shimmingtoolbox.coils.spherical_harmonics`` to generate 1st and 2nd order spherical
+    harmonic ``basis`` fields at the grid positions given by arrays ``X,Y,Z``. *Following Siemens convention*, ``basis`` is
     then:
 
         - Reordered along the 4th dimension as *X, Y, Z, Z2, ZX, ZY, X2-Y2, XY*
@@ -19,11 +19,11 @@ def siemens_basis(x, y, z):
             - 1 micro-T/m for *X,Y,Z* gradients (= 0.042576 Hz/mm)
             - 1 micro-T/m^2 for 2nd order terms (= 0.000042576 Hz/mm^2)
 
-    The returned `basis` is thereby in the form of ideal "shim reference maps", ready for optimization.
+    The returned ``basis`` is thereby in the form of ideal "shim reference maps", ready for optimization.
 
     Args:
-        orders (numpy.ndarray): *not yet implemented* Degrees of the desired terms in the series expansion, specified
-                                as a vector of non-negative integers (`[0:1:n]` yields harmonics up to n-th order)
+        orders (numpy.ndarray): **not yet implemented** Degrees of the desired terms in the series expansion, specified
+                                as a vector of non-negative integers (``[0:1:n]`` yields harmonics up to n-th order)
         x (numpy.ndarray): 3-D arrays of grid coordinates, "Right->Left" grid coordinates in the patient coordinate
                            system (i.e. DICOM reference, units of mm)
         y (numpy.ndarray): 3-D arrays of grid coordinates (same shape as x). "Anterior->Posterior" grid coordinates in
@@ -36,18 +36,18 @@ def siemens_basis(x, y, z):
         numpy.ndarray: 4-D array of spherical harmonic basis fields
 
     NOTES:
-        For now, `orders` is, in fact, ignored: fixed as [1:2]—which is suitable for the Prisma (presumably other
+        For now, ``orders`` is, in fact, ignored: fixed as [1:2]—which is suitable for the Prisma (presumably other
         Siemens systems as well) however, the 3rd-order shims of the Terra should ultimately be accommodated too.
         (Requires checking the Adjustments/Shim card to see what the corresponding terms and values actually are). So,
-        for now, `basis` will always be returned with 8 terms along the 4th dim.
+        for now, ``basis`` will always be returned with 8 terms along the 4th dim.
     """
 
     # Local functions
     def reorder_to_siemens(spher_harm):
         """
         Reorder 1st - 2nd order basis terms along 4th dim. From
-        1. Y, Z, X, XY, ZY, Z2, ZX, X2 - Y2(output by shimmingtoolbox.coils.spherical_harmonics.spherical_harmonics), to
-        2. X, Y, Z, Z2, ZX, ZY, X2 - Y2, XY( in line with Siemens shims)
+        1. Y, Z, X, XY, ZY, Z2, ZX, X2 - Y2 (output by shimmingtoolbox.coils.spherical_harmonics.spherical_harmonics), to
+        2. X, Y, Z, Z2, ZX, ZY, X2 - Y2, XY (in line with Siemens shims)
         """
 
         assert spher_harm.shape[3] == 8
@@ -65,7 +65,7 @@ def siemens_basis(x, y, z):
 
     def get_scaling_factors():
         """
-        Returns a vector of `scaling_factors` to apply to the(Siemens-reordered) 1st + 2nd order spherical harmonic fields
+        Returns a vector of ``scaling_factors`` to apply to the(Siemens-reordered) 1st + 2nd order spherical harmonic fields
         for rescaling field terms as "shim reference maps" in units of Hz/unit-shim:
 
         Gx, Gy, and Gz should yield 1 micro-T of field shift per metre: equivalently, 0.042576 Hz/mm
@@ -76,7 +76,7 @@ def siemens_basis(x, y, z):
         at which we know what the field *should* be, and use that to calculate the appropriate scaling factor.
 
         NOTE: The method has been worked out empirically and has only been tested for 2 Siemens Prisma systems.
-        E.g.re: Y, Z, ZX, and XY terms, it was noted that their polarity had to be flipped relative to the form given by
+        E.g.re: Y, Z, ZX, and XY terms, their polarity had to be flipped relative to the form given by
         shimmingtoolbox.coils.spherical_harmonics. To adapt the code to other systems, arbitrary(?) changes along these
         lines will likely be needed.
         """
@@ -120,10 +120,10 @@ def siemens_basis(x, y, z):
 
     # Check inputs
     if not (x.ndim == y.ndim == z.ndim == 3):
-        raise RuntimeError('Input arrays X, Y, and Z must be 3d')
+        raise RuntimeError("Input arrays X, Y, and Z must be 3d")
 
     if not (x.shape == y.shape == z.shape):
-        raise RuntimeError('Input arrays X, Y, and Z must be identically sized')
+        raise RuntimeError("Input arrays X, Y, and Z must be identically sized")
 
     # Create spherical harmonics from first to second order
     orders = np.array(range(1, 3))
--- a/shimmingtoolbox/coils/siemens_basis.py
+++ b/shimmingtoolbox/coils/siemens_basis.py
@@ -1,0 +1,144 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*
+
+import numpy as np
+from shimmingtoolbox.coils.spherical_harmonics import spherical_harmonics
+
+
+def reorder_to_siemens(spher_harm):
+    """
+    Reorder 1st - 2nd order basis terms along 4th dim. From
+    1. Y, Z, X, XY, ZY, Z2, ZX, X2 - Y2(output by shimmingtoolbox.coils.spherical_harmonics.sppherical_harmonics), to 
+    2. X, Y, Z, Z2, ZX, ZY, X2 - Y2, XY( in line with Siemens shims)
+    """
+
+    assert spher_harm.shape[3] == 8
+    reordered = np.zeros(spher_harm.shape)
+    reordered[:, :, :, 0] = spher_harm[:, :, :, 2]
+    reordered[:, :, :, 1] = spher_harm[:, :, :, 0]
+    reordered[:, :, :, 2] = spher_harm[:, :, :, 1]
+    reordered[:, :, :, 3] = spher_harm[:, :, :, 5]
+    reordered[:, :, :, 4] = spher_harm[:, :, :, 6]
+    reordered[:, :, :, 5] = spher_harm[:, :, :, 4]
+    reordered[:, :, :, 6] = spher_harm[:, :, :, 7]
+    reordered[:, :, :, 7] = spher_harm[:, :, :, 3]
+
+    return reordered
+
+
+def scale(spher_harm):
+    """
+    Returns a vector of `scaling_factors` to apply to the(Siemens-reordered) 1st + 2nd order spherical harmonic fields
+    for rescaling field terms as "shim reference maps" in units of Hz/unit-shim:
+
+    Gx, Gy, and Gz should yield 1 micro-T of field shift per metre: equivalently, 0.042576 Hz/mm
+
+    2nd order terms should yield 1 micro-T of field shift per metre-squared: equivalently, 0.000042576 Hz/mm^2
+
+    Gist: given the stated nominal values, we can pick several arbitrary reference positions around the origin/isocenter
+    at which we know what the field *should* be, and use that to calculate the appropriate scaling factor.
+
+    NOTE: The method has been worked out empirically and has only been tested for 2 Siemens Prisma systems.
+    E.g.re: Y, Z, ZX, and XY terms, it was noted that their polarity had to be flipped relative to the form given by
+    shimmingtoolbox.coils.spherical_harmonics. To adapt the code to other systems, arbitrary(?) changes along these
+    lines will likely be needed.
+    """
+
+    [x_iso, y_iso, z_iso] = np.meshgrid(np.array(range(-1, 2)), np.array(range(-1, 2)), np.array(range(-1, 2)),
+                                        indexing='xy')
+    orders = np.array(range(1, 3))
+    sh = spherical_harmonics(orders, x_iso, y_iso, z_iso)
+    sh = reorder_to_siemens(sh)
+
+    n_channels = sh.shape[3]
+    scaling_factors = np.zeros(n_channels)
+
+    # indices of reference positions for normalization:
+    i_x1 = np.nonzero((x_iso == 1) & (y_iso == 0) & (z_iso == 0))
+    i_y1 = np.nonzero((x_iso == 0) & (y_iso == 1) & (z_iso == 0))
+    i_z1 = np.nonzero((x_iso == 0) & (y_iso == 0) & (z_iso == 1))
+
+    i_x1z1 = np.nonzero((x_iso == 1) & (y_iso == 0) & (z_iso == 1))
+    i_y1z1 = np.nonzero((x_iso == 0) & (y_iso == 1) & (z_iso == 1))
+    i_x1y1 = np.nonzero((x_iso == 1) & (y_iso == 1) & (z_iso == 0))
+
+    # order the reference indices like the sh field terms
+    i_ref = [i_x1, i_y1, i_z1, i_z1, i_x1z1, i_y1z1, i_x1, i_x1y1]
+
+    # distance from iso/origin to adopted reference point[units: mm]
+    r = [1, 1, 1, 1, np.sqrt(2), np.sqrt(2), 1, np.sqrt(2)]
+
+    # invert polarity
+    # Y, Z, ZX, and XY terms only(determined empirically)
+    sh[:, :, :, [1, 2, 4, 7]] = -sh[:, :, :, [1, 2, 4, 7]]
+
+    # scaling:
+    orders = [1, 1, 1, 2, 2, 2, 2, 2]
+
+    for i_ch in range(0, n_channels):
+        field = sh[:, :, :, i_ch]
+        scaling_factors[i_ch] = 42.576 * ((r[i_ch] * 0.001) ** orders[i_ch]) / field[i_ref[i_ch]]
+
+    return scaling_factors
+
+
+def siemens_basis(x, y, z):
+    """
+    The function first wraps `shimmingtoolbox.coils.spherical_harmonics` to generate 1st and 2nd order spherical
+    harmonic `basis` fields at the grid positions given by arrays `X,Y,Z`. *Following Siemens convention*, `basis` is
+    then:
+
+    - Reordered along the 4th dimension as *X, Y, Z, Z2, ZX, ZY, X2-Y2, XY*
+
+    - Rescaled to Hz/unit-shim, where "unit-shim" refers to the measure displayed in the Adjustments card of the Syngo
+    console UI, namely:
+
+        - 1 micro-T/m for *X,Y,Z* gradients (= 0.042576 Hz/mm)
+        - 1 micro-T/m^2 for 2nd order terms (= 0.000042576 Hz/mm^2)
+
+    The returned `basis` is thereby in the form of ideal "shim reference maps", ready for optimization.
+
+    Args:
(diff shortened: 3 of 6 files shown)


shimmingtoolbox/coils/siemens_basis.py
```diff
@@ -0,0 +1,143 @@
...
+
+    NOTES:
+        For now, `orders` is, in fact, ignored: fixed as [1:2]—which is suitable for the Prisma (presumably other
+        Siemens systems as well) however, the 3rd-order shims of the Terra should ultimately be accommodated too.
+        (Requires checking the Adjustments/Shim card to see what the corresponding terms and values actually are). So,
+        for now, `basis` will always be returned with 8 terms along the 4th dim.
+    """
+
+    # Local functions
+    def reorder_to_siemens(spher_harm):
```
jcohenadad, line 46: docstring to add (what input type is `spher_harm`?)

shimmingtoolbox/coils/spherical_harmonics.py
```diff
@@ -0,0 +1,136 @@
...
+        cos_theta = np.cos(np.arctan2(np.sqrt(pos_x ** 2 + pos_y ** 2), pos_z))
+        # cos_theta = pos_z / r
+
+        if m >= 0:
+            c = 1
+        else:
+            c = 0
+            m = -m
+
+        # y_mn = leg_rec(n, m, cos_theta)  # Does the same as lpmv function (Will remove after more tests)
```
jidicula, line 83: leftover commented-out code that was to be removed?

shimmingtoolbox/coils/spherical_harmonics.py
```diff
@@ -0,0 +1,136 @@
...
+        Returns harmonic field for the required solid harmonic addressed by n, m based on the Legendre
+        polynomial calculation. Positive m values correspond to the cosine component, negative to the sine. Returned
+        fields will eventually follow RRI s convention
+        """
+
+        r2 = pos_x ** 2 + pos_y ** 2 + pos_z ** 2
+        r = np.sqrt(r2)
+        phi = np.arctan2(pos_y, pos_x)
+        cos_theta = np.cos(np.arctan2(np.sqrt(pos_x ** 2 + pos_y ** 2), pos_z))
+        # cos_theta = pos_z / r
```
jidicula, line 75: old code? does it need to be removed?

shimmingtoolbox/coils/spherical_harmonics.py
```diff
@@ -0,0 +1,136 @@
...
+        n = orders[iOrder]
+        if n == 0:
+            m = np.zeros([1], dtype=int)
+        else:
+            m = np.array(range(-orders[iOrder], orders[iOrder] + 1, 1))
+
+        for mm in range(0, m.size):
+
+            # The first 2 * n(1) + 1 columns of the output correspond to harmonics of order n(1), and the next
+            # 2 * n(2) + 1 columns correspond to harmonics of order n(2), etc.
```
jidicula, line 124: These are good clarifying comments, especially for contributors like me that have no background in MRI physics 🙂

test/test_spherical_harmonics.py
```diff
@@ -0,0 +1,52 @@
...
+
+@pytest.mark.parametrize('x,y,z', dummy_data)
+def test_wrong_input_dimension(x, y, z):
+    orders = np.array(range(0, 2, 1))
+
+    # Call spherical harmonics with wrong dimension
+    try:
+        spherical_harmonics(orders, x[:, :, 0], y, z)
+    except RuntimeError:
+        # If an exception occurs, this is the desired behaviour
```
jidicula, line 31: good comments for clarifying expected failure test cases

.gitignore
```diff
@@ -12,6 +12,7 @@ testing_data
 # Pycharm stuff
 *.pyc
 .idea
+venv
```
jidicula, line 15: see comment about housekeeping scope

shimmingtoolbox/coils/siemens_basis.py
```diff
@@ -0,0 +1,143 @@
...
+
+        - Reordered along the 4th dimension as *X, Y, Z, Z2, ZX, ZY, X2-Y2, XY*
+
+        - Rescaled to Hz/unit-shim, where "unit-shim" refers to the measure displayed in the Adjustments card of the
+          Syngo console UI, namely:
+
+            - 1 micro-T/m for *X,Y,Z* gradients (= 0.042576 Hz/mm)
+            - 1 micro-T/m^2 for 2nd order terms (= 0.000042576 Hz/mm^2)
+
+    The returned `basis` is thereby in the form of ideal "shim reference maps", ready for optimization.
```
jidicula, line 22:
```suggestion
    The returned ``basis`` is thereby in the form of ideal "shim reference maps", ready for optimization.
```

shimmingtoolbox/coils/siemens_basis.py
```diff
@@ -0,0 +1,143 @@
...
+        (Requires checking the Adjustments/Shim card to see what the corresponding terms and values actually are). So,
+        for now, `basis` will always be returned with 8 terms along the 4th dim.
+    """
+
+    # Local functions
+    def reorder_to_siemens(spher_harm):
+        """
+        Reorder 1st - 2nd order basis terms along 4th dim. From
+        1. Y, Z, X, XY, ZY, Z2, ZX, X2 - Y2(output by shimmingtoolbox.coils.spherical_harmonics.spherical_harmonics), to
+        2. X, Y, Z, Z2, ZX, ZY, X2 - Y2, XY( in line with Siemens shims)
```
jidicula, line 50:
```suggestion
        1. Y, Z, X, XY, ZY, Z2, ZX, X2 - Y2 (output by shimmingtoolbox.coils.spherical_harmonics.spherical_harmonics), to
        2. X, Y, Z, Z2, ZX, ZY, X2 - Y2, XY (in line with Siemens shims)
```

jcohenadad:
Would be cool to create sub-entries for the coil heading on RTD (siemens_basis, spherical_harmonics, etc.):

image

shimmingtoolbox/coils/siemens_basis.py
```diff
@@ -0,0 +1,143 @@
...
+
+        Gx, Gy, and Gz should yield 1 micro-T of field shift per metre: equivalently, 0.042576 Hz/mm
+
+        2nd order terms should yield 1 micro-T of field shift per metre-squared: equivalently, 0.000042576 Hz/mm^2
+
+        Gist: given the stated nominal values, we can pick several arbitrary reference positions around the origin/isocenter
+        at which we know what the field *should* be, and use that to calculate the appropriate scaling factor.
+
+        NOTE: The method has been worked out empirically and has only been tested for 2 Siemens Prisma systems.
+        E.g.re: Y, Z, ZX, and XY terms, it was noted that their polarity had to be flipped relative to the form given by
```
jidicula, line 79: could take the active tense here to make it less wordy:
```suggestion
        E.g.re: Y, Z, ZX, and XY terms, their polarity had to be flipped relative to the form given by
```

jcohenadad:
@po09i i've added a "Rendered documentation" in your PR body with a link to the rendered RTD-- send me your RTD username so i can add you to the RTD admin team and you can select the build yourself to include in future PRs

shimmingtoolbox/coils/siemens_basis.py
```diff
@@ -0,0 +1,143 @@
...
+        # invert polarity
+        # Y, Z, ZX, and XY terms only(determined empirically)
+        sh[:, :, :, [1, 2, 4, 7]] = -sh[:, :, :, [1, 2, 4, 7]]
+
+        # scaling:
+        orders = [1, 1, 1, 2, 2, 2, 2, 2]
+
+        for i_ch in range(0, n_channels):
+            field = sh[:, :, :, i_ch]
+            scaling_factors[i_ch] = 42.576 * ((r[i_ch] * 0.001) ** orders[i_ch]) / field[i_ref[i_ch]]
```
jidicula, line 117: I noticed

> Gx, Gy, and Gz should yield 1 micro-T of field shift per metre: equivalently, 0.042576 Hz/mm
> 2nd order terms should yield 1 micro-T of field shift per metre-squared: equivalently, 0.000042576 Hz/mm^2

in the docstring. Which of these numbers is referred to by `42.576` in this line? Perhaps you could include a comment mentioning which value in the docstring is relevant here? My misunderstanding is likely because I'm new to the MRI world, so disregard if this is obvious to someone who has more background 😅

shimmingtoolbox/coils/siemens_basis.py
```diff
@@ -0,0 +1,143 @@
...
+
+        return scaling_factors
+
+    # Check inputs
+    if not (x.ndim == y.ndim == z.ndim == 3):
+        raise RuntimeError('Input arrays X, Y, and Z must be 3d')
+
+    if not (x.shape == y.shape == z.shape):
+        raise RuntimeError('Input arrays X, Y, and Z must be identically sized')
```
jidicula, line 126: I think this might be an appropriate place for [double quotes](https://github.com/neuropoly/spinalcordtoolbox/wiki/Single-or-double-quotes).
```suggestion
    if not (x.ndim == y.ndim == z.ndim == 3):
        raise RuntimeError("Input arrays X, Y, and Z must be 3d")

    if not (x.shape == y.shape == z.shape):
        raise RuntimeError("Input arrays X, Y, and Z must be identically sized")
```

shimmingtoolbox/coils/spherical_harmonics.py
```diff
@@ -0,0 +1,136 @@
...
+            - calc_spherical_harmonics_arb_points_cz.m by jaystock@nmr.mgh.harvard.edu
+    """
+
+    def number_of_coef(orders):
+        out = 0
+        for n in orders:
+            m = 2 * n + 1
+            out += m
+
+        return out
```
jidicula, line 62: I'd recommend a clearer variable name than `out`, reflecting what the output is:
```suggestion
        n_coef = 0
        for n in orders:
            m = 2 * n + 1
            n_coef += m

        return n_coef
```

shimmingtoolbox/coils/spherical_harmonics.py
```diff
@@ -0,0 +1,136 @@
...
+            m = -m
+
+        # y_mn = leg_rec(n, m, cos_theta)  # Does the same as lpmv function (Will remove after more tests)
+        y_mn = lpmv(m, n, cos_theta)
+
+        rri_norm = math.factorial(n + m + 1) / math.factorial(n - m) / factorial2(2 * m)
+
+        out = (n + m + 1) * (r ** n) * (np.cos(m * phi) * c + np.sin(m * phi) * (1 - c)) * y_mn / rri_norm
+
+        return out
```
jidicula, line 90: see comment about var names above

```suggestion
        harmonic_field = (n + m + 1) * (r ** n) * (np.cos(m * phi) * c + np.sin(m * phi) * (1 - c)) * y_mn / rri_norm

        return harmonic_field
```

shimmingtoolbox/coils/spherical_harmonics.py
```diff
@@ -0,0 +1,136 @@
...
+    else:
+        raise RuntimeError('Input arrays X, Y, and Z must have 3 dimensions')
+
+    if not np.all(orders >= 0):
+        raise RuntimeError('Orders must be positive')
+
+    # Initialize variables
+    n_voxels = x.size
+    n_orders = orders.size
+    n_basis = number_of_coef(orders)
```
jidicula, line 107: these are clear names 🙂

test/test_siemens_basis.py
```diff
@@ -0,0 +1,23 @@
...
+
+@pytest.mark.parametrize('x,y,z', dummy_data)
+def test_normal_siemens_basis(x, y, z):
+
+    basis = siemens_basis(x, y, z)
+
+    # Test for shape
+    assert(np.all(basis.shape == (x.shape[0], x.shape[1], x.shape[2], 8)))
+    # Test for first value
+    assert(math.isclose(basis[0, 0, 0, 0], -4.25760000e-02, rel_tol=1e-09))
```
jidicula, line 23: `-4.25760000e-02` appears again... would it be useful to have a comment explaining why that literal is appearing in a test case?

test/test_spherical_harmonics.py
```diff
@@ -0,0 +1,52 @@
...
+        return 0
+
+    # If there isn't an error, then there is a problem
+    print('\nWrong dimensions for input x does not throw an error')
```
jidicula, line 35: double quotes
```suggestion
    print("\nWrong dimensions for input x does not throw an error")
```

test/test_spherical_harmonics.py
```diff
@@ -0,0 +1,52 @@
...
+
+    # Call spherical harmonics with negative order
+    try:
+        spherical_harmonics(orders, x, y, z)
+    except RuntimeError:
+        # If an exception occurs, this is the desired behaviour
+        return 0
+
+    # If there isn't an error, then there is a problem
+    print('\nNegative order does not throw an error')
```
jidicula, line 51: double quotes
```suggestion
    print("\nNegative order does not throw an error")
```

shimmingtoolbox/coils/siemens_basis.py
```diff
@@ -0,0 +1,143 @@
...
+
+        for i_ch in range(0, n_channels):
+            field = sh[:, :, :, i_ch]
+            scaling_factors[i_ch] = 42.576 * ((r[i_ch] * 0.001) ** orders[i_ch]) / field[i_ref[i_ch]]
+
+        return scaling_factors
+
+    # Check inputs
+    if not (x.ndim == y.ndim == z.ndim == 3):
+        raise RuntimeError('Input arrays X, Y, and Z must be 3d')
```
jidicula, line 123: double quotes
```suggestion
        raise RuntimeError("Input arrays X, Y, and Z must be 3d")
```

jidicula (changes requested):
Thanks for requesting my review!

I've suggested some minor changes to improve code quality for better reader understanding (and therefore long-term maintainability). Additionally, there's a comment in there about scoping of commits – these don't need to be changed this time around, but is definitely something to keep in mind moving forward.

In terms of the actual implementation, I'm trusting the expertise of other reviewers that the synthetic coil profiles are correct 😛

po09i:
> @po09i i've added a "Rendered documentation" in your PR body with a link to the rendered RTD-- send me your RTD username so i can add you to the RTD admin team and you can select the build yourself to include in future PRs

Should be po09i

docs/source/6_api_reference/api.rst
```diff
@@ -1,38 +1,50 @@
...
-.. automodule:: shimmingtoolbox.masking.shapes
-    :members:
+.. automodule:: shimmingtoolbox.cli.b0map
+   :members:
+
+.. automodule:: shimmingtoolbox.cli.download_data
+   :members:
+
+.. automodule:: shimmingtoolbox.cli.referencemaps
+   :members:
```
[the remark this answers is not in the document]
po09i: We should use the rebase option if it is available. That being said I agree with your suggestion I'll try to be more careful next time.

shimmingtoolbox/coils/siemens_basis.py
```diff
@@ -0,0 +1,143 @@
...
+        for now, `basis` will always be returned with 8 terms along the 4th dim.
+    """
+
+    # Local functions
+    def reorder_to_siemens(spher_harm):
+        """
+        Reorder 1st - 2nd order basis terms along 4th dim. From
+        1. Y, Z, X, XY, ZY, Z2, ZX, X2 - Y2(output by shimmingtoolbox.coils.spherical_harmonics.spherical_harmonics), to
+        2. X, Y, Z, Z2, ZX, ZY, X2 - Y2, XY( in line with Siemens shims)
+        """
```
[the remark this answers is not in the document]
po09i: Since they are "local functions" (functions declared within a function), they are not rendered in the documentation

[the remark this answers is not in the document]
jidicula: Great, thanks for clarifying 🙂

shimmingtoolbox/coils/siemens_basis.py
```diff
@@ -0,0 +1,143 @@
...
+
+        2nd order terms should yield 1 micro-T of field shift per metre-squared: equivalently, 0.000042576 Hz/mm^2
+
+        Gist: given the stated nominal values, we can pick several arbitrary reference positions around the origin/isocenter
+        at which we know what the field *should* be, and use that to calculate the appropriate scaling factor.
+
+        NOTE: The method has been worked out empirically and has only been tested for 2 Siemens Prisma systems.
+        E.g.re: Y, Z, ZX, and XY terms, it was noted that their polarity had to be flipped relative to the form given by
+        shimmingtoolbox.coils.spherical_harmonics. To adapt the code to other systems, arbitrary(?) changes along these
+        lines will likely be needed.
```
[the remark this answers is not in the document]
po09i: This is a problem for future development, I was planning to open an issue for the 3rd order spherical harmonics which arises when we switch to the 7T MRI. I'll mention it in the issue since it is a case where we change systems.

shimmingtoolbox/coils/spherical_harmonics.py
po09i: good catch

shimmingtoolbox/coils/siemens_basis.py
po09i: It is the gyromagnetic ratio (γ = 42.576 Mhz/T) in spin angular precession: ω = γB.

test/test_siemens_basis.py
po09i: I could check that the entire matrix is correct but it would require explicitly creating an array that is 2x2x2x8 (1x1x1x8 is not really useful) so I wrote the test to only look at `basis[0,0,0,0]` (it arbitrary happens to be 4.2576).

shimmingtoolbox/coils/siemens_basis.py
```diff
@@ -0,0 +1,154 @@
...
+        Gist: given the stated nominal values, we can pick several arbitrary reference positions around the
+        origin/isocenter at which we know what the field *should* be, and use that to calculate the appropriate scaling
+        factor.
+
+        Returns:
+             numpy.ndarray:  1D vector of ``scaling_factors``
+
+        NOTE: The method has been worked out empirically and has only been tested for 2 Siemens Prisma systems.
+        E.g.re: Y, Z, ZX, and XY terms, their polarity had to be flipped relative to the form given by
+        shimmingtoolbox.coils.spherical_harmonics.
```
jidicula, line 91:
```suggestion
        ``shimmingtoolbox.coils.spherical_harmonics``.
```

shimmingtoolbox/coils/spherical_harmonics.py
```diff
@@ -0,0 +1,134 @@
...
+
+        rri_norm = math.factorial(n + m + 1) / math.factorial(n - m) / factorial2(2 * m)
+
+        harmonic_field = (n + m + 1) * (r ** n) * (np.cos(m * phi) * c + np.sin(m * phi) * (1 - c)) * y_mn / rri_norm
+
+        return harmonic_field
+
+    # Check inputs
+    if not (x.shape == y.shape == z.shape):
+        raise RuntimeError('Input arrays X, Y, and Z must be identically sized')
```
jidicula, line 92: double quotes (missed these the first time around)

```suggestion
        raise RuntimeError("Input arrays X, Y, and Z must be identically sized")
```

shimmingtoolbox/coils/spherical_harmonics.py
```diff
@@ -0,0 +1,134 @@
...
+
+    if x.ndim == 3:
+        grid_size = x.shape
+    else:
+        raise RuntimeError('Input arrays X, Y, and Z must have 3 dimensions')
```
jidicula, line 97: double quotes 
```suggestion
        raise RuntimeError("Input arrays X, Y, and Z must have 3 dimensions")
```

shimmingtoolbox/coils/spherical_harmonics.py
```diff
@@ -0,0 +1,134 @@
...
+
+    if not np.all(orders >= 0):
+        raise RuntimeError('Orders must be positive')
```
jidicula, line 100: double quotes
```suggestion
        raise RuntimeError("Orders must be positive")
```

jidicula (approved):
A few instances of double quotes and double backticks that I missed on my first review, but otherwise looks good to merge!

test/test_siemens_basis.py
jcohenadad: i agree about adding a comment on what this -4.25... number comes from (i was wondering the same thing, and most people who will read this code will wonder the same thing, meaning: it deserves a comment)

po09i: done

shimmingtoolbox/coils/siemens_basis.py
gab-berest: A good practice would be to include the constants in a constant variable. Since the gyromagnetic ratio is constant, you could include it in a constant named GYROMAGNETIC_RATIO = 42.576 #MHz/T. Like this it would be easier for everyone even with less knowledge in MRI. Same thing for 0.001 which is a scaling factor for units. The only constants that should be used outside of constant variables are 0 and 1.

shimmingtoolbox/coils/siemens_basis.py
```diff
@@ -0,0 +1,154 @@
...
+        ``shimmingtoolbox.coils.spherical_harmonics``.
+        """
+
+        [x_iso, y_iso, z_iso] = np.meshgrid(np.array(range(-1, 2)), np.array(range(-1, 2)), np.array(range(-1, 2)),
+                                            indexing='xy')
+        orders = np.array(range(1, 3))
+        sh = spherical_harmonics(orders, x_iso, y_iso, z_iso)
+        sh = reorder_to_siemens(sh)
+
+        n_channels = sh.shape[3]
```
gab-berest, line 100: Is there a need to get the number of channels from the shape since it will always be 8 from reorder_to_siemens function?

shimmingtoolbox/coils/siemens_basis.py
```diff
@@ -0,0 +1,154 @@
...
+
+        # order the reference indices like the sh field terms
+        i_ref = [i_x1, i_y1, i_z1, i_z1, i_x1z1, i_y1z1, i_x1, i_x1y1]
+
+        # distance from iso/origin to adopted reference point[units: mm]
+        r = [1, 1, 1, 1, np.sqrt(2), np.sqrt(2), 1, np.sqrt(2)]
+
+        # invert polarity
+        # Y, Z, ZX, and XY terms only(determined empirically)
+        sh[:, :, :, [1, 2, 4, 7]] = -sh[:, :, :, [1, 2, 4, 7]]
```
gab-berest, line 120: A good programming convention is to put magical numbers (like 2,4,7) in constants with relevant ames so that you don't need to specify what they correspond to in comments. It will be directly explicit in the code. Like: `Y = 1`, `ZX = 4` and then you can just call sh[:, :, :, [Y, Z, ZX, XY]] = -sh[:, :, :, [Y, Z, ZX, XY]]. In this case, it is clear that you check the Y,Z, ZX, XY terms

shimmingtoolbox/coils/siemens_basis.py
```diff
@@ -0,0 +1,154 @@
...
+            spher_harm (numpy.ndarray): 4d basis set of spherical harmonics with order/degree ordered along 4th
+                                        dimension. ``spher_harm.shape[3]` must equal 8.
+
+        Returns:
+            numpy.ndarray: 4d basis set of spherical harmonics ordered following siemens convention
+        """
+
+        assert spher_harm.shape[3] == 8
+        reordered = np.zeros(spher_harm.shape)
+        reordered[:, :, :, 0] = spher_harm[:, :, :, 2]
```
gab-berest, line 62: Instead of doing index by index, you could make a one liner like this:
```suggestion
        reordered = spher_harm[:, :, :, [2, 0, 1, 5, 6, 4, 7, 3]]
```

shimmingtoolbox/coils/siemens_basis.py
```diff
@@ -0,0 +1,154 @@
...
+        # invert polarity
+        # Y, Z, ZX, and XY terms only(determined empirically)
+        sh[:, :, :, [1, 2, 4, 7]] = -sh[:, :, :, [1, 2, 4, 7]]
+
+        # scaling:
+        orders = [1, 1, 1, 2, 2, 2, 2, 2]
+
+        for i_ch in range(0, n_channels):
+            field = sh[:, :, :, i_ch]
+            gyro_ratio = 42.576
```
gab-berest, line 127: As said before, you should put the guromagnetic ratio in a file scope constant at the beginnign of the file like:
```suggestion
           GYROMAGNETIC_RATIO = 42.576
```

shimmingtoolbox/coils/siemens_basis.py
```diff
@@ -0,0 +1,154 @@
...
+            scaling_factors[i_ch] = gyro_ratio * ((r[i_ch] * 0.001) ** orders[i_ch]) / field[i_ref[i_ch]]
+
+        return scaling_factors
+
+    # Check inputs
+    if not (x.ndim == y.ndim == z.ndim == 3):
```
gab-berest, line 133: Why define functions in another function?

shimmingtoolbox/coils/spherical_harmonics.py
```diff
@@ -0,0 +1,134 @@
...
+            - 8: *x2y2*
+
+        Based on
+            - spherical_harmonics.m by topfer@ualberta.ca
+            - calc_spherical_harmonics_arb_points_cz.m by jaystock@nmr.mgh.harvard.edu
+    """
+
+    def number_of_coef(orders):
+        n_coef = 0
+        for n in orders:
```
gab-berest, line 58: Here, you could use the built-in function sum form python:
```suggestion
       return sum(2*orders+1)
```

shimmingtoolbox/coils/siemens_basis.py
```diff
@@ -0,0 +1,154 @@
...
+    # Create spherical harmonics from first to second order
+    orders = np.array(range(1, 3))
+    spher_harm = spherical_harmonics(orders, x, y, z)
+
+    # Reorder according to siemens convention: X, Y, Z, Z2, ZX, ZY, X2-Y2, XY
+    reordered_spher = reorder_to_siemens(spher_harm)
+
+    # scale according to
+    # - 1 micro-T/m for *X,Y,Z* gradients (= 0.042576 Hz/mm)
+    # - 1 micro-T/m^2 for 2nd order terms (= 0.000042576 Hz/mm^2)
```
lcbMGH, line 148: I need to look into the physical situation behind this, but is this indicating that different channels have different units, or is this scaling to account for that difference? What are the final units for the returned coil profiles -- Hz?

shimmingtoolbox/coils/spherical_harmonics.py
```diff
@@ -0,0 +1,134 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*
+
+import numpy as np
+import math
+from scipy.special import factorial2
+from scipy.special import lpmv
```
lcbMGH, line 7: Wider scope: on my installation, I'm getting "No name 'lpmv' in module 'scipy.special'". I either made an installation error or this isn't included in the installation packages

ghost (commented):
Nothing massive, just a few questions -- I'm trusting those more involved to have stronger opinions

shimmingtoolbox/coils/siemens_basis.py
po09i: Nice trick

shimmingtoolbox/coils/siemens_basis.py
po09i: We will eventually implement this function up to the third order so we will need to check for the size.

shimmingtoolbox/coils/siemens_basis.py
po09i: Not sure what the best practice is, I could not see a use for these functions outside the scope of siemens_basis

shimmingtoolbox/coils/spherical_harmonics.py
po09i: Since the function would have become a one liner, I removed the function and replaced it for the one liner

shimmingtoolbox/coils/siemens_basis.py
gab-berest: Normally, you would create private functions (a function that the name contains a leading _ or __). It doesn't do anything other than telling the user that these function are not supposed to be used form outside this scope. So, in your file, you would have other functions that would be called _get_scaling_factor or __get_scaling_factor, etc.

shimmingtoolbox/coils/spherical_harmonics.py
gab-berest: From what I see, indeed there is no lpvm function or object in scipy.special at [https://docs.scipy.org/doc/scipy-1.5.0/reference/special.html](url). @po09i could you give us your scipy version installed?

lcbMGH: Actually, I do see it -- https://docs.scipy.org/doc/scipy-1.5.0/reference/generated/scipy.special.lpmv.html#scipy.special.lpmv 

I can install this locally, but it would also be good to make sure this is included in the installation dependencies.

shimmingtoolbox/coils/siemens_basis.py
po09i: I'll just make them normal functions since their is no benefit other than not being callable from outside the scope, (which I see as a very slim potential inconvenience)

shimmingtoolbox/coils/siemens_basis.py
po09i: I found this handy [pdf](https://cbbi.udel.edu/wp-content/uploads/2017/01/Sys-DataManagement.pdf) which talks about the siemens console, on page 127, it shows a picture that shows how to perform shims on the console using the built in coils. I see it as an `Ax=-b` problem where A is the matrix of coil profiles with the siemens coefficients in micro-T/m for the first order and in micro-T/m^2 for the second order. `b` is the fieldmap to solve for in hz and `x` are the coefficients we are looking to input in the system. I'm also new to the subject so confirmation from @jcohenadad or @rtopfer would be great.

shimmingtoolbox/coils/siemens_basis.py
```diff
@@ -12,6 +12,86 @@
 GYROMAGNETIC_RATIO = 42.576
 
 
+def reorder_to_siemens(spher_harm):
```
gab-berest, line 15: If they are just used here would not need to be called elsewhere, I would call them _reorder_to_siemens and _get_scaling_factor.

shimmingtoolbox/coils/siemens_basis.py
rtopfer: @lcbMGH @po09i 
By design, the three 1st-order shims are linear fields (the X, Y and Z gradients).
E.g. Assume an initial field distribution that is on resonance everywhere (perfectly shimmed), were you to enter a value of "1" in the "X" row of the "Temporary" column of the Siemens 3D Shim adjustments page and press "Apply", the local field 1mm to the left of isocenter will be 1 nano-T greater than the field at isocenter (the origin where no field shift is introduced); another 1mm to the left, the field will be 2 nano-T. Equivalently, from the Larmor equation, dividing by gamma puts the shift in terms of Hz rather than Tesla.
Same concept applies to the 2nd-order terms, only the given units differ since the rate by which the field-shift increases/decreases moving away from isocenter is now quadratic rather than linear.
For any inductive shim coil, you're applying a current and introducing a certain amount of field shift. For the local "multi-coil" arrays, we're generally not designing them to meet a highly specific profile in the way the scanner shims are designed; i.e. spatial derivatives of the local coil profiles aren't constants: if it just so happened to be 1 micro-T/m somewhere, the figure will be entirely different somewhere. So, instead, we typically map out the field profiles and normalize them at 1 ampere (since the current circulating in a coil is what we actually control in an experiment). Siemens is likewise changing the current circulating in their shims, they've just opted to reframe the settings in this way since, in principal, it obviates the need to save the field profiles, which are supposed to conform to theoretical spherical harmonics, with the displayed units indicating the amplitudes/scaling values. (In practice, as one moves further and further away from isocenter, the actual Siemens profiles will deviate more and more from spec and their spatial derivatives won't conform reliably to the given constants).

The 3D Shim card also features the "Measure" & "Calculate" buttons, where "Measure" is acquiring _some_ form of field measurement and "Calculate" is performing _some_ type of optimization. There are different options available for the type of measurement, but the overall shim routine here is something of a black box. Surely it's along the lines of solving the least-squares `Ax=-b` problem, but the general consensus seems to be that manually acquiring a field map and performing the optimization offline does a better job than Siemens, so I'm not sure what they're doing exactly (and they weren't willing to reveal anything else about it when we asked a few years ago).

PR #78, opened 2020-09-01
Repair dicom_to_nifti missing calls [bug]

### LINKED ISSUE #77: Missing tsv and dataset_description.json from dicom_to_nifti
When using dicom_to_nifti function, most of the dataset information files in json or tsv format are missing. dcm2bids should be correctly configured to generate at least the template files without information.

gab-berest:
There was a bug in the dicom_to_nifti function that forgot to call dcm2bids_scaffold and dcm2bids_helper which create the file hierarchy needed to respect BIDS convention and be used by other BIDS library (missing mandatory calls to work).

This pull-request wants to add these calls and check which code is necessary or not to have a final BIDS file hierarchy.

### DIFF
--- a/shimmingtoolbox/dicom_to_nifti.py
+++ b/shimmingtoolbox/dicom_to_nifti.py
@@ -7,7 +7,10 @@ from distutils.dir_util import copy_tree
 import json
 import numpy as np
 import os
+import sys
 import subprocess
+import dcm2bids
+from dcm2bids.scaffold import scaffold
 
 from shimmingtoolbox import __dir_config_dcm2bids__
 
@@ -30,9 +33,11 @@ def dicom_to_nifti(path_dicom, path_nifti, subject_id='sub-01', path_config_dcm2
         raise FileNotFoundError("No dcm2bids config file found")
     if not os.path.exists(path_nifti):
         os.makedirs(path_nifti)
+
+    # dcm2bids is broken for windows as a python package so using CLI
     # Create bids structure for data
-    # TODO; use dcm2bids as python package (no system call)
     subprocess.run(['dcm2bids_scaffold', '-o', path_nifti], check=True)
+
     #
     # # Copy original dicom files into nifti_path/sourcedata
     copy_tree(path_dicom, os.path.join(path_nifti, 'sourcedata'))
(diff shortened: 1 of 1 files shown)


gab-berest:
We can have the discussion on what code is necessary or not for the dicom_to_nifti function @gaspardcereza  @jcohenadad  @po09i

gaspardcereza:
> For me there are some folder checks that need to be included so there is no error regarding the existance of directories.

I personally agree with the fact that we need to check the existence of the created directories (it's not time consuming and can be useful for debugging).

> The config file part should not be part of this for me since, if there is weird or special file system or acquisition types, the user should provide for me his configuration so we know how he got his acquisition. Tell me if I'm wrong.

Regarding the configuration file, we decided to go with a manually written file located in `/config/dcm2bids.json` so the automatic writting of the file is indeed not necessary anymore and can be removed.

gab-berest:
From what I've seen in the MATLAB version and the previous code, all the for loops are used to parse through acquisitions and call dcm2bids for each nifti generated file. I don't think these parts are necessary since calling dcm2bids on the folder containing these files ave the same result, if not cleaner since it creates a sub directory and we do not have to create manually the parent directory.

gab-berest:
I made the changes discussed. I also transformed the code for use of dcm2bids as a python package instead of CLI subprocess. However, the dcm2bids.scaffold is broken actually so I commented this part for now.

shimmingtoolbox/dicom_to_nifti.py
```diff
@@ -1,20 +1,98 @@
...
+
+    #
+    # # Copy original dicom files into nifti_path/sourcedata
+    copy_tree(path_dicom, os.path.join(path_nifti, 'sourcedata'))
+    #
+    # # Call the dcm2bids_helper
+    subprocess.run(['dcm2bids_helper', '-d', path_dicom, '-o', path_nifti], check=True)
+    #
+    # Check if the helper folder has been created
+    helper_path = os.path.join(path_nifti, 'tmp_dcm2bids', 'helper')
```
gaspardcereza, line 49: With our convention, it should now be named `path_helper`

gaspardcereza (approved):

shimmingtoolbox/dicom_to_nifti.py
```diff
@@ -1,20 +1,98 @@
...
+import numpy as np
+import os
+import sys
 import subprocess
+import dcm2bids
+# from dcm2bids.scaffold import scaffold
 
 from shimmingtoolbox import __dir_config_dcm2bids__
 
+# TODO: check in unit test if dcm2bids_scaffold is installed, and also check for the required version.
```
po09i, line 17: Add as an issue instead of a todo. Everything else seems good, i'll approve right after

gab-berest:
@po09i Everything is changed and good from your comments.

shimmingtoolbox/dicom_to_nifti.py
```diff
@@ -1,20 +1,98 @@
-# coding: utf-8
+#!usr/bin/env python3
+# -*- coding: utf-8
 
 # TODO: check in unit test if dcm2bids_scaffold is installed, and also check for the required version.
```
po09i, line 4: Forgot this other todo

gab-berest: Oups. Didn't see this one.

po09i (approved):

PR #90, opened 2020-09-16
[chore][cli/download_data] Remove stale TODO

### LINKED ISSUE #89: TODO in cli/download_data.py has been addressed
The comment states:

``` python
# TODO: logging does not seem to output on the terminal
```

Running the file in the Terminal with and without the `verbose` flag executes the script with the expected behaviour: with `--verbose`, a series of `INFO` logs are output to the terminal. Without the `--verbose flag`, these logs are not generated and output to the terminal.

jidicula:
The `--verbose` flag works as expected, so the TODO has already been
addressed.

**What this change does**
Remove the TODO.

**Any side-effects?**
None

**Additional context/notes/links**

Resolves #89

----

#

### DIFF
--- a/shimmingtoolbox/cli/download_data.py
+++ b/shimmingtoolbox/cli/download_data.py
@@ -28,7 +28,6 @@ def main(verbose, output, data):
         output: Output folder.
         data: The data to be downloaded.
     """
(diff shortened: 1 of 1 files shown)


gab-berest (approved):
All good!

PR #92, opened 2020-09-16
Added example shell script for realtime z-shimming

jcohenadad:
Note: this is a work in progress, with comments about CLIs yet to be created.


evaalonsoortiz:
Is this script for performing realtime z-shimming, ie computing the gradient of the field map along z? I am asking because I see that the coil profiles are downloaded:
https://github.com/shimming-toolbox/shimming-toolbox-py/blob/189e9fe1210ace76098f032571aa68cd5e51bd7f/examples/demo_realtime_zshimming.sh#L27

evaalonsoortiz:
Note: sorry I closed the PR by accident (and then reopened)

jcohenadad:
> Is this script for performing realtime z-shimming, ie computing the gradient of the field map along z? I am asking because I see that the coil profiles are downloaded:

Sorry for the late reply-- indeed, we do not need the coil profile. We started drafted a general scenario, but in the case of Gz shimming (where we do not rely on a calibrated coil profile but instead trust the scanners's input Gz value), we don't need this additional coil profile.

po09i:
I tried running the testing-data on the matlab repo to get a reference point of what the output should look like. Since only a portion of the data was uploaded to data-testing for size constraints (3 echoes and 10 time points), the matlab repo which depends on the header (which does not fit anymore) throws a lot of errors. Which acquisition did you use @evaalonsoortiz for the realtime_zshim data in this [PR](https://github.com/shimming-toolbox/data-testing/pull/4) so I can test it locally.

po09i:
The name of the file is ACDC112 so I'll assume this is correct

examples/demo_realtime_zshimming.sh
```diff
@@ -0,0 +1,64 @@
...
+# Go inside folder
+cd testing_data/realtime_zshimming_data
+
+# dcm2bids -d . -o nifti -p sub-example -c ../../config/dcm2bids.json
+st_dicom_to_nifti -i . -o nifti -sub sub-example
+cd nifti/sub-example/fmap
+
+# Calling FSL directly
+fsl_prepare_fieldmap SIEMENS sub-example_phasediff.nii.gz sub-example_magnitude1.nii.gz sub-example_phasediff_unwrapped.nii.gz 2.46 --nocheck
+fslmaths sub-example_phasediff_unwrapped.nii.gz -div 0.00246 sub-example_fieldmap.nii.gz
```
po09i: Should be 6.28 to go from radians/s to hz

examples/demo_realtime_zshimming.sh
```diff
@@ -0,0 +1,81 @@
...
+# Download example data
+st_download_data testing_data
+
+# Go inside folder
+cd testing_data/realtime_zshimming_data
+
+# dcm2bids -d . -o nifti -p sub-example -c ../../config/dcm2bids.json
+st_dicom_to_nifti -input . -output ../nifti -subject sub-example
+cd ../nifti/sub-example/fmap
+# TODO: Name of phase2 should be phasediff
```
jcohenadad: open issue to address that (should be implemented in `st_dicom_to_nifti`

po09i (approved):

PR #97, opened 2020-09-21
Moved testing and docs packages in default installation

### LINKED ISSUE #96: Installation issue with zsh
the following command does not run on zsh
````
pip install -e .[testing]
zsh: no matches found: .[testing]
````

@lcbMGH will fill up with more info

ghost:
Minor change shifting testing packages to required installation, instead of extra, to add to ease of installation.

### DIFF
--- a/docs/source/2_getting_started/1_installation.rst
+++ b/docs/source/2_getting_started/1_installation.rst
@@ -159,7 +159,7 @@ that you followed the ``git clone`` instruction above:
 .. code:: bash
 
    cd shimming-toolbox
-   pip install -e ".[testing,docs]"
+   pip install -e .
 
 .. NOTE ::
    If you downloaded shimming-toolbox using the link above
@@ -178,7 +178,7 @@ version of the project from GitHub and reinstall the application:
 
    cd shimming-toolbox
    git pull
-   pip install -e ".[testing,docs]"
+   pip install -e .
 
 Testing the installation
 ------------------------
--- a/docs/source/2_getting_started/1_installation.rst
+++ b/docs/source/2_getting_started/1_installation.rst
@@ -159,7 +159,7 @@ that you followed the ``git clone`` instruction above:
 .. code:: bash
 
    cd shimming-toolbox
-   pip install -e .
+   pip install -e ".[docs]"
 
 .. NOTE ::
    If you downloaded shimming-toolbox using the link above
@@ -178,7 +178,7 @@ version of the project from GitHub and reinstall the application:
 
    cd shimming-toolbox
    git pull
-   pip install -e .
+   pip install -e ".[docs]"
 
 Testing the installation
 ------------------------
--- a/setup.py
+++ b/setup.py
@@ -36,9 +36,10 @@ setup(
         "scipy~=1.5.0",
         "tqdm",
         'matplotlib~=3.1.2',
+        'testing': ["pytest~=4.6.3", "pytest-cov~=2.5.1"],
     ],
     extras_require={
-        'testing': ["pytest~=4.6.3", "pytest-cov~=2.5.1"],
+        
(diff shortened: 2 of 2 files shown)


jcohenadad (commented):
should the doc also be updated? (ie: remove the "[testing]" with pip command)

lcbMGH:
What's the process for updating the docs? That feels like a good call, how can I do it?

jcohenadad:
> @jcohenadad What's the process for updating the docs? That feels like a good call, how can I do it?

just remove the [testing] [here](https://github.com/shimming-toolbox/shimming-toolbox-py/blob/master/docs/source/2_getting_started/1_installation.rst).

lcbMGH:
> > @jcohenadad What's the process for updating the docs? That feels like a good call, how can I do it?
> 
> just remove the [testing] [here](https://github.com/shimming-toolbox/shimming-toolbox-py/blob/master/docs/source/2_getting_started/1_installation.rst).
> 
> EDIT 2020-09-21 12:53:12: actually, you might also want to do the same for [docs].
Should the instructions still include docs instructions? What's the use and decision around that being the only thing in the extras tab currently?

jcohenadad:
> Should the instructions still include docs instructions? What's the use and decision around that being the only thing in the extras tab currently?

sorry i'm not sure i understand the question. Are you referring to the pip install command? (ie: if it should still include "docs")?

lcbMGH:
> > Should the instructions still include docs instructions? What's the use and decision around that being the only thing in the extras tab currently?
> 
> sorry i'm not sure i understand the question. Are you referring to the pip install command? (ie: if it should still include "docs")?

Yes, and particularly why docs is in the extras section and not the required installation section. If we feel that it's simpler to move testing into the required install section so we don't have to worry about zsh bracket escape, why are we keeping docs there?

jcohenadad:
> Yes, and particularly why docs is in the extras section and not the required installation section. If we feel that it's simpler to move testing into the required install section so we don't have to worry about zsh bracket escape, why are we keeping docs there?

I agree with you, hence my comment in https://github.com/shimming-toolbox/shimming-toolbox-py/pull/97#issuecomment-696239312: "you might also want to do the same for [docs]." (ie: put it in the default install pckgs), sorry if i was unclear.

lcbMGH:
> > Yes, and particularly why docs is in the extras section and not the required installation section. If we feel that it's simpler to move testing into the required install section so we don't have to worry about zsh bracket escape, why are we keeping docs there?
> 
> I agree with you, hence my comment in [#97 (comment)](https://github.com/shimming-toolbox/shimming-toolbox-py/pull/97#issuecomment-696239312): "you might also want to do the same for [docs]." (ie: put it in the default install pckgs), sorry if i was unclear.

Phenomenal, I think I just misread. Thanks for the clarification!

jcohenadad:
@gaspardcereza just told me he had trouble running pytest because not all packages were properly installed, so this is a +1 for doing this change (@gaspardcereza pls let us know if the error was coming from something else)

jcohenadad (approved):

gaspardcereza:
> @gaspardcereza just told me he had trouble running pytest because not all packages were properly installed, so this is a +1 for doing this change (@gaspardcereza pls let us know if the error was coming from something else)

Yes I just tried it with this branch and it worked ! So I agree that it would be useful to merge it.

jcohenadad (approved):

po09i (changes requested):
I'm not sure this is necessary, as discussed in the https://github.com/shimming-toolbox/shimming-toolbox-py/issues/96#issuecomment-696394639.

po09i:
Added the `docs` back in the optional set up as discussed here https://github.com/shimming-toolbox/shimming-toolbox-py/issues/96#issuecomment-696863526

po09i (approved):

jcohenadad (approved):

PR #106, opened 2020-09-25
Update download_data CLI dataset links

### LINKED ISSUE #84: Update testing data
The current testing data release (r20200713) is fairly old and does not include the latest additions.

We should consider updating the link on the `st_download_data` CLI.

Things to check before updating the link: 
- [ ] Make sure the testing data contain examples for realtime z-shimming scenario

### LINKED ISSUE #85: st_download_data -h fails
~~~
st_download_data -h
Usage: st_download_data [OPTIONS] DATA
Try 'st_download_data --help' for help.
~~~

po09i:
This PR updates the link to the latest datasets in data-testing. The new dataset mainly adds the realtime_zshim dataset so that we have the required data to perform realtime zshimming.

The PR also adds the ability to show the help by using the `-h` option.

We could consider changing the prefix to the CLI in the same PR.

Fixes #84
Fixes #85

Future PRs:
Create and update CLIs for realtime zshimming

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ setup(
         'console_scripts': [
             "st_referencemaps=shimmingtoolbox.cli.referencemaps:main",
             "st_b0maps=shimmingtoolbox.cli.b0map:main",
-            "st_download_data=shimmingtoolbox.cli.download_data:main",
+            "st_download_data=shimmingtoolbox.cli.download_data:download_data",
         ]
     },
     packages=find_packages(exclude=["contrib", "docs", "tests"]),
--- a/shimmingtoolbox/cli/download_data.py
+++ b/shimmingtoolbox/cli/download_data.py
@@ -48,7 +48,6 @@ def download_data(verbose, output, data):
         output: Output folder.
         data: The data to be downloaded.
     """
-    click.echo("test")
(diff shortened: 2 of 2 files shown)


jcohenadad:
ready for review?

po09i:
Yes it is now ready for review, I wanted to add the -h option in the same PR.

jcohenadad (approved):

jcohenadad:
@po09i why is that labeled as "bug" instead of "refactoring"? the PR body does not mention any bug

jidicula (approved):
I've modified the PR title to clarify that the link update was the primary concern addressed.

Other fixes with CLI flags and renaming `main()` are technically out of scope for this and belong in their own PR but since this is a small fix, we can probably let it slide. Make sure not to squash when you rebase-merge so we can roll back each of these fixes individually if we need to without losing the others.

po09i:
My original plan was to have this PR to address a couple of issues with the download_data CLI which is why the PR title was vague. I'll try to stick to one issue at a time per PR as this reduces the length of the PR and is in line with https://github.com/shimming-toolbox/shimming-toolbox-py/issues/14#issuecomment-700796893.

PR #124, opened 2020-10-05
Optimizing optimizer code to run faster on large fieldmaps with small masks

### LINKED ISSUE #123: Optimizers run slow on large fieldmaps with small masks
The optimizer currently runs by creating a full mask of the entire fieldmap, including tracking 0 elements. This should be refactored to only optimize on nonzero mask elements.

ghost:
## Description
The optimizers run slowly on large fieldmaps, even with small masks -- this is because they continue to track the areas zeroed out in the masks. This PR addresses tracking sizes -- it trims the area to optimize down to the mask size, improving speeds by a lot. 

## Linked issues
Resolves #123


po09i (approved):

jidicula (approved):
LGTM

PR #131, opened 2020-10-06
Update __dir_testing__ to be an absolute path

### LINKED ISSUE #121: __dir_testing__ should include the __dir_shimming_toolbox__ path
`__dir_testing__` simply is `testing_data` and needs to be used in conjunction with `__dir_shimmingtoolbox__` to get an absolute path to avoid errors when being used. 

It should have the same behaviour as `__dir_config_dcm2bids__` which returns an abolute path.

Implementation suggestion:
`__dir_testing__ = os.path.join(__dir_shimmingtoolbox__, 'testing_data')`

po09i:
## Description
This PR updates the `__dir_testing__` global variable to be an absolute path instead of a relative path. It makes it more in line with the other variable such as `__dir_config_dcm2bids__`.  

Changing this has consequences on the general_demo script which used this variable. The problem was fixed.

## Linked issues
Fixes #121

### DIFF
--- a/examples/general_demo.py
+++ b/examples/general_demo.py
@@ -45,8 +45,8 @@ def general_demo(path_output=os.path.join(os.path.curdir, 'output_dir')):
     logging.basicConfig(level='INFO')
 
     # Download example data
-    path_testing_data = os.path.join(path_output, __dir_testing__)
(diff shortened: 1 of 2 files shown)


jidicula (approved):
LGTM!

PR #132, opened 2020-10-06
fix remove_tmp argument in dicom_to_nifti

### LINKED ISSUE #125: remove_tmp API flag does not work in dicom_to_nifti
### Versions
Project version: e5edd40dfe32e118c14201338684dbdef0d6adfb
Python: 3.7.3

### Bug Description
remove_tmp flag removes the `tmp_dcm2bids` folder from the dicom path instead of the nifti path.

FileErrorNotFound

### Reproduction Steps
Use dicom_to_nifti API with remove_tmp flas as True

Look at the output nifti folder, it will still contain the tmp_dcm2bids

https://github.com/shimming-toolbox/shimming-toolbox/blob/42bdddfd35e6a5140e65b344d4a615eab4e9492c/shimmingtoolbox/dicom_to_nifti.py#L93
This line should be: 
os.removedirs(os.path.join(path_nifti, 'tmp_dcm2bids'))

po09i:
## Description
This PR adresses a bug where the remove_tmp argument would try to remove the tmp_dcm2bids folder from the dicom path instead of the nifti path. A test was also created for it.

## Linked issues
Fixes #125


shimmingtoolbox/dicom_to_nifti.py
```diff
@@ -90,4 +91,4 @@ def dicom_to_nifti(path_dicom, path_nifti, subject_id='sub-01', path_config_dcm2
     #     bids_info.run()
 
     if remove_tmp:
-        os.removedirs(os.path.join(path_dicom, 'tmp_dcm2bids'))
+        shutil.rmtree(os.path.join(path_nifti, 'tmp_dcm2bids'))
```
jidicula, line 94: good fix, `shutil.rmtree` is indeed [recommended over `os.removedirs` in the docs](https://docs.python.org/3/library/os.html#os.rmdir) to prevent throwing an error when dir is not empty

jidicula (approved):
LGTM!

Issue #37, opened 2020-07-28
Add example scenario for realtime z-shimming


gaspardcereza:
A few specifications:
- The scenario would be based on the previous [Matlab implementation ](https://github.com/shimming-toolbox/shimming-toolbox/blob/master/example/realtime_zshim.m)
- Would be using our [testing data](https://github.com/shimming-toolbox/data-testing/tree/master/realtime_zshimming_data)
- The example scripts should be shell scripts, wrapping the various CLIs, with a short explanation of what each CLI does. That way, users will be able to copy/paste and go through the example script line by line, and play with the various inputs/outputs and change parameters for each CLIs. This follows the philosophy of other example scripts, e.g. [SCT's](https://github.com/neuropoly/spinalcordtoolbox/blob/master/batch_processing.sh).

jcohenadad:
Relevant comment: https://github.com/shimming-toolbox/shimming-toolbox-py/issues/13#issuecomment-666004847

po09i:
Issue can be closed, PR #92 implements a shell script which demos how to use the toolbox and realtime zshimming

PR #134, opened 2020-10-07
Realtime_zshim shim CLI

### LINKED ISSUE #101: Create an API and  a CLI for realtime_zshimming sequencer and optimizer
## What is realtime_zhim
TODO

## API
The API will:
- resample coil profile into the space of anat
- resample (in time) the physio trace to the 4d fieldmap data so that each time point of the fieldmap has its corresponding respiratory probe value.
- run optimizer within mask.nii
- outputs:
  - fieldmap_shimmed.nii
  - coefficients.csv
  - figures

Inputs
- Fieldmap
- Anat
- PMU
- Mask
- Output
- Coil Profile

## CLI

CLI name suggestion
- st_realtime_zshim

CLI flags
- fmap
- anat
- resp
- mask
- output
- coil_profile

Motivated by #37

po09i:
## Description
This PR adds the ability to do realtime zshimming with the CLI. The code is based on the realtime_zshim script Matlab repo.

Some definitions for the gradients: 
- Gslice, Gphase, Gfreq are in the patient coordinate system.
- Gx, Gy, Gz are in the scanner coordinate system

Some background about z-shimming workflow:
- Acquire a multi-echo fieldmap data (eg could be sagittal)
- B0 fieldmap is calculated for each voxel
- Use `phys_gradient` to calculate voxelwise Gx, Gy, Gz
- calculate RIRO, STATIC for Gx, Gy and Gz
- Resample and regrid to the axial scan (So we have what would be Gx, Gy and Gz mapped on the axial scan voxels)
- Calculate Gslice of the axial scan  using `phys_to_vox_gradient`
- We call it "z-shimming", where the "z" is relative to the coordinate system of the image (ie: the slice select), not the physical coordinate system. Ie: there will be contribution from the X and Y gradients as well for non "pure" axial orientation which is why we want Gslice.

## Linked issues
Related to #101
Fixes to #37


po09i:
Read the docs timed out but when looking at the log it says it succeeded. I think this is a readthedocs bug not one on our side. We will see on the next push.

po09i:
There is a bug in lsq_optimizer which is why the build is failing. I opened an issue #152

evaalonsoortiz:
In the matlab code, we are doing option 2, as listed above (see [here](https://github.com/shimming-toolbox/shimming-toolbox-matlab/blob/7f0ba2d480a697a16a0d4b20154c61171f7e13ca/Img/MaRdI.m#L1150)). I think both options are equivalent, but with option 1 you are limited by the fact that you may have a small number of slices for the MGRE scan, which would introduce errors in the Gslice calculation.

po09i:
I think the relevant line is [here](https://github.com/shimming-toolbox/shimming-toolbox-matlab/blob/7f0ba2d480a697a16a0d4b20154c61171f7e13ca/example/realtime_zshim.m#L191). From what I understand, this gives a gradient which is wrong if Gslice and Gz are not aligned. The gradient attempts to give Gslice but if the image is rotated, the z voxel positions given are right but are not appropriate to calculate the gradient. The coordinates should be along Gslice and not along Gz. I'll reiterate that I think the current implementation works for past experiments since the fieldmap is aligned with the scanner coordinate system.

I think the `resliceImg()` [function](https://github.com/shimming-toolbox/shimming-toolbox-matlab/blob/7f0ba2d480a697a16a0d4b20154c61171f7e13ca/Img/MaRdI.m#L1150) you are mentioning remaps the gradient to the appropriate coordinates of the axial scan but if the initial fieldmap gradient is wrong (due to the above explanation) or if the gradient is along Gslice (Gslice of the fieldmap where Gslice is not along Gz), then the resliced img will be wrong as well.

shimmingtoolbox/coils/coordinates.py
```diff
@@ -125,3 +127,49 @@ def phys_to_vox_gradient(gx, gy, gz, affine):
              (gz * inv_affine[2, z_vox] * z_vox_spacing)
 
     return gx_vox, gy_vox, gz_vox
+
+
+def resample_from_to(nii_from_img, nii_to_vox_map, order=3, mode='constant', cval=0., out_class=nib.Nifti1Image):
+    """ Wrapper to nibabel's ``resample_from_to`` function. Resample image `from_img` to mapped voxel space
+    `to_vox_map`. The wrapper adds support for 2D input data (adds a singleton) and 4D timeseries support.
```
jcohenadad:
```suggestion
    `to_vox_map`. The wrapper adds support for 2D input data (adds a singleton) and for 4D time series.
```

shimmingtoolbox/shim/realtime_zshim.py
```diff
@@ -0,0 +1,259 @@
...
+from shimmingtoolbox.utils import st_progress_bar
+from shimmingtoolbox.coils.coordinates import resample_from_to
+from shimmingtoolbox.coils.coordinates import phys_gradient
+from shimmingtoolbox.coils.coordinates import phys_to_vox_gradient
+
+
+def realtime_zshim(nii_fieldmap, nii_anat, pmu, json_fmap, nii_mask_anat=None, path_output=None):
+    """ This function will generate static and dynamic (due to respiration) Gz components based on a fieldmap time
+    series and respiratory trace information obtained from Siemens bellows  (PMUresp_signal.resp). An additional
+    multi-gradient echo (MGRE) magnitiude image is used to generate an ROI and resample the static and dynamic Gz
```
jcohenadad:
```suggestion
    multi-gradient echo (MGRE) magnitude image is used to generate an ROI and resample the static and dynamic Gz
```

shimmingtoolbox/shim/realtime_zshim.py
```diff
@@ -0,0 +1,259 @@
...
+        nii_mask_anat = nib.Nifti1Image(np.ones_like(anat), nii_anat.affine)
+        
+    if is_outputting_figures:
+        nib.save(nii_mask_fmap, os.path.join(path_output, 'fig_mask_fmap.nii.gz'))
+
+    masked_fieldmaps = np.zeros_like(fieldmap)
+    for i_t in range(nt):
+        masked_fieldmaps[..., i_t] = mask_fmap * fieldmap[..., i_t]
+
+    # Calculate gz gradient
```
jcohenadad:
```suggestion
    # Calculate Gx, Gy and Gz gradients (in the physical coordinate system)
```

shimmingtoolbox/coils/coordinates.py
```diff
@@ -125,3 +127,49 @@ def phys_to_vox_gradient(gx, gy, gz, affine):
              (gz * inv_affine[2, z_vox] * z_vox_spacing)
 
     return gx_vox, gy_vox, gz_vox
+
+
+def resample_from_to(nii_from_img, nii_to_vox_map, order=3, mode='constant', cval=0., out_class=nib.Nifti1Image):
```
jcohenadad: maybe change default order to 2 (trilinear), because 3 creates oscillations (maybe problematic for our applications?) and is longer

shimmingtoolbox/shim/realtime_zshim.py
```diff
@@ -0,0 +1,259 @@
...
+                                                 -gradient[g_axis][i_x, i_y, i_z, :])
+                    # Multiplying by the RMS of the pressure allows to make abstraction of the tightness of the bellow
+                    # between scans. This allows to compare results between scans.
+                    riro[g_axis][i_x, i_y, i_z] = reg.coef_ * pressure_rms
+                    static[g_axis][i_x, i_y, i_z] = reg.intercept_
+                    progress_bar.update(1)
+
+    # Resample masked_fieldmaps to target anatomical image
+    nii_masked_fieldmaps = nib.Nifti1Image(masked_fieldmaps, nii_fieldmap.affine)
+    nii_resampled_fmap = resample_from_to(nii_masked_fieldmaps, nii_anat, order=2, mode='nearest')
```
jcohenadad: order not needed if defaulted to 2

shimmingtoolbox/shim/realtime_zshim.py
```diff
@@ -0,0 +1,259 @@
...
+    # Since this is zshimming, only the slice component is used.
+    _, _, resampled_static_vox = phys_to_vox_gradient(resampled_static[0], resampled_static[1], resampled_static[2],
+                                                      nii_anat.affine)
+    nii_resampled_static_vox = nib.Nifti1Image(resampled_static_vox, nii_anat.affine)
+    nii_resampled_static_masked = nib.Nifti1Image(resampled_static_vox * nii_mask_anat.get_fdata(),
+                                                  nii_resampled_static_vox.affine)
+    if is_outputting_figures:
+        nib.save(nii_resampled_static_masked, os.path.join(path_output, 'fig_resampled_static.nii.gz'))
+
+    # Resample riro to target anatomical image
```
jcohenadad, line 153: high duplication with previous paragraph-- could potentially be a private local function

po09i:
@jcohenadad
Current approach (Option 1):
- Calculate Gx, Gy, Gz
- Calculate riro and static for each gradient
- Resample to the target anat image
- Calculate Gss

As discussed in the internal meeting, I implemented the approach (Option 2) where we:
- Calculate Gx, Gy, Gz 
- Resample to the target anat image
- Calculate Gss
- Calculate riro and static

Both give the same results as we would expect however, 
Option 1: 8s
Option 2: 52s, neater code

The time difference seems to be related to having to resample the fieldmap to the anat image for all timepoints and axis(x, y and z) instead of doing it only for the static and riro. Since both give the same results, I prefer Option 1.

jcohenadad (approved):
Awesome contribution @po09i 🏅 !!!

PR #136, opened 2020-10-08
Ensure test environment is set up with test data and soft dependencies before running tests

### LINKED ISSUE #118: Make sure the environment is properly setup and binaries are present before running pytest
A few steps are required for pytest to run, such as:
- [x] download testing data
- [ ] download 3rd party binaries (eg prelude, dcm2niix)

To prevent confusion (eg https://github.com/shimming-toolbox/shimming-toolbox/issues/102), we should have pytest validate the presence of the necessary files before running. Or, we could also have pytest download those files, as being done in [SCT](https://github.com/neuropoly/spinalcordtoolbox/blob/master/conftest.py)

jidicula:
## Checklist

## Description

We want to ensure that test data has been pulled into the test
environment before running tests.

**What this change does**
Executes the download_data() CLI inside a pytest hook that runs first
when the test suite session begins.

**Any side-effects?**
Creates a new temp directory and populates it with test data.

## Linked issues

Resolves #118 - Make sure the environment is properly setup and
binaries are present before running pytest

May want to refactor with an @setup decorator:
https://stackoverflow.com/a/12600154/6310633.

### DIFF
--- a/test/test_prelude.py
+++ b/test/test_prelude.py
@@ -7,12 +7,13 @@ import glob
 import nibabel as nib
 import numpy as np
 import logging
+import pytest
 
 from shimmingtoolbox.unwrap import prelude
 
 
+@pytest.mark.prelude
 class TestCore(object):
-
     def setup(self):
         # Get the directory where this current file is saved
         self.full_path = Path(__file__).resolve().parent
@@ -41,7 +42,9 @@ class TestCore(object):
         path_data = glob.glob(os.path.join(self.toolbox_path, 'testing_data*'))[0]
 
         # Open phase data
-        fname_phases = glob.glob(os.path.join(path_data, 'sub-fieldmap', 'fmap', '*phase*.nii.gz'))
+        fname_phases = glob.glob(
+            os.path.join(path_data, 'sub-fieldmap', 'fmap', '*phase*.nii.gz')
+        )
         if len(fname_phases) > 2:
             raise IndexError('Phase data parsing is wrongly parsed')
 
@@ -53,7 +56,9 @@ class TestCore(object):
         phase_e2 = np.interp(nii_phase_e2.get_fdata(), [0, 4096], [-np.pi, np.pi])
 
         # Open mag data
-        fname_mags = glob.glob(os.path.join(path_data, 'sub-fieldmap', 'fmap', '*magnitude*.nii.gz'))
+        fname_mags = glob.glob(
+            os.path.join(path_data, 'sub-fieldmap', 'fmap', '*magnitude*.nii.gz')
+        )
 
         if len(fname_mags) > 2:
             raise IndexError('Mag data parsing is wrongly parsed')
@@ -75,7 +80,7 @@ class TestCore(object):
         # default prelude call
         unwrapped_phase_e1 = prelude(self.phase_e1, self.mag_e1, self.affine_phase_e1)
 
-        assert (unwrapped_phase_e1.shape == self.phase_e1.shape)
+        assert unwrapped_phase_e1.shape == self.phase_e1.shape
 
     def test_non_default_mask(self):
         """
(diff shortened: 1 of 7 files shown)


jidicula:
Questions:

1. Should the temp directory be removed after the testing run is
complete? Can we expect that the testing environment will be connected
to the internet (so they would be able to redownload test data each time the testing suite is run)?

2. Not sure if this is the right way to go about it – I was mostly following the SCT approach for test data download but it still seems a bit janky to me.

Next steps:

- [x] check for dcm2niix with litmus command
- [x] check for prelude with litmus command
- [x] make those soft dependency checks optional, and add some conditions to the corresponding tests to only run if the user specifies
- [ ] ~~add an `--all` flag that runs pytest with all the flags for soft dependencies~~ vanilla `pytest` runs all tests, user must specify if they're testing without the soft dependencies. See RTD build for details.

jcohenadad:
> Should the temp directory be removed after the testing run is complete? 

I guess we could, and have a flag to keep it for debugging purpose

> Can we expect that the testing environment will be connected to the internet (so they would be able to redownload test data each time the testing suite is run)?

not sure i understand-- do you mean for the CI?

po09i:
Current implementation uses testing_data which is assumed to be in the root of shimming-toolbox. If this is changed lets keep in mind that the `.gitignore` will have to be changed and the `__dir_testing__` variable as well. Not an issue but something to put on the todo list.

Ideally testing-data is downloaded once and not downloaded again if testing-data is intact or if the data is up to data but it seems like quite a bit of work something minor. So I would go with what Julien said.

>  Can we expect that the testing environment will be connected to the internet (so they would be able to redownload test data each time the testing suite is run)?

I think this is a fair assumption, maybe a flag could be set up with a path to a local version of testing-data if the need ever arises

jidicula:
> not sure i understand-- do you mean for the CI?

I mean for users – are we assuming that the device they've installed shimming-toolbox on will be connected to the Internet to be able to fetch test data if they run `pytest` on it?

jidicula:
> I think this is a fair assumption, maybe a flag could be set up with a path to a local version of testing-data if the need ever arises

out of scope for now, but definitely a future ticket

jidicula:
🎉 Excellent, I found [this](https://docs.pytest.org/en/latest/example/markers.html#custom-marker-and-command-line-option-to-control-test-runs) pytest feature that should give us the ability to selectively run tests with the `dcm2niix` and `prelude` soft dependencies – we can add a specific setup in `conftest.py` checking if those deps are available, then add the corresponding decorator to any test function that requires that soft dependency.

edit: jk we're going to use markers defined in a `pytest.ini` file, scroll to top of the same page for an example.

jidicula:
Ok, I think this is ready for review. If you're reviewing it, I recommend you read my commit messages to understand some of my reasoning in taking this approach.

jidicula:
Note that once this is merged, the correct marker must be added to any tests depending on those soft dependencies. We can also add new markers for any future soft dependencies, or for other conditional tests.

test/test_prelude.py
```diff
@@ -85,10 +91,16 @@ def test_non_default_mask(self):
         mask = np.ones(self.phase_e1.shape)
 
         # Call prelude with mask (is_unwrapping_in_2d is also used because it is significantly faster)
-        unwrapped_phase_e1 = prelude(self.phase_e1, self.mag_e1, self.affine_phase_e1, mask, is_unwrapping_in_2d=True)
+        unwrapped_phase_e1 = prelude(
```
jcohenadad: @jidicula do you have an automatic formatting in your IDE? if so maybe we should discuss formatting conventions. I personally don't mind how the code was written before-- it was still compliant with PEP8. PEP8 is fairly flexible when it comes to how code should be wrapped, so it's a matter of internal convention (i tend to like concise code, not spread out across multiple lines)

jidicula: Yes, sorry about that! I'll disable it in my local settings for this repo.

jidicula: Shall I try to roll back these formatting changes in a fixup commit?

jcohenadad: if you don't mind-- otherwise no big deal- it's still clean code 😊

po09i:
What is the new situation for the testing_data? I noticed the code uses `__dir_testing__` which leads to shimming-toolbox/testing_data so the new tmp/ directory won't be used.

Also if the testing_data is downloaded in the new tmp folder, we probably want to remove the call to `st_download_data` in .travis.yml

jidicula:
> What is the new situation for the testing_data? I noticed the code uses `__dir_testing__` which leads to shimming-toolbox/testing_data so the new tmp/ directory won't be used.
> 
> Also if the testing_data is downloaded in the new tmp folder, we probably want to remove the call to `st_download_data` in .travis.yml

There's not really a new situation for testing_data afaik – it's creating `./testing_data` and downloading the data into it. I think this is the expected location for the testing data, correct?

I'll update the Travis config as well.

jidicula:
Coveralls [seems to be failing](https://travis-ci.com/github/shimming-toolbox/shimming-toolbox/builds/189211908#L463) in our Travis build, but it's after success. Does something have to be fixed for this?

Edit: seems related to [this](https://github.com/z4r/python-coveralls/issues/74#issuecomment-582170116)

jcohenadad (approved):

Issue #66, opened 2020-08-11
Improve masking capabilities


po09i:
## Context
The initial issue #47 which was partially fixed by PR #60 added some basic capabilities to mask some data (threshold, shape: square, cube). More masking capability could be implemented to segment the brain and spinal cord as well as other shapes.

Algorithms that could help
- SCT: for spinal cord + simple shapes (license MIT)
- FSL BET  (non-commercial): for brain
- Others?

## Suggestion
- Add disk and ball to `shimmingtoolbox.masking.shape
- Add capability to handle phase images to `shimmingtoolbox.masking.threshold
- `shimmingtoolbox.masking.bet()` --> input is file name
- `shimmingtoolbox.masking.sct()` --> input could be file name or an np.array (if we use SCT's API)
output: mask as nd.array (of file in the case of bet)

4rnaudB:
#542

PR #137, opened 2020-10-08
Add square, cube and threshold cli

### LINKED ISSUE #88: Create CLI for creating mask
## Context

Command Line Interface allow to use the python software on the command line. Currently, the masking methods implemented are [threshold](https://github.com/shimming-toolbox/shimming-toolbox-py/blob/master/shimmingtoolbox/masking/threshold.py) and [shape](https://github.com/shimming-toolbox/shimming-toolbox-py/blob/master/shimmingtoolbox/masking/shapes.py) where shape can be a square or a cube. The goal is to implement a basic CLI that enables to call the different implemented masking methods while considering what will be implemented in the future (SCT, BET)

## Discussion
Suggestion for name:
- st_mask

Suggestions for flags:

~~~
-i <file> input
-method {sct, bet, shape, threshold}
-shape {cube, ball} Only for method=shape
-size <int> Radius of the shape in voxel (not mm). Only for method=shape
~~~

Potential new feature: have it in mm. To discuss...

## To get started
Where: [shimmingtoolbox/cli](https://github.com/shimming-toolbox/shimming-toolbox-py/tree/master/shimmingtoolbox/cli)
Example: [shimmingtoolbox/cli/download_data](https://github.com/shimming-toolbox/shimming-toolbox-py/blob/master/shimmingtoolbox/cli/download_data.py)
Useful link to start [contributing](https://shimming-toolbox.org/en/latest/3_contributing/CONTRIBUTING.html)

maheuss:
Create CLI for creating mask

Implementation of st_mask_cube, st_mask_threshold and st_mask_square which allow to apply these masks to the desired files. Each function has its own required input sizes.

Fixes #88
Related to #66


shimmingtoolbox/cli/st_mask_square.py
```diff
@@ -0,0 +1,46 @@
...
+                                                                                                "of the square along "
+                                                                                                "third dimension (in "
+                                                                                                "pixels)")
+@click.option("--len_dim2", '-l2', required=True, type=click.IntRange(0, 255, clamp=True), help="Length of the side "
+                                                                                                "of the square along "
+                                                                                                "third dimension (in "
+                                                                                                "pixels)")
+@click.option("--center_dim", nargs=2, default=0, help="Center of the square along first and second dimension (in "
+                                                       "pixels). If no center is provided, the middle is used.")
+def st_mask_cube(filename, len_dim1, len_dim2, center_dim):
```
po09i:
```suggestion
def st_mask_square(filename, len_dim1, len_dim2, center_dim):
```

shimmingtoolbox/cli/st_mask_square.py
```diff
@@ -0,0 +1,46 @@
...
+                len_dim2 (int): Length of the side of the square along second dimension (in pixels).
+                center_dim (int): Center of the square along first and second dimension (in pixels). If no center is
+                                provided, the middle is used.
+
+            Return:
+                mask_square (numpy.ndarray): Mask with booleans.
+            """
+    img_nifti = nib.load(filename)
+    data = img_nifti.get_fdata()  # convert nifty file to numpy array
+    mask_square = np.array([[[]]])  # initialization of 3D array
```
po09i: Initialize to the correct size. You can use `np.zeros_like` I think

shimmingtoolbox/cli/st_mask_square.py
```diff
@@ -0,0 +1,46 @@
...
+
+    for z in range(data.shape[2]):
```
po09i: Square will mostly be used for 2d images, so I would make sure it works for 3d but this will crash for arrays with only 2 dimensions since shape[2] will not exist

shimmingtoolbox/cli/st_mask_cube.py
```diff
@@ -0,0 +1,45 @@
...
+            len_dim1 (int): Length of the side of the cube along first dimension (in pixels).
+            len_dim2 (int): Length of the side of the cube along second dimension (in pixels).
+            len_dim3 (int): Length of the side of the cube along third dimension (in pixels).
+            center_dim (int): Center of the cube along first, second and third dimension (in pixels). If no center is
+                            provided, the middle is used.
+
+        Return:
+            mask_cube (numpy.ndarray): Mask with booleans. True where the cube is located and False in the background.
+        """
+    img_nifti = nib.load(filename)
```
po09i: You can look at the wiki for variable name conventions. In this case, it should be `nii_data`

setup.py
```diff
@@ -22,6 +22,9 @@
             "st_referencemaps=shimmingtoolbox.cli.referencemaps:main",
             "st_b0maps=shimmingtoolbox.cli.b0map:main",
             "st_download_data=shimmingtoolbox.cli.download_data:download_data",
+            "st_mask_cube=shimmingtoolbox.cli.mask_cube:mask_cube",
```
jcohenadad: don't we want to instead deal with a single CLI called `st_mask`, and input the shape as param?

setup.py
```diff
@@ -38,6 +41,7 @@
         "matplotlib~=3.1.2",
         "pytest~=4.6.3",
         "pytest-cov~=2.5.1",
+        "SCT~=4.3",
```
po09i: Wrong PR?

setup.py
maheuss: Feasible with the click.group command

setup.py
```diff
@@ -22,6 +22,7 @@
             "st_referencemaps=shimmingtoolbox.cli.referencemaps:main",
             "st_b0maps=shimmingtoolbox.cli.b0map:main",
             "st_download_data=shimmingtoolbox.cli.download_data:download_data",
+            "st_mask_shimmingtoolbox=shimmingtoolbox.cli.mask_shimmingtoolbox:main",
```
po09i:
```suggestion
            "st_mask=shimmingtoolbox.cli.mask:main",
```

The name of the file should change to fit the new name.

shimmingtoolbox/cli/mask_shimmingtoolbox.py
```diff
@@ -0,0 +1,149 @@
...
+
+from shimmingtoolbox.masking.threshold import threshold
+from shimmingtoolbox.masking.shapes import shape_square
+from shimmingtoolbox.masking.shapes import shape_cube
+
+CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
+
+
+@click.group()
+def main():
```
po09i: Can we add the same context_settings to be able to display -h

test/cli/test_cli_mask_shimmingtoolbox.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env python3
+
+from click.testing import CliRunner
+from shimmingtoolbox.cli.mask_shimmingtoolbox import cube, square, mask_threshold
```
po09i: I think this is bypassing the click.group. I'm thinking importing main (which could be called mask) and the particular  tests would need the correct argument ('cube', ...)

po09i (commented):
Just a couple of suggestions to make things more uniform.

- Name of CLI should not have shimmingtoolbox in the name. --> st_mask
- The main could be the same name as the file (mask).

Output: The CLI should save the mask as a nifti array using an option spcifying where to put the output file. The fullpath should be returned.

po09i:
LGTM!

shimmingtoolbox/cli/mask.py
```diff
@@ -0,0 +1,185 @@
...
+import os
+
+import shimmingtoolbox.masking.threshold
+from shimmingtoolbox.masking.shapes import shape_square
+from shimmingtoolbox.masking.shapes import shape_cube
+
+CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
+
+
+@click.group(context_settings=CONTEXT_SETTINGS, help=f"Creates a cube, square or threshold mask.")
```
jcohenadad:
```suggestion
@click.group(context_settings=CONTEXT_SETTINGS, help=f"Create a mask based on a specified shape (cube, square, etc.) or based on the thresholding of an input image.")
```

shimmingtoolbox/cli/mask.py
```diff
@@ -0,0 +1,185 @@
...
+              help=f"Creates a cube mask from the input file. Returns an output nifti file with cube mask.")
+@click.option('-input', 'fname_input', type=click.Path(), required=True, help="Complete input path of the nifti file to"
+                                                                              " mask")
+@click.option('-output', type=click.Path(), default=os.curdir, help="Output folder for mask in nifti file")
+@click.option("-len_dim1", required=True, type=int, help="Length of the side of the cube along first dimension (in "
+                                                         "pixels)")
+@click.option("-len_dim2", required=True, type=int, help="Length of the side of the cube along second dimension (in "
+                                                         "pixels)")
+@click.option("-len_dim3", required=True, type=int, help="Length of the side of the cube along third dimension (in "
+                                                         "pixels)")
```
jcohenadad: in a cube, the three dimensions are the same, hence you only one flag

shimmingtoolbox/cli/mask.py
```diff
@@ -0,0 +1,185 @@
...
+@click.option("-center_dim1", type=int, default=None, help="Center of the cube along first dimension (in "
+                                                           "pixels). If no center is provided, the middle is "
+                                                           "used.")
+@click.option("-center_dim2", type=int, default=None, help="Center of the cube along second dimension (in "
+                                                           "pixels). If no center is provided, the middle is "
+                                                           "used.")
+@click.option("-center_dim3", type=int, default=None, help="Center of the cube along third dimension (in "
+                                                           "pixels). If no center is provided, the middle is "
+                                                           "used.")
```
jcohenadad: it's heavy to have to enter those dims separately. you could have a simpler approach:
~~~
st_mask -center x y z
~~~

shimmingtoolbox/cli/mask.py
jcohenadad: same comment as in https://github.com/shimming-toolbox/shimming-toolbox/pull/137/files#r519030186, lighter code and more intuitive to have (replacing len by size):
~~~
st_mask size x
~~~

also: avoid mixing - and _

shimmingtoolbox/cli/mask.py
```diff
@@ -0,0 +1,185 @@
...
+                center_dim1 (int): Center of the square along first dimension (in pixels). If no center is
+                                provided, the middle is used.
+                center_dim2 (int): Center of the square along second dimension (in pixels). If no center is
+                                provided, the middle is used.
+
+            Return:
+                output (str): Output nifti file with square mask.
+            """
+    # Create the folder where the nifti file will be stored
+    if not os.path.exists(fname_input):
```
jcohenadad: should be checked by the parser

jcohenadad: argparse can check for that-- i'm sure click can do it (otherwise that's a strong cons against click)

shimmingtoolbox/cli/mask.py
```diff
@@ -0,0 +1,185 @@
...
+        Args:
+            fname_input (str): Complete input path of the nifti file to mask.
+            output (str): Output folder for square mask.
+            thr: Value to threshold the data: voxels will be set to zero if their
+                value is equal or less than this threshold
+
+        Returns:
+            output (str): Filename for the output mask.
+        """
+    # Create the folder where the nifti file will be stored
```
jidicula: good use of comments

shimmingtoolbox/cli/mask.py
```diff
@@ -0,0 +1,185 @@
...
+
+    nii = nib.load(fname_input)
+    data = nii.get_fdata()  # convert nifti file to numpy array
+
+    mask_thr = shimmingtoolbox.masking.threshold.threshold(data, thr)  # creation of the threshold mask
+    mask_thr = mask_thr.astype(int)
+    nii_img = nib.Nifti1Image(mask_thr, nii.affine)
+    fname_mask = os.path.join(output, 'mask.nii.gz')
+    nib.save(nii_img, fname_mask)
+    click.echo(f"The filename for the output mask is: {os.path.abspath(fname_mask)}")
```
jidicula: I'm always happy to see f-strings!

jidicula (changes requested):
Looks good! Just one small change requested so we can keep our docs up-to-date with the library code:

Could you add this CLI to the documentation? An example can be seen [here](https://github.com/shimming-toolbox/shimming-toolbox/blame/master/docs/source/6_api_reference/api.rst#L51) for `download_data`. If you want to check that it builds correctly with Sphinx, you can `cd` to `docs/` and run `make html`, then check `shimming-toolbox/docs/build/html/6_api_reference/api.html` in any web browser. If you're not able to build the docs locally, you can push your .rst changes anyway and check the branch preview on ReadTheDocs: 
Screen Shot 2020-11-08 at 14 27 05

jcohenadad (changes requested):
I left a few comments that need to be addressed before being able to merge

shimmingtoolbox/cli/mask.py
maheuss: Is it the same principle for the square mask or do we have to consider 2 sizes for this one?

jcohenadad: same principle

shimmingtoolbox/cli/mask.py
```diff
@@ -0,0 +1,153 @@
...
+        fname_mask = os.path.join(output, 'mask.nii.gz')
+        nib.save(nii_img, fname_mask)
+        click.echo(f"The filename for the output mask is: {os.path.abspath(fname_mask)}")
+        return output
+
+    else:
+        raise ValueError("The nifti file does not have 3 dimensions.")
+
+
+@mask.command(context_settings=CONTEXT_SETTINGS,
```
jidicula: I think this should be `@click.command()` instead of `@mask.command()` 😅 
```suggestion
@click.command(context_settings=CONTEXT_SETTINGS,
```

jidicula: (this is what was causing the RTD build to fail)

jidicula (changes requested):
Small change requested to fix a decorator typo. Once fixed, RTD build should succeed.

shimmingtoolbox/cli/mask.py
maheuss: Knowing that I use click.group(), each subcommand must be implemented with name_main_function.command(). In this case, the parent function is mask so each subcommand is implemented with mask.command(). I tried multiple methods to document all these functions with automodule but none of them managed to display mask and its subcommands. Otherwise, there is a sphinx extension specially designed for the click library (sphinx_click.ext) and I managed to generate the documentation with it. The problem with this alternative is that the layout is not exactly the same as with automodule.

docs/source/conf.py
```diff
@@ -33,7 +33,7 @@
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
-extensions = ["sphinx.ext.githubpages", "sphinx.ext.napoleon"]
+extensions = ["sphinx.ext.githubpages", "sphinx.ext.napoleon", "sphinx_click.ext"]
```
po09i, line 36: sphinx_click needs to be installed with pypi. So you need to add it to the `setup` file within extra_requires: docs.

docs/source/6_api_reference/api.rst
```diff
@@ -56,3 +56,7 @@ misc
 
 .. automodule:: shimmingtoolbox.cli.referencemaps
    :members:
+
+.. click:: shimmingtoolbox.cli.mask:mask_cli
```
po09i: `mask_cli` does not exist, it is named `mask`.

docs/source/conf.py
maheuss: Added now !

docs/source/6_api_reference/api.rst
maheuss: The name of my function in mask.py is now changed to mask_cli

shimmingtoolbox/cli/mask.py
maheuss: I adapted it this way, thank you!

shimmingtoolbox/cli/mask.py
maheuss: I removed it as it was not necessary

shimmingtoolbox/cli/mask.py
```diff
@@ -0,0 +1,153 @@
...
+    pass
+
+
+@mask_cli.command(context_settings=CONTEXT_SETTINGS,
+              help=f"Create a cube mask from the input file. Return the filename for the output mask.")
+@click.option('-input', 'fname_input', type=click.Path(), required=True, help="Input path of the nifti file to mask")
+@click.option('-output', type=click.Path(), default=os.curdir, help="Output folder for mask in nifti file")
+@click.option("-size", required=True, type=int, help="Length of the side of the cube along first, second and third "
+                                                     "dimension (in pixels)")
+@click.option("-center", nargs=3, type=int, default=(None, None, None), help="Center of the cube along first, second "
```
jidicula: @jcohenadad I wonder if the triple-quoted docstrings for click functions become redundant now that we use `sphinx_click` extension. can we forego those docstrings for any click functions and document them through these decorators instead?

jidicula (approved):
Travis and RTD builds look good to me - good call with using the sphinx_click extension. This makes much more sense for autogenerating documentation of the CLI modules going forward.

I'll leave it to others to review whether the actual mask generation UX is sound - it's outside my expertise 😅

jcohenadad (changes requested):
The CLI doc is not very helpful:
image

test/cli/test_cli_mask.py
```diff
@@ -0,0 +1,53 @@
...
+
+        inp = os.path.join(__dir_testing__, 'sub-fieldmap', 'fmap', 'sub-fieldmap_phase1.nii.gz')
+        out = os.path.join(tmp, 'nifti1')
+        size = 5
+
+        result = runner.invoke(mask_cli, ['cube', '-input', inp, '-output', out, '-size', size])
+
+        assert result.exit_code == 0
+        assert result is not None
+        assert len(os.listdir(out)) != 0
```
jcohenadad: it would be nice to check that the expected mask is created

shimmingtoolbox/cli/mask.py
```diff
@@ -0,0 +1,153 @@
...
+def mask_cli():
+    pass
+
+
+@mask_cli.command(context_settings=CONTEXT_SETTINGS,
+              help=f"Create a cube mask from the input file. Return the filename for the output mask.")
+@click.option('-input', 'fname_input', type=click.Path(), required=True, help="Input path of the nifti file to mask")
+@click.option('-output', type=click.Path(), default=os.curdir, help="Output folder for mask in nifti file")
+@click.option("-size", required=True, type=int, help="Length of the side of the cube along first, second and third "
+                                                     "dimension (in pixels)")
```
jcohenadad: every edge of a cube have the same length, so no need to mention the 1st 2nd and 3rd dim.
also, use single quote to be consistent (and as per guidelines)
```suggestion
@click.option('-size', required=True, type=int, help="Edge size of the cube (in pixels)")
```

shimmingtoolbox/cli/mask.py
jcohenadad:
```suggestion
@click.option('-center', nargs=3, type=int, default=(None, None, None), help="Center of the cube along first, second "
```

shimmingtoolbox/cli/mask.py
```diff
@@ -0,0 +1,153 @@
...
+        """
+    # Create the folder where the nifti file will be stored
+    if not os.path.exists(output):
+        os.makedirs(output)
+
+    nii = nib.load(fname_input)
+    data = nii.get_fdata()  # convert nifti file to numpy array
+
+    if len(data.shape) == 3:
+        mask_cb = shape_cube(data, size, size, size, center[0], center[1], center[2])  # creation
```
jcohenadad: is this working if center is none?

shimmingtoolbox/cli/mask.py
```diff
@@ -0,0 +1,153 @@
...
+        # of the cube mask
+        mask_cb = mask_cb.astype(int)
+        nii_img = nib.Nifti1Image(mask_cb, nii.affine)
+        fname_mask = os.path.join(output, 'mask.nii.gz')
+        nib.save(nii_img, fname_mask)
+        click.echo(f"The filename for the output mask is: {os.path.abspath(fname_mask)}")
+        return output
+
+    else:
+        raise ValueError("The nifti file does not have 3 dimensions.")
```
jcohenadad, line 55: why not accepting 4d data? e.g. our temporal fieldmaps used for realtime shimming.
what do you think @po09i ?

shimmingtoolbox/cli/mask.py
```diff
@@ -0,0 +1,153 @@
...
+@click.option("-center", nargs=2, type=int, default=(None, None), help="Center of the cube along first and second "
+                                                                       "dimension (in pixels). If no center is "
+                                                                       "provided, the middle is used.")
+def square(fname_input, output, size, center):
+    """
+            Create a square mask from the input file. Return an output nifti file with square mask.
+
+            Args:
+                fname_input (str): Complete input path of the nifti file to mask.
+                output (str): Output folder for square mask.
```
jcohenadad: i thought that with the new sphinx_click, these three-quotes were no more required?

shimmingtoolbox/cli/mask.py
```diff
@@ -0,0 +1,153 @@
...
+
+@mask_cli.command(context_settings=CONTEXT_SETTINGS,
+              help=f"Create a square mask from the input file. Return an output nifti file with square mask.")
+@click.option('-input', 'fname_input', type=click.Path(), required=True, help="Input path of the nifti file to mask")
+@click.option('-output', type=click.Path(), default=os.curdir, help="Output folder for mask in nifti file")
+@click.option("-size", required=True, type=int, help="Length of the side of the square along first and second dimension"
+                                                     " (in pixels)")
+@click.option("-center", nargs=2, type=int, default=(None, None), help="Center of the cube along first and second "
+                                                                       "dimension (in pixels). If no center is "
+                                                                       "provided, the middle is used.")
```
jcohenadad: apply my comments raised for the cube case

shimmingtoolbox/cli/mask.py
```diff
@@ -0,0 +1,153 @@
...
+            mask_sqr = np.zeros_like(data)  # initialization of 3D array of zeros
+            img_2d = data[:, :, z]  # extraction of a MRI slice (2D)
+            mask_slice = shape_square(img_2d, size, size, center[0], center[1])  # creation of the mask
+            # on each slice (2D)
+            mask_sqr[:, :, z] = mask_slice  # addition of each masked slice to form a 3D array
+            mask_sqr = mask_sqr.astype(int)
+            nii_img = nib.Nifti1Image(mask_sqr, nii.affine)
+            nib.save(nii_img, fname_mask)
+            click.echo(f"The filename for the output mask is: {os.path.abspath(fname_mask)}")
+            return output
```
jcohenadad: this behaviour should be described in the usage

shimmingtoolbox/cli/mask.py
```diff
@@ -0,0 +1,153 @@
...
+
+    else:
+        raise ValueError("The nifti file does not have 2 or 3 dimensions.")
+
+
+@mask_cli.command(context_settings=CONTEXT_SETTINGS,
+              help=f"Create a threshold mask from the input file. Return an output nifti file with threshold mask.")
+@click.option('-input', 'fname_input', type=click.Path(), required=True, help="Input path of the nifti file to mask")
+@click.option('-output', type=click.Path(), default=os.curdir, help="Output folder for mask in nifti file")
+@click.option("-thr", default=30, help="Value to threshold the data: voxels will be set to zero if their value is "
```
jcohenadad:
```suggestion
@click.option('-thr', default=30, help="Value to threshold the data: voxels will be set to zero if their value is "
```

shimmingtoolbox/cli/mask.py
po09i: The mask is based off the API named cube which takes all 3 dimensions as inputs, the API could be named prism or rectangle to have a more appropriate name but to be consistent with the API I would leave it as is and revisit in another PR.

shimmingtoolbox/cli/mask.py
po09i: Indeed, it would be useful for masking the temporal data, I think it makes more sense to implement the 4d capability in the masking API rather than in the CLI.

shimmingtoolbox/cli/mask.py
jcohenadad: i see. in that case i would not call it “cube” because this is wrong. How about “box”?

and “rect” for the 2d version

shimmingtoolbox/cli/mask.py
maheuss: The shape_square and shape_cube functions are used on 2D and 3D arrays respectively, that's why I limited the use of CLI to 3D nifti files. I could extend and adapt the CLI to 4D arrays but I don't know if this would be optimal compared to adapting the masking API directly.

shimmingtoolbox/cli/mask.py
maheuss: The shape_cube function indicates that if the center (for a certain dimension) is None, then the middle of the dimension is used by default. The center is None for all 3 dimensions in my test_cli_mask_cube API and I got no error.

shimmingtoolbox/cli/mask.py
maheuss: Ok I will change the name of cube and square in this case!

shimmingtoolbox/cli/mask.py
jcohenadad: adapting the masking API would probably be better-- if everyone agrees pls open an issue and cross-reference it here so we don,t forget-- thanks!

shimmingtoolbox/cli/mask.py
maheuss: True, it's useless. I will remove it and be more precise in my decorators for a better documentation.

shimmingtoolbox/cli/mask.py
maheuss: Done now !

test/cli/test_cli_mask.py
maheuss: I'm going to think about how to add this check

docs/source/6_api_reference/api.rst
```diff
@@ -56,3 +56,10 @@ misc
 
 .. automodule:: shimmingtoolbox.cli.referencemaps
    :members:
+
+cli
+---
+
+.. click:: shimmingtoolbox.cli.mask:mask_cli
+   :prog: mask
+   :nested: full
```
jidicula: We have a new section for CLI documentation [here](https://github.com/shimming-toolbox/shimming-toolbox/blob/master/docs/source/6_cli_reference/cli.rst) - could you move these stubs into there?

jidicula (changes requested):

docs/source/6_api_reference/api.rst
maheuss: Done now, thank you for the information !

shimmingtoolbox/cli/mask.py
```diff
@@ -0,0 +1,139 @@
...
+
+@mask_cli.command(context_settings=CONTEXT_SETTINGS,
+                  help=f"Create a box mask from the input file. The nifti file is converted to a numpy array. If this "
+                       f"array is in 3D dimensions, then a binary mask is created from this array in the form of a box"
+                       f" with lengths defined in 'size'. This box is centered according to the 3 dimensions indicated"
+                       f" in 'center'. The mask is stored by default under the name 'mask.nii.gz' in the output folder."
+                       f"Return the filename for the output mask.")
+@click.option('-input', 'fname_input', type=click.Path(), required=True,
+              help="(str): Input path of the nifti file to mask. This nifti file must have 3D.")
+@click.option('-output', type=click.Path(), default=os.curdir,
```
po09i: Le output devrait être un nom de fichier (comme ça il sera possible d'appeler mask 2 fois de suite sans avoir un problèeme de duplication du nom de fichier) le default pourrait être: `default=os.path.join(os.curdir, 'mask.nii.gz')`. N'oublie pas de mentionné dans le help les extensions supportés soit `.nii` et `.nii.gz`.

test/cli/test_cli_mask.py
po09i: You can open it and make sure some pixels are what you are expecting

jcohenadad (approved):

po09i:
Remaining work to do summarized in this [issue](https://github.com/shimming-toolbox/shimming-toolbox/issues/192). This can be merged.

jidicula (approved):
LGTM 🚢

shimmingtoolbox/cli/mask.py
po09i: #193 opened

po09i (approved):

PR #141, opened 2020-10-09
Fixed biot-savart scaling

### LINKED ISSUE #140: Biot-savart coil generation code generates in T/A, not Hz/A
Minor change needed to scale final output before returning fieldmap.

ghost:
Added a scaling term to translate output from T/A to Hz/A, as is standard. Fixes #140

### DIFF
--- a/test/test_biot_savart.py
+++ b/test/test_biot_savart.py
@@ -6,10 +6,12 @@ import pytest
 
 from shimmingtoolbox.coils.biot_savart import biot_savart
 
+H_GYRO_R = 42.577478518e+6 # Hz/T
+
 dummy_data = [
-    ([(0, 0, -0.5)], [(0, 0, 1)], [1, ], [250, ], (0, 0, 0), (0, 0, 1), (1, 1, 3), (0.4495881427866065e-3,
-                                                                                    2.2214414690791835e-4,
-                                                                                    1.0723951147113824e-4)),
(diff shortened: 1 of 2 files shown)


jcohenadad (approved):
LGTM

test/test_biot_savart.py
```diff
@@ -23,6 +25,8 @@ def test_normal_use(centers, normals, radii, segment_numbers, fov_min, fov_max,
     assert(basis[:, :, :, 0].shape == fov_n)
 
     # Check expected values
-    actual = np.round(basis[:, :, :, 0].reshape(len(axis_answers)), decimals=7)
-    expected = np.round(np.asarray(axis_answers), decimals=7)
+    actual = np.round(basis[:, :, :, 0].reshape(len(axis_answers)), decimals=-2)
+    expected = np.round(np.asarray(axis_answers), decimals=-2)
+    print(actual)
+    print(expected)
```
po09i: Once we have a debug flag I would print the output if we are in debug mode, for now I would remove.

test/test_biot_savart.py
```diff
@@ -6,10 +6,12 @@
 
 from shimmingtoolbox.coils.biot_savart import biot_savart
 
+H_GYRO_R = 42.577478518e+6 # Hz/T
```
po09i: I would rename and update the value to be the same as the other file in this PR. This value has more decimal so I would update the other value.

po09i (approved):
LGTM

PR #142, opened 2020-10-10
Add ability to read from the PMU

### LINKED ISSUE #26: Adding the PMU association for probe
Working for the reworking and addition of the PMU association to the image data. Functions will be to:

- Read saved probe data and live incoming probe data
- Associate probe auxiliary data to the image
- Return the associated image and auxiliary information

If you have any suggestion, feel free to drop them here.

po09i:
## Description
This PR adds the ability to read Siemens respiratory files. It introduces the Class `PmuResp()`, with methods:
- `read_resp(fname)`: read PMU siemens file
- `interp_resp_trace(acq_time)`: Interpolate Siemens PMU data points onto the timing grid given as input argument. This will be used to match PMU trace with individual time point image.

## Linked issues
Fixes #26

### DIFF
--- a/shimmingtoolbox/pmu.py
+++ b/shimmingtoolbox/pmu.py
@@ -8,72 +8,111 @@
 import numpy as np
 
 
-def read_resp(fname_pmu):
+class PmuResp(object):
     """
-    Read a Siemens Physiological Log file. Returns a tuple
-    with the logging data as numpy integer array and times
-    in the form of milliseconds past midnight.
+    PMU object containing the pressure values of a Siemens .resp file
 
-    Args:
-        fname_pmu: srt or file object
-                   a file name to open or an already opened file-like object
-    Returns:
-        tuple:
-            data_trimmed: centered around 0 (goes from -2047 to 2048)
-            start_time_mdh
-            stop_time_mdh
-            start_time_mpcu
-            stop_time_mpcu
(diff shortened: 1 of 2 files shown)


shimmingtoolbox/pmu.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+    Read a Siemens Physiological Log file. Returns a tuple
+    with the logging data as numpy integer array and times
+    in the form of milliseconds past midnight.
+
+    Args:
+        fname_pmu: srt or file object
+                   a file name to open or an already opened file-like object
+    Returns:
+        tuple:
```
jcohenadad: wouldn't a dict provide more flexibility? Pros:
- you can easily add/insert a field without breaking everything
- you can add a description for each field (eg: key 'description', in addition to the key 'value')
- less prone to errors (eg: selecting the wrong index)

po09i: I'm considering moving to an o.o.p. approach for the PMU as it is its own entity with multiple attributes. Adding different methods for interpolating to the acquisition's timepoints would be neatly organized within the class.

jcohenadad: i guess that makes sense.

jcohenadad (approved):

PR #149, opened 2020-10-15
Add ability to get timing data from scans

### LINKED ISSUE #148: Fetch timing of image acquisition to be used with PMU

There is currently a PR which aims at reading PMU (resp trace) from the scanner https://github.com/shimming-toolbox/shimming-toolbox/pull/142. The PMU information includes absolute time `*_mdh`. It would be necessary to also fetch the absolute time of MR images.

Problems/considerations:
- the timing indicated in the .json sidecar (converted from dicom's "acquisition time"), e.g. `AcquisitionTime": "12:18:16.452500",` does not directly correspond to the `mdh` value of the PMU. So we need to figure out how to make them correspond. 
  - Useful link: https://cfn.upenn.edu/aguirre/wiki/public:pulse-oximetry_during_fmri_scanning
  - [MaRdI.m : estimating when k-space lines of an image were acquired](https://github.com/shimming-toolbox/shimming-toolbox-matlab/blob/7f0ba2d480a697a16a0d4b20154c61171f7e13ca/Img/MaRdI.m#L585)
  - [Read PMU file, cleanup metavalues and fetch metadata](https://github.com/shimming-toolbox/shimming-toolbox-matlab/blob/7f0ba2d480a697a16a0d4b20154c61171f7e13ca/aux_hardware/RespiratoryProbe/ProbeTracking.m#L1161)

- Add an "integration" test procedure that uses actual nifti and PMU files as input, and find a way to verify the validity (e.g.: the resp trace should somehow be correlated with the pixel content value).

### Additional information

The image timestamp should ideally be fetched via dcm2bids routine. This is currently not the case, but a [feature request](https://github.com/UNFmontreal/Dcm2Bids/issues/90) was posted.

po09i:
## Description
Realtime processing requires the extraction of the time from the scans. The PMU for example requires the time in milliseconds past midnight to interpolate the pressure data to those points. The times can easily be fetched using the dicom files but it becomes more tricky with the niftis.

## Changes
- Adds a function `get_acquisition_times(nii_data, json_data)` to easily get the timepoints of a time series using the json sidecar and nibabel object
- Adds a function `iso_times_to_ms(iso_times)` to convert dicom iso format time to milliseconds past midnight
- Adds a test `test_timing_images()` to look at the correlation between the PMU pressure and the fieldmap. See this https://github.com/shimming-toolbox/shimming-toolbox/pull/149#issuecomment-712450819

## Linked issues
Fixes #148

### DIFF
--- a/test/test_pmu.py
+++ b/test/test_pmu.py
@@ -88,26 +88,24 @@ def test_timing_images():
 
         return np.array(ms_times)
 
-    fname_pmu = os.path.join(__dir_testing__, 'realtime_zshimming_data', 'PMUresp_signal.resp')
-    pmu = PmuResp(fname_pmu)
-
-    fname_fieldmap = os.path.join(__dir_testing__, 'realtime_zshimming_data', 'nifti', 'sub-example', 'fmap',
-                                  'sub-example_fieldmap.nii.gz')
-    fname_json_phase_diff = os.path.join(__dir_testing__, 'nifti', 'sub-example', 'fmap',
-                                         'sub-example_phasediff.json')
-
     # get time between volumes and acquisition start time
+    fname_json_phase_diff = os.path.join(__dir_testing__, 'realtime_zshimming_data', 'nifti', 'sub-example', 'fmap',
+                                         'sub-example_phasediff.json')
     json_data = json.load(open(fname_json_phase_diff))
     delta_t = json_data['RepetitionTime'] * 1000  # [ms]
-    acq_start_time = json_data['AcquisitionTime']  # ISO format
-    acq_start_time = dicom_times_to_ms(np.array([acq_start_time]))[0]  # [ms]
+    acq_start_time_iso = json_data['AcquisitionTime']  # ISO format
+    acq_start_time_ms = dicom_times_to_ms(np.array([acq_start_time_iso]))[0]  # [ms]
 
     # Get the number of volumes
+    fname_fieldmap = os.path.join(__dir_testing__, 'realtime_zshimming_data', 'nifti', 'sub-example', 'fmap',
+                                  'sub-example_fieldmap.nii.gz')
     nii_fieldmap = nib.load(fname_fieldmap)
     n_volumes = nii_fieldmap.header['dim'][4]
 
-    # Get the pressure values at the iterpolated timestamps
-    fieldmap_timestamps = np.linspace(acq_start_time, ((n_volumes - 1) * delta_t) + acq_start_time, n_volumes)
+    # Get the pressure values at the interpolated timestamps
+    fname_pmu = os.path.join(__dir_testing__, 'realtime_zshimming_data', 'PMUresp_signal.resp')
+    pmu = PmuResp(fname_pmu)
+    fieldmap_timestamps = np.linspace(acq_start_time_ms, ((n_volumes - 1) * delta_t) + acq_start_time_ms, n_volumes)
     acquisition_pressures = pmu.interp_resp_trace(fieldmap_timestamps)
 
     # Get B0 data
--- a/test/test_pmu.py
+++ b/test/test_pmu.py
@@ -49,6 +49,7 @@ from matplotlib.figure import Figure
 from shimmingtoolbox import __dir_shimmingtoolbox__
 import nibabel as nib
 import json
+from shimmingtoolbox.masking.shapes import shapes
 
 
 def test_timing_images():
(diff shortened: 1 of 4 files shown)


jcohenadad:
One possibility would be to:
- create a module that reads dcm files
  - pydicom
- fetch necessary field (should be agnostic, to be used by other functions)
  - pydicom already solves this. 
```
a_dicom = pydicom.read_file(filename)
acq_time = a_dicom['AcquisitionTime'].value
```
- higher-level function includes that in the JSON produced by dcm2bids
  - Write in config file a custom field to be included in the bids sidecar. [This](https://github.com/UNFmontreal/Dcm2Bids/issues/22) may be of interest.
  - potential problem: There may be different acquisition within a given dicom directory (realtime_zshimming contains acq 15, 16 and 17). This means the acquistion times must be seperated by acquisitions (calling dicom_to_nifti for each acquisitions)

po09i:
My idea to implement the feature is to implement it within dicom_to_nifti. The function would:
- Read all the dicoms in the given directory (pydicom)
- Find the different acquisitions (15, 16 and 17 for realtime_zshimming_data)
- Edit the dcm2bids config file and replicate the description for each acquisition and adding
  - A criteria for a particular acquisition
  - Add a custom field in the config file to add acquisitionTimes custom field in the JSON produced by dcm2bids with the time information according to the acquisition (sidecarChanges)

jcohenadad:
@po09i as you probably saw in https://github.com/UNFmontreal/Dcm2Bids/issues/90#issuecomment-710402803 you can now rely on `TimeBetweenVolumes`.

po09i:
Since the data required needs the development version of dcm2niix, I suggest updating testing_data with the updated fieldmap data processed locally on the development version. There seems to be a consensus on the feature request page to use the `RepetitionTime` field to get the time between volumes.

In the meantime, here is the result of the sanity check:
- A mask was computed in the ROI:
mask
- Then the PMU values were interpolated to the fieldmap timeseries using the dcm2niix development version and using the `TimeBetweenVolumes` which will eventually become `RepetionTime` according to the feature request on dcm2niix.
- The average over the ROI of the fieldmap timeseries was calculated

Here are the results:

pmu_plot

test/test_pmu.py
```diff
@@ -44,3 +44,131 @@ def test_interp_resp_trace():
...
+import nibabel as nib
+import json
+from shimmingtoolbox.masking.shapes import shapes
+
+
+def test_timing_images():
+    """Check the matching of timing between MR images and PMU timestamps"""
+
+    # Convert to ms
+    def dicom_times_to_ms(dicom_times):
```
jcohenadad: that looks like something that could go in an API. Not sure where would make more sense.

po09i: That's why it is still in the test, my first thought was in /shimmingtoolbol/utils.py but it does not really fit with what is currently in there. Another option is in pmu.py but the file is already a class so I would prefer not mix a class and a standalone function.

test/test_pmu.py
```diff
@@ -44,3 +44,131 @@ def test_interp_resp_trace():
...
+        """
+        Convert dicom acquisition times to ms
+
+        Args:
+            dicom_times (numpy.ndarray): 1D array of time strings from dicoms.
+                                         Suported formats: "HHMMSS.mmmmmm" or "HH:MM:SS.mmmmmm
```
jcohenadad:
```suggestion
                                         Supported formats: "HHMMSS.mmmmmm" or "HH:MM:SS.mmmmmm
```

test/test_pmu.py
po09i: Related to #151

po09i:
The development branch of dcm2niix has now been updated. We can rely on `RepetitionTime`. I will update our testing data repo with the updated niftis processed by the development branch of dcm2niix.

rtopfer:
to be sure, the attempt in the [matlab code](https://github.com/shimming-toolbox/shimming-toolbox-matlab/blob/7f0ba2d480a697a16a0d4b20154c61171f7e13ca/Img/MaRdI.m#L585) to estimate when the central echo (k-origin) was sampled is fairly important since _that's_ the time that best represents the field map. most sequences begin by sampling at outskirts of k-space and _that's_ the time indicated by the DICOM AcquisitionTime field. presumably for spiral trajectories, AcquistionTime will actually correspond to the k-origin time; but our field mapping sequence isn't spiral, nor is the EPI, so the time-difference can be a few hundred ms. Relative to the avg. resp. cycle, that lag might be ~10% or more, which will mess up the fit of b0 field to probe/PMU trace and, consequentially, the shim corrections

test/test_pmu.py
po09i: Function moved to /utils

jcohenadad (approved):
nice 👍

PR #158, opened 2020-10-23
New example shell script for ASL scenario

jcohenadad:
This PR introduces a shell script for performing ASL experiment. The main specs are:
- dual shimming on two different ROIs
- output txt file that will be read by pulse sequence (that will play each shim config for the appropriate pulse)
- a cli for multi-region shimming: `/cli/multi_shim.py`

The expected output of the script is a CSV text file that will be read by the pulse sequence. E.g. for 2-region volume-wise shimming with 32 coils:
~~~
R1C1,R1C2, ... ,R1C32
R2C1,R2C2, ..., R2C32
~~~


examples/dual_shim_asl.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
```
jidicula:
```suggestion
#!usr/bin/env bash
```

jidicula: (more reliable for matching dev env with production installation behaviour)

examples/dual_shim_asl.sh
```diff
@@ -0,0 +1,48 @@
...
+# TODO
+
+# Create masks: mask1 for ASL labeling region, mask2: imaging region
+# TODO
+
+# Generate a coil profile based on custom coils
+# TODO
+
+# Generate coil profile from SH basis
+st_generate_profile_spherical_harmonics -i sub-example_fieldmap.nii.gz -order 2 -o coil_profile_sh2.nii.gz
```
jidicula:
```suggestion
st_generate_profile_spherical_harmonics -i "sub-example_fieldmap.nii.gz" -order "2" -o "coil_profile_sh2.nii.gz"
```

double-quote string literals

examples/dual_shim_asl.sh
```diff
@@ -0,0 +1,48 @@
...
+st_generate_profile_spherical_harmonics -i sub-example_fieldmap.nii.gz -order 3 -o coil_profile_sh3.nii.gz
```
jidicula:
```suggestion
st_generate_profile_spherical_harmonics -i "sub-example_fieldmap.nii.gz" -order "3" -o "coil_profile_sh3.nii.gz"
```

jidicula (changes requested):
Couple of small changes suggested for better durability of the script across more systems

examples/dual_shim_asl.sh
jcohenadad: the purpose of this shell script is to be used as a tutorial: users will see these command lines as what they are supposed to write in their Terminal. I don't want to "overweight" them with additional quotes if these are not required

jidicula: Sure thing!

PR #164, opened 2020-11-05
Create CLI for SCT mask #88

maheuss:
This PR implements masking of spinal coords with the use of SCT. 

The PR adds a CLI, st_mask sct

```
st_mask sct -h
Usage: st_mask sct [OPTIONS]

  Creates a mask around the spinal cord using the Spinal Cord Toolbox (SCT).
  The mask, which size can be specified, requires to identify the spinal
  cord centerline. The method of identification is specified by the flag '--
  centerline'. The output of this function is a NIfTI file containing the
  mask.

Options:
  --input PATH                    Input nifti file to mask. Must be 3D.
                                  Supported extensions are .nii or .nii.gz.
                                  Example: data.nii.gz  [required]

  --output PATH                   Name of output mask. Supported extensions
                                  are .nii or .nii.gz. Example: data.nii.
                                  [default: ./mask.nii.gz]

  --size INTEGER                  Size of the mask in the axial plane, given
                                  in pixel (Example: 35) or in millimeter
                                  (Example: 35mm). If shape=gaussian, size
                                  corresponds to sigma (Example: 45).
                                  [default: 20]

  --shape [cylinder|box|gaussian]
                                  Shape of the mask.
  --contrast [t1|t2|t2s|dwi]      Type of image contrast.  [default: t2s]
  --centerline [svm|cnn|viewer|file]
                                  Method used for extracting the centerline: -
                                  svm: Automatic detection using Support
                                  Vector Machine algorithm. - cnn: Automatic
                                  detection using Convolutional Neural
                                  Network. - viewer: Semi-automatic detection
                                  using manual selection of a few points with
                                  an interactive viewer followed by
                                  regularization. - file: Use an existing
                                  centerline (use with flag --file_centerline)
                                  [default: svm]

  --file-centerline PATH          Input centerline file. This option is only
                                  valid with '--centerline file'. Example:
                                  t2_centerline_manual.nii.gz

  --brain INTEGER RANGE           Set to 1 if the image contains the brain (or
                                  part of it), set to 0 otherwise (to speed up
                                  the segmentation). This option is only valid
                                  with '--centerline cnn'.

  --kernel [2d|3d]                Choice of kernel shape for the CNN.
                                  Segmentation with 3D kernels is slower than
                                  with 2D kernels.  [default: 2d]

  --remove-tmp BOOLEAN            Remove temporary files.  [default: True]
  --verbose INTEGER RANGE         Verbose: 0 = nothing, 1 = classic, 2 =
                                  expended.  [default: 1]

  -h, --help                      Show this message and exit.
```

It also adds the corresponding tests. We can now segment the spinal cord. SCT needs to be installed to run the command.

Related to #66


jcohenadad (changes requested):
there is a lot of duplicated code, which could be avoided by wrapping SCT's `sct_create_mask` only once and passing the function's argument, instead of creating a wrapper for each use of `sct_create_mask`.

shimmingtoolbox/cli/mask.py
```diff
@@ -0,0 +1,186 @@
+#!/usr/bin/env python3
+
+import click
+import nibabel as nib
+import numpy as np
+import os
+import sys
+import subprocess
+
+sys.path.insert(1, 'C:/Users/heuss/spinalcordtoolbox')
```
jcohenadad: shouldn't be required if SCT is callable in your SHELL environment (it should)

jcohenadad:
@maheuss you requested me for review but this PR is still in draft mode-- when ready for review, please remove the draft mode

jcohenadad:
Useful info: the relevant SCT function to wrap is `sct_deepseg_sc`: it segments the spinal cord automatically. From this segmentation we can the apply some morphomath, e.g. to dilate the output segmentation. This morphomat feature would probably be a separate API (which would be used by other masking functionalities, e.g. BET)

shimmingtoolbox/cli/sct_features.py
```diff
@@ -0,0 +1,184 @@
...
+              help="Shape of the structuring element for the mathematical morphology operation. Default: ball. If a 2D "
+                   "shape {'disk', 'square'} is selected, -dim must be specified. (default: ball)")
+@click.option('-dim', type=click.IntRange(0, 2),
+              help="Dimension of the array which 2D structural element will be orthogonal to. For example, if you wish "
+                   "to apply a 2D disk kernel in the X-Y plane, leaving Z unaffected, parameters will be: shape=disk, "
+                   "dim=2.")
+@click.option('-symmetrize', type=click.IntRange(0, 2), default=1, help="Symmetrize data along the specified "
+                                                                        "dimension.")
+@click.option('-verbose', type=click.IntRange(0, 2), default=1, help="Verbose: 0 = nothing, 1 = classic, 2 = expended.")
+def erode(fname_input, output, dilate, shape, dim, symmetrize, verbose):
```
po09i: You have two erode functions

shimmingtoolbox/cli/sct_features.py
```diff
@@ -0,0 +1,184 @@
...
+@click.option('-thr', type=float, required=True,
+              help="Dilate binary or greyscale image with specified size. If shape={'square', 'cube'}: size corresponds"
+                   "to the length of an edge (size=1 has no effect). If shape={'disk', 'ball'}: size corresponds to the"
+                   " radius, not including the center element (size=0 has no effect).")
+@click.option('-brain', type=click.IntRange(0, 1), help="Indicate if the input image contains brain sections (to speed "
+                                                        "up segmentation). Only use with '-centerline cnn'.")
+@click.option('-kernel', type=click.Choice(['2d', '3d']), default='2d',
+              help="Choice of kernel shape for the CNN. Segmentation with 3D kernels is slower than with 2D kernels. "
+                   "(default: 2d)")
+@click.option('-ofolder', type=click.Path(), default='/home/docs/checkouts/readthedocs.org/user_builds/'
```
po09i: Not sure if the default folder should be within SCT

shimmingtoolbox/cli/sct_features.py
```diff
@@ -0,0 +1,184 @@
...
+@click.option('-r', type=click.IntRange(0, 1), default=1, help="Remove temporary files. (default: 1)")
+@click.option('-v', type=click.IntRange(0, 2), default=1, help="1: display on (default), 0: display off, 2: extended "
+                                                               "(default: 1)")
+@click.option('-qc', help="The path where the quality control generated content will be saved")
+@click.option('-qc_dataset', help="If provided, this string will be mentioned in the QC report as the dataset the "
+                                  "process was run on")
+@click.option('-qc_subject', help="If provided, this string will be mentioned in the QC report as the subject the "
+                                  "process was run on")
+@click.option('-igt', help="File name of ground-truth segmentation.")
+def deepseg_sc(fname_input, c, centerline, file_centerline, thr, brain, kernel, ofolder, r, v, qc, qc_dataset, qc_subject, igt):
```
jcohenadad: deepseg_sc is a cord segmentation algo-- not a morphomath algo

what API will be calling deepseg_sc?

shimmingtoolbox/cli/sct_features.py
jcohenadad: why is it sct-specific? if it's only an erosion it could probably go somewhere else

jcohenadad (changes requested):
it is difficult to understand the overall logic of the proposed changes, and what will be the final organization/usage of those changes. 
what CLIs and APIs will be created?  when will they be called and by what routines? what is the overall logic / inter-relation of these APIs (eg: can the "erosion" be part of another sct-agnostic API that deals with morphomath stuff?).

i would strongly encourage you to first draft an algorithm for what you are planning to code, get feedback, and _then_ implement the code.

shimmingtoolbox/cli/sct_features.py
maheuss: I agree, I will change that. This branch will be more oriented towards the creation of an SCT mask and the get_centerline feature which is necessary to create a centerline process mask.

maheuss: I took it off, thank you

po09i:
I've been thinking hw we want to support sct in the future. @jcohenadad you mentioned about `sct_deepseg_sc`, since you know SCT, any other algorithms that you think we should implement in the future (sct_create_mask)? I would reorganize the code so that:
`st_create_mask sct -h` would list the commands available from sct that is implemented in the ST.

Any thoughts? @gaspardcereza @charjl

gaspardcereza:
I like the idea of listing the sct's functionalities that are implemented in the `ST` but how many of them would we expect ? If there are only a few of them, maybe they could be returned by a more general `st_create_mask -h` along with other non-sct masking functionalities.

jcohenadad:
a suggestion of backend for generating the mask:
```bash
sct_deepseg_sc -i t2.nii.gz -c t2
sct_create_mask -i t2_seg.nii.gz -p centerline,t2_seg.nii.gz -f cylinder -size 35mm 
```

In the future, we could improve robustness, e.g. if segmentation has missing slice, we could add morphomath before `sct_create_mask`, e.g.:
```bash
sct_maths -i SEG -dilate 3
```

Additional justifications:
- why not using `sct_get_centerline` instead of `sct_deepseg_sc`?
  - get_centerline is faster, however, it is a bit less accurate. More investigations needed in the future, esp. if we want to apply those algorithms on "unorthodox" data (e.g. poorly shimmed 7T data...)

jcohenadad:
if data are 4D, you could average them in time, and then run the segmentatino-- you will have better SNR, and potentially more robust SC segmentation. Also see [sct forum](https://forum.spinalcordmri.org/t/setorient-rpi-not-responding/608/5?u=jcohenadad)

If user inputs 4D data, the output should be a single 3D mask. However, it could be interested/relevant to have a mask per volume ---> should be further investigated in separate issue @po09i

shimmingtoolbox/cli/check_env.py
```diff
@@ -80,6 +80,19 @@ def check_dependencies():
         print_fail()
         print(f"Error {dcm2niix_exit_code}: dcm2niix is not installed or not in your PATH.")
 
+    # SCT
+    sct_check_msg = check_name.format("Spinal Cord Toolbox")
+    print_line(sct_check_msg)
+    # 0 indicates sct_check_msg is installed.
+    sct_exit_code: int = check_sct_installation()
+    # negating condition because 0 indicates sct is installed.
+    if not sct_exit_code:
```
charjl: This check_prelude_installation should return true or false, not a number because a not in a return in an if is not best practise. It makes it harder to read. You can do a try/catch in the check_sct_installation(), and return true there.

shimmingtoolbox/cli/check_env.py
```diff
@@ -80,6 +80,19 @@ def check_dependencies():
...
+        print_ok()
```
charjl: you can put these error/successes in check_sct_installation()as well

shimmingtoolbox/cli/check_env.py
```diff
@@ -108,6 +121,19 @@ def check_dcm2niix_installation() -> int:
...
+def check_sct_installation() -> int:
+    """Checks that ``SCT`` is installed.
+
+    This function calls ``which sct_check_dependencies`` and checks the exit code to verify
+    that ``sct`` is installed.
+
+    Returns:
+        int: Exit code. 0 on success, nonzero on failure.
+    """
+    return subprocess.check_call(['which', 'sct_check_dependencies'], stdout=subprocess.DEVNULL,
```
charjl: see comments on line 89

shimmingtoolbox/cli/check_env.py
```diff
@@ -149,6 +175,23 @@ def get_dcm2niix_version() -> str:
     return version_output
 
 
+def get_sct_version() -> str:
```
charjl: Just a comment: All of these SCT things should be pulled into a DDD domain. I believe I've already done this in my PR. but I'll check that on my end. check_env needs to call domains that have the logic for each of these that need to be checked, not just have a long file because it'll be easier to read/reuse

shimmingtoolbox/cli/check_env.py
```diff
@@ -149,6 +175,23 @@ def get_dcm2niix_version() -> str:
...
+
+    This function calls ``sct_check_dependencies -short`` and captures the output to
+    obtain the installation version.
+
+    Returns:
+        str: Version of the ``SCT`` installation.
+    """
+    # `sct_check_dependencies -short` returns
+    sct_version: str = subprocess.run(["sct_check_dependencies", "-short"], capture_output=True, encoding="utf-8")
+    assert sct_version.returncode == 0
```
charjl: Assert should only be used in debugging/tests. Raise an error instead. Asserts can be turned off "at runtime" via configuration. THis isn't fun to debug.

shimmingtoolbox/cli/mask.py
```diff
@@ -128,3 +129,65 @@ def threshold(fname_input, output, thr):
     nib.save(nii_img, output)
     click.echo(f"The filename for the output mask is: {os.path.abspath(output)}")
     return output
+
+
+@mask_cli.command(context_settings=CONTEXT_SETTINGS,
+                  help="Creates a SCT (SpinalCordToolbox) mask from the input file. Depending on the shape (cylinder,"
```
charjl: SO these help messages really clutter things up. I have a different way of doing these types of messages, which I'll be implementing in my PR and I'll fix these up, too.   Just helps readability

shimmingtoolbox/cli/mask.py
```diff
@@ -128,3 +129,65 @@ def threshold(fname_input, output, thr):
...
+                   "interpolate if slices are missing and extrapolate beyond the segmentation boundaries (i.e., every "
+                   "axial slice will exhibit a centerline pixel). (default: optic)")
+@click.option('-centerline_algo', type=click.Choice(['polyfit', 'bspline', 'linear', 'nurbs']), default='bspline',
+              help="(str): Algorithm for centerline fitting. Only relevant with -method fitseg (default: bspline)")
+@click.option('-centerline_smooth', default=30, help="(int): Degree of smoothing for centerline fitting. Only for "
+                                                     "-centerline-algo {bspline, linear}. (default: 30)")
+@click.option('-remove', type=click.IntRange(0, 1), default=1, help="(int): Remove temporary files. (default: 1)")
+@click.option('-verbose', type=click.IntRange(0, 2), default=1,
+              help="(int): Verbose: 0 = nothing, 1 = classic, 2 = expended. (default: 1)")
+def sct(fname_input, output, method, contrast, centerline_algo, centerline_smooth, size, shape, remove, verbose):
```
charjl: needs to be refactored out of this file.  I will do.

charjl: needs a more explanatory name; whats sct do? if this was being debugged it would be confusing

shimmingtoolbox/cli/mask.py
```diff
@@ -128,3 +129,65 @@ def threshold(fname_input, output, thr):
...
+
+    # Get the centerline
+    path_centerline = os.path.join(os.path.dirname(output), 'centerline')
```
charjl: should be a separate method. Code duplication can be shortened here to be more readable

shimmingtoolbox/cli/mask.py
```diff
@@ -128,3 +129,65 @@ def threshold(fname_input, output, thr):
...
+    elif method == "fitseg" and (centerline_algo == "polyfit" or centerline_algo == "nurbs"):
+        run_subprocess(f"sct_get_centerline -i {fname_input} -method {method} -centerline-algo {centerline_algo} "
+                       f"-o {path_centerline} -v {str(verbose)}")
+
+    elif method == "fitseg" and (centerline_algo == "bspline" or centerline_algo == "linear"):
+        run_subprocess(f"sct_get_centerline -i {fname_input} -method {method} -centerline-algo {centerline_algo} "
+                       f"-centerline-smooth {str(centerline_smooth)} -o {path_centerline} -v {str(verbose)}")
+
+    else:
+        raise ValueError("Could not get centerline.")
```
charjl: sure this is a value error? if it is, what value was wrong that caused it to fail? if it's not an input value error, pick one of these guys: https://docs.python.org/3/library/exceptions.html
If none of those fit we can custom make one

shimmingtoolbox/cli/mask.py
```diff
@@ -128,3 +129,65 @@ def threshold(fname_input, output, thr):
...
+
+    # Create the mask
```
charjl: create and remove should be in a domain file. I'll do this

test/cli/test_cli_mask.py
```diff
@@ -85,3 +85,65 @@ def test_cli_mask_threshold():
         assert result.exit_code == 0
         assert result is not None
         assert np.all(mask[58:62, 28:31, 7:9] == expected)
+
+
+def test_cli_mask_sct_default():
```
charjl: A recommendation is to comment above each test with the specific scenereo that would lead to that circumstance you're testing against.  eg: "User used input type incorrectly as x and not y" , "Input x was missing" etc. , "Happy Path" (works as expected).

po09i:
@jidicula The CI build was working in this commit b82823b (minus coveralls), then a change that should not have broken the PR did break it in commit 49b19b9. This seems to be related to the SCT installation and caching of SCT. I think this might be related to the path not being set when caching.
Screen Shot 2021-02-16 at 09 18 44

jidicula:
> @jidicula The CI build was working in this commit [b82823b](https://github.com/shimming-toolbox/shimming-toolbox/commit/b82823b1dfe3f015333fd9d1576edbe5bbdbb00d) (minus coveralls), then a change that should not have broken the PR did break it in commit [49b19b9](https://github.com/shimming-toolbox/shimming-toolbox/commit/49b19b91b4763fb060a2df068b28f3bdee764983). This seems to be related to the SCT installation and caching of SCT. I think this might be related to the path not being set when caching.
> Screen Shot 2021-02-16 at 09 18 44

Yep, looks like it. The fix would be to add another always-run step adding SCT to the PATH. This step would go after the "Spinal Cord Toolbox Installation" step that's run conditionally when the cache is invalid. I can work on this over my lunch.

jidicula:
Actually scratch that, I'll take a look this evening.

po09i:
@jidicula 
> Yep, looks like it. The fix would be to add another always-run step adding SCT to the PATH. This step would go after the "Spinal Cord Toolbox Installation" step that's run conditionally when the cache is invalid. I can work on this over my lunch.

I implemented your solution in commit d1ac404 and it used the cached version in commit 457450b. It's working! I fixed it in this PR since there was already some fixes to the SCT installation and it's the only PR that's using SCT so a lot easier to test the CI. Thanks for your input!

jidicula:
> @jidicula
> 
> > Yep, looks like it. The fix would be to add another always-run step adding SCT to the PATH. This step would go after the "Spinal Cord Toolbox Installation" step that's run conditionally when the cache is invalid. I can work on this over my lunch.
> 
> I implemented your solution in commit [d1ac404](https://github.com/shimming-toolbox/shimming-toolbox/commit/d1ac4043e54b2011dce62a5feff91b985efac621) and it used the cached version in commit [457450b](https://github.com/shimming-toolbox/shimming-toolbox/commit/457450b6a031d67ff4d7662cadfd546710c52353). It's working! I fixed it in this PR since there was already some fixes to the SCT installation and it's the only PR that's using SCT so a lot easier to test the CI. Thanks for your input!

Glad to help! 😁

jcohenadad (changes requested):
i find it difficult to find what are the required and optional arguments form the help:

<details><summary>current help</summary>

```
Options:
  --input PATH                    (str): Input nifti file to mask. Must be 3D.
                                  Supported extensions are .nii or .nii.gz.
                                  Example: data.nii.gz  [required]

  --output PATH                   (str): Name of output mask. Supported
                                  extensions are .nii or .nii.gz. Example:
                                  data.nii. (default: (os.curdir,
                                  'mask.nii.gz')).

  --size TEXT                     (str): Size of the mask in the axial plane,
                                  given in pixel (Example: 35) or in
                                  millimeter (Example: 35mm). If
                                  shape=gaussian, size corresponds to sigma
                                  (Example: 45). (default: 41)

  --shape [cylinder|box|gaussian]
                                  (str): Shape of the mask. (default:
                                  cylinder)

  --contrast [t1|t2|t2s|dwi]      (str): Type of image contrast. (default:
                                  t2s)

  --centerline [svm|cnn|viewer|file]
                                  (str): Algorithm for centerline fitting.
  --file-centerline TEXT          (str):  Input centerline file (to use with
                                  flag -centerline file). Example:
                                  t2_centerline_manual.nii.gz

  --thr FLOAT                     Binarization threshold (between 0 and 1) to
                                  apply to the segmentation prediction. Set to
                                  -1 for no binarization (i.e. soft
                                  segmentation output). The default threshold
                                  is specific to each contrast and
                                  wasestimated using an optimization
                                  algorithm. More details at:
                                  https://github.com/sct-pipeline/deepseg-
                                  threshold.

  --brain INTEGER RANGE           (str):  Indicate if the input image contains
                                  brain sections (to speed up segmentation).
                                  Only use with '-centerline cnn'.

  --kernel [2d|3d]                (str):  Choice of kernel shape for the CNN.
                                  Segmentation with 3D kernels is slower than
                                  with 2D kernels. (default: 2d)

  --remove INTEGER RANGE          (int): Remove temporary files. (default: 1)
  --verbose INTEGER RANGE         (int): Verbose: 0 = nothing, 1 = classic, 2
                                  = expended. (default: 1)
```

</details>

how about a separated section, with mandatory/optional args:

<details><summary>suggested help</summary>

```
MANDATORY ARGUMENTS:
  --input PATH                    (str): Input nifti file to mask. Must be 3D.
                                  Supported extensions are .nii or .nii.gz.
                                  Example: data.nii.gz  [required]

OPTIONAL ARGUMENTS:
  --output PATH                   (str): Name of output mask. Supported
                                  extensions are .nii or .nii.gz. Example:
                                  data.nii. (default: (os.curdir,
                                  'mask.nii.gz')).

  --size TEXT                     (str): Size of the mask in the axial plane,
                                  given in pixel (Example: 35) or in
                                  millimeter (Example: 35mm). If
                                  shape=gaussian, size corresponds to sigma
                                  (Example: 45). (default: 41)

  --shape [cylinder|box|gaussian]
                                  (str): Shape of the mask. (default:
                                  cylinder)

  --contrast [t1|t2|t2s|dwi]      (str): Type of image contrast. (default:
                                  t2s)

  --centerline [svm|cnn|viewer|file]
                                  (str): Algorithm for centerline fitting.
  --file-centerline TEXT          (str):  Input centerline file (to use with
                                  flag -centerline file). Example:
                                  t2_centerline_manual.nii.gz

  --thr FLOAT                     Binarization threshold (between 0 and 1) to
                                  apply to the segmentation prediction. Set to
                                  -1 for no binarization (i.e. soft
                                  segmentation output). The default threshold
                                  is specific to each contrast and
                                  wasestimated using an optimization
                                  algorithm. More details at:
                                  https://github.com/sct-pipeline/deepseg-
                                  threshold.

  --brain INTEGER RANGE           (str):  Indicate if the input image contains
                                  brain sections (to speed up segmentation).
                                  Only use with '-centerline cnn'.

  --kernel [2d|3d]                (str):  Choice of kernel shape for the CNN.
                                  Segmentation with 3D kernels is slower than
                                  with 2D kernels. (default: 2d)

  --remove INTEGER RANGE          (int): Remove temporary files. (default: 1)
  --verbose INTEGER RANGE         (int): Verbose: 0 = nothing, 1 = classic, 2
                                  = expended. (default: 1)
```

</details>

shimmingtoolbox/cli/mask.py
```diff
@@ -128,3 +130,146 @@ def threshold(fname_input, output, thr):
     nib.save(nii_img, output)
     click.echo(f"The filename for the output mask is: {os.path.abspath(output)}")
     return output
+
+
+@mask_cli.command(context_settings=CONTEXT_SETTINGS,
+                  help="Creates a SCT (SpinalCordToolbox) mask from the input file. Depending on the shape (cylinder,"
+                       " box or Gaussian), a mask is created along z direction. To generate this mask, its center"
+                       " must be specified by the user according the method."
+                       " Return an output nifti file with SCT mask.")
```
jcohenadad:
```suggestion
                  help="Creates a mask around the spinal cord using the Spinal Cord Toolbox (SCT). The mask, which size can be specified, requires to identify the spinal cord centerline. The method of identification is specified by the flag '--centerline'. The output of this function is a NIfTI file containing the mask.")
```

shimmingtoolbox/cli/mask.py
```diff
@@ -128,3 +130,146 @@ def threshold(fname_input, output, thr):
...
+              help="(str): Name of output mask. Supported extensions are .nii or .nii.gz. Example: data.nii. (default:"
+                   " (os.curdir, 'mask.nii.gz')).")
+@click.option('--size', default='20',
+              help="(str): Size of the mask in the axial plane, given in pixel (Example: 35) or in millimeter "
+                   "(Example: 35mm). If shape=gaussian, size corresponds to sigma (Example: 45). (default: 41)")
+@click.option('--shape', type=click.Choice(['cylinder', 'box', 'gaussian']), default='cylinder',
+              help="(str): Shape of the mask. (default: cylinder)")
+@click.option('--contrast', type=click.Choice(['t1', 't2', 't2s', 'dwi']), default='t2s',
+              help="(str): Type of image contrast. (default: t2s)")
+@click.option('--centerline', type=click.Choice(['svm', 'cnn', 'viewer', 'file']), default='svm',
```
jcohenadad: explain what these options are

shimmingtoolbox/cli/mask.py
```diff
@@ -128,3 +130,146 @@ def threshold(fname_input, output, thr):
...
+              help="(str): Algorithm for centerline fitting.")
+@click.option('--file-centerline', 'file_centerline', type=str,
+              help="(str):  Input centerline file (to use with flag -centerline file). "
+                   "Example: t2_centerline_manual.nii.gz")
+@click.option('--thr', type=float,
+              help="Binarization threshold (between 0 and 1) to apply to the segmentation prediction. Set to -1 for no "
+                   "binarization (i.e. soft segmentation output). The default threshold is specific to each contrast "
+                   "and wasestimated using an optimization algorithm. More details at: "
+                   "https://github.com/sct-pipeline/deepseg-threshold.")
```
jcohenadad: we definitely don't need this parameter (this is for segmentation only)-- you can remove it and use the default

shimmingtoolbox/cli/mask.py
```diff
@@ -128,3 +130,146 @@ def threshold(fname_input, output, thr):
...
+@click.option('--brain', type=click.IntRange(0, 1),
+              help="(str):  Indicate if the input image contains brain sections (to speed up segmentation). "
+                   "Only use with '-centerline cnn'.")
+@click.option('--kernel', type=click.Choice(['2d', '3d']), default='2d',
+              help="(str):  Choice of kernel shape for the CNN. Segmentation with 3D kernels is slower than with "
+                   "2D kernels. (default: 2d)")
+@click.option('--remove', type=click.IntRange(0, 1), default=1, help="(int): Remove temporary files. (default: 1)")
```
jcohenadad: not a big fan of this flag name-- how about `--remove-tmp`, i find it more explicit-- that shoudl apply to all ST software-- so maybe put a note about harmonization somewhere...

shimmingtoolbox/cli/mask.py
```diff
@@ -128,3 +130,146 @@ def threshold(fname_input, output, thr):
...
+    if remove:
+        os.remove(fname_seg)
+        if ndim == 4:
+            os.remove(fname_mean)
+
+    click.echo(f"The path for the output mask is: {os.path.abspath(fname_output)}")
+    return fname_output
+
+
+# def _get_centerline(fname_process, fname_output, method='optic', contrast='t2', centerline_algo='bspline',
```
jcohenadad: needed?

shimmingtoolbox/cli/mask.py
```diff
@@ -128,3 +130,146 @@ def threshold(fname_input, output, thr):
...
+              help="(str): Size of the mask in the axial plane, given in pixel (Example: 35) or in millimeter "
+                   "(Example: 35mm). If shape=gaussian, size corresponds to sigma (Example: 45). (default: 41)")
+@click.option('--shape', type=click.Choice(['cylinder', 'box', 'gaussian']), default='cylinder',
+              help="(str): Shape of the mask. (default: cylinder)")
+@click.option('--contrast', type=click.Choice(['t1', 't2', 't2s', 'dwi']), default='t2s',
+              help="(str): Type of image contrast. (default: t2s)")
+@click.option('--centerline', type=click.Choice(['svm', 'cnn', 'viewer', 'file']), default='svm',
+              help="(str): Algorithm for centerline fitting.")
+@click.option('--file-centerline', 'file_centerline', type=str,
+              help="(str):  Input centerline file (to use with flag -centerline file). "
```
jcohenadad:
```suggestion
              help="(str):  Input centerline file. This option is only valid with '--centerline file'. "
```

shimmingtoolbox/cli/mask.py
```diff
@@ -128,3 +130,146 @@ def threshold(fname_input, output, thr):
...
+                   "Example: t2_centerline_manual.nii.gz")
+@click.option('--thr', type=float,
+              help="Binarization threshold (between 0 and 1) to apply to the segmentation prediction. Set to -1 for no "
+                   "binarization (i.e. soft segmentation output). The default threshold is specific to each contrast "
+                   "and wasestimated using an optimization algorithm. More details at: "
+                   "https://github.com/sct-pipeline/deepseg-threshold.")
+@click.option('--brain', type=click.IntRange(0, 1),
+              help="(str):  Indicate if the input image contains brain sections (to speed up segmentation). "
+                   "Only use with '-centerline cnn'.")
```
jcohenadad:
```suggestion
                   "This option is only valid with '--centerline cnn'.")
```

shimmingtoolbox/cli/mask.py
```diff
@@ -128,3 +130,146 @@ def threshold(fname_input, output, thr):
...
+@click.option('--file-centerline', 'file_centerline', type=str,
+              help="(str):  Input centerline file (to use with flag -centerline file). "
+                   "Example: t2_centerline_manual.nii.gz")
+@click.option('--thr', type=float,
+              help="Binarization threshold (between 0 and 1) to apply to the segmentation prediction. Set to -1 for no "
+                   "binarization (i.e. soft segmentation output). The default threshold is specific to each contrast "
+                   "and wasestimated using an optimization algorithm. More details at: "
+                   "https://github.com/sct-pipeline/deepseg-threshold.")
+@click.option('--brain', type=click.IntRange(0, 1),
+              help="(str):  Indicate if the input image contains brain sections (to speed up segmentation). "
```
jcohenadad:
```suggestion
              help="(int):  Set to 1 if the image contains the brain (or part of it), set to 0 otherwise (to speed up the segmentation). "
```

jcohenadad:
while reviewing this PR i noticed potential improvements in the CLI help, suggested in https://github.com/shimming-toolbox/shimming-toolbox/issues/228

shimmingtoolbox/cli/mask.py
po09i: This goes in line with this https://github.com/shimming-toolbox/shimming-toolbox/pull/164#issuecomment-768551211. More testing needs to be done in the future on whether to use sct_get_centerline or sct_deepseg_sc. Since we have more urgent features to work on, I'll open an issue on the matter and keep the code commented out for now.

shimmingtoolbox/cli/mask.py
```diff
@@ -128,3 +130,153 @@ def threshold(fname_input, output, thr):
...
     click.echo(f"The filename for the output mask is: {os.path.abspath(output)}")
     return output
+
+
+@mask_cli.command(context_settings=CONTEXT_SETTINGS,
+                  help="""Creates a mask around the spinal cord using the Spinal Cord Toolbox (SCT). The mask, which
+                   size can be specified, requires to identify the spinal cord centerline. The method of identification
+                   is specified by the flag '--centerline'. The output of this function is a NIfTI file containing the
+                   mask.""")
+@click.option('--input', 'fname_input', type=click.Path(), required=True,
```
evaalonsoortiz: Is there are reason why all the options for sct have a double dash?

shimmingtoolbox/cli/mask.py
```diff
@@ -128,3 +130,153 @@ def threshold(fname_input, output, thr):
...
+    # _get_centerline(fname_process, fname_seg)
+
+    # Run sct_deepseg_sc
+    # Use sct parameter convention
+    if remove_tmp:
+        remove = 1
+    else:
+        remove = 0
+
+    cmd = f"sct_deepseg_sc -i {fname_process} -o {fname_seg} -c {contrast} -centerline {centerline} -kernel {kernel} " \
```
evaalonsoortiz: sct_deepseg_sc does not have -o as an option. We can use -ofolder instead, in which case it seems that the centerline segmentation file will be named the same as the input file with _seg appended at the end.

shimmingtoolbox/cli/mask.py
```diff
@@ -128,3 +130,153 @@ def threshold(fname_input, output, thr):
...
+          f"-r {str(remove)} -v {str(verbose)}"
+    if centerline == 'file':
+        cmd += f" -file_centerline {file_centerline}"
+    if brain is not None and centerline == 'cnn':
+        cmd += f" -brain {brain}"
+
+    run_subprocess(cmd)
+
+    # Create the mask
+    run_subprocess(f"sct_create_mask -i {fname_process} -p centerline,{fname_seg} -size {size} -f {shape} "
```
evaalonsoortiz: fname_seg here needs to be replaced with the output file mentioned in my comment above.

shimmingtoolbox/cli/mask.py
po09i: We are moving towards having all double dash options. Here was the discussion #222

shimmingtoolbox/cli/mask.py
po09i: I do have the option. Here is the output of `sct_deepseg_sc -h`

<details>
  <summary>Terminal output</summary>
  
```bash
sct_deepseg_sc -h
--

usage: sct_deepseg_sc -i <file> -c {t1,t2,t2s,dwi} [-h] [-centerline {svm,cnn,viewer,file}] [-file_centerline <str>]
                      [-thr <float>] [-brain {0,1}] [-kernel {2d,3d}] [-ofolder <str>] [-o <file>] [-r {0,1}]
                      [-v <int>] [-qc <str>] [-qc-dataset <str>] [-qc-subject <str>] [-igt <str>]

Spinal Cord Segmentation using convolutional networks. Reference: Gros et al. Automatic segmentation of the spinal cord and intramedullary multiple sclerosis lesions with convolutional neural networks. Neuroimage. 2019 Jan 1;184:901-915.

MANDATORY ARGUMENTS:
  -i <file>             Input image. Example: t1.nii.gz
  -c {t1,t2,t2s,dwi}    Type of image contrast.

OPTIONAL ARGUMENTS:
  -h, --help            show this help message and exit
  -centerline {svm,cnn,viewer,file}
                        Method used for extracting the centerline:
                         svm: Automatic detection using Support Vector Machine algorithm.
                         cnn: Automatic detection using Convolutional Neural Network.
                         viewer: Semi-automatic detection using manual selection of a few points with an interactive
                         viewer followed by regularization.
                         file: Use an existing centerline (use with flag -file_centerline) (default: svm)
  -file_centerline <str>
                        Input centerline file (to use with flag -centerline file). Example:
                        t2_centerline_manual.nii.gz
  -thr <float>          Binarization threshold (between 0 and 1) to apply to the segmentation prediction. Set to -1
                        for no binarization (i.e. soft segmentation output). The default threshold is specific to
                        each contrast and wasestimated using an optimization algorithm. More details at:
                        https://github.com/sct-pipeline/deepseg-threshold.
  -brain {0,1}          Indicate if the input image contains brain sections (to speed up segmentation). Only use with
                        "-centerline cnn".
  -kernel {2d,3d}       Choice of kernel shape for the CNN. Segmentation with 3D kernels is slower than with 2D
                        kernels. (default: 2d)
  -ofolder <str>        Output folder. Example: My_Output_Folder/ (default:
                        /Users/alex/Documents/School/Polytechnique/Master/project)
  -o <file>             Output filename. Example: spinal_seg.nii.gz
  -r {0,1}              Remove temporary files. (default: 1)
  -v <int>              Verbosity. 0: Display only errors/warnings, 1: Errors/warnings + info messages, 2: Debug mode
                        (default: 1)
  -qc <str>             The path where the quality control generated content will be saved
  -qc-dataset <str>     If provided, this string will be mentioned in the QC report as the dataset the process was
                        run on
  -qc-subject <str>     If provided, this string will be mentioned in the QC report as the subject the process was
                        run on
  -igt <str>            File name of ground-truth segmentation.
```
</details>


I just updated SCT to the latest dev version (git-master-1eee837ba2531a392bea5cb09229dd55e65c49d5). You can see the -o option. Updating your SCT version should probably fix it? If that's the case, I'll update the doc to specify a version requirement.

shimmingtoolbox/cli/mask.py
po09i: See above comment

shimmingtoolbox/cli/mask.py
evaalonsoortiz: I Updated SCT and now have the -o option! No issues now.

po09i:
@evaalonsoortiz I opened #245 to help when we have low SNR images, I'll get to it in a separate PR when this one is merged since the fix is outside the scope of this PR.

kousu:
Please 2 reopen?

PR #168, opened 2020-11-10
Update generate meshgrid

### LINKED ISSUE #167: generate_meshgrid bug for some affine transformations
### Bug Description

When using generate_meshgrid for affine transformation that are not along the diagonal (example: realtime_zshimming fieldmap data). The function does not output the correct scanner coordinates.

### Reproduction Steps

`coord = generate_meshgrid( nii_fmap.get_fdata()[..., 0].shape, nii_fmap.affine)`

The values do not match the values from fsl_eyes

Order of operation of the affine matrix and the voxels

po09i:
## Description
The function `generate_meshgrid` allows to generate coordinates according to an affine transformation.

This PR fixes several problems with `generate_meshgrid`:
- The dot product was reversed, this is fixed by using `np.apply_affine` which does the dot product correctly. See #167 
- Casts `np.zeros_like` to floats to avoid rounding errors
- Fixes the test in `test_siemens_basis` to correct for the rounding errors
- Creates a test for `generate_meshgrid`

## Linked issues
Fixes #167

### DIFF
--- a/shimmingtoolbox/coils/coordinates.py
+++ b/shimmingtoolbox/coils/coordinates.py
@@ -2,8 +2,6 @@
 # -*- coding: utf-8 -*
 # Deals with coordinate systems, going from voxel-based to physical-based coordinates.
 
-# TODO: create a test for this API
-
 import numpy as np
 from nibabel.affines import apply_affine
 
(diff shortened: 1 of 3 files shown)


jcohenadad (approved):
i haven't tested but LGTM 👍

Issue #45, opened 2020-07-30
Implement another phase unwrapper


po09i:
Prelude is currently implemented but other algorithms are also available such as 

- SEGUE
- QGU

PR #169, opened 2020-11-10
Add ability to create fieldmaps

### LINKED ISSUE #162: Implement function to prepare fieldmap
In relation to this epic PR https://github.com/shimming-toolbox/shimming-toolbox/pull/92, we would need a function that does the following:
https://github.com/shimming-toolbox/shimming-toolbox/blob/492f38c156f62cb4c7a0b8948349663566d3aa35/examples/demo_realtime_zshimming.sh#L19-L37

We could create a CLI, that could be named similarly to FSL's `fsl_prepare_fieldmap`, eg: `st_prepare_fieldmap`.

po09i:
## Description
Fieldmaps are at the base of shimming. We currently use fsl to generate fieldmaps. This should be changed with our own API to generate fieldmaps. This PR adds the ability to generate fieldmaps from phase and magnitude images. 

## Suggested design

- Input: 
  - `phase`: list of NIfTI phase files
  - `-mag`: magnitude image (Could be used by some uwrappers, e.g. prelude)
  - `-unwrapper`: method for unwrapping
  - `-output`: Output filename for the fieldmap
  - `-mask`: input path for a mask.
  - `-threshold`: Threshold for masking
- nb of echoes
  - 1 echo: we assume the input phase is in fact a phase difference, so we only scale it (see below)
  - 2 echoes: compute the diff, divide by echo time and scale
  - 3+: currently not supported, but we could imagine plugging different algorithm, which will be specified by input flag `-method`.
- Output:
  - Write fieldmap in Hz as a NIfTI file.

TODO:
- [x]  `prepare_fieldmap` API
- [x] `prepare_fieldmap_cli` that wraps `prepare_fieldmap` API
- [x] `unwrap_phase` API
- [x] Tests for the CLI and API
- [x] Check if possible to change `phase` as required "option" with click, as per our convention -- Not possible

## Linked issues
Fixes #162


po09i:
Added a wrapper to the different unwrappers as suggested in #45

jcohenadad (approved):

PR #175, opened 2020-11-12
Fix singleton on last dimension error in prelude

### LINKED ISSUE #174: Prelude does not output the same shape as the input
### Bug Description
When prelude has a phase input with a singleton in the last dimension (i.e.` wrapped_phase.shape = [5, 5, 1]`), the return unwrapped phase drops the last dimension: ` unwrapped_phase.shape = [5, 5]`.

### Suggested fix
Look at the input and add a dimension if the singleton is dropped.

po09i:
## Description
This PR addresses an issue in the prelude wrapper that returned the wrong number of dimensions if the input had a singleton in the last dimension. The fix looks at the input and compares it to the output to return the appropriate number of dimensions. This allows to also fix a case where both last dimensions would be 1. I've added tests for all cases (X,Y,1), (X,1) and (X,1,1)

## Linked issues
Fixes #174

### DIFF
--- a/shimmingtoolbox/unwrap/prelude.py
+++ b/shimmingtoolbox/unwrap/prelude.py
@@ -10,6 +10,7 @@ import nibabel as nib
 import pathlib
 import tempfile
 import logging
+import numpy as np
 
 from shimmingtoolbox.utils import run_subprocess
 
@@ -80,4 +81,9 @@ def prelude(wrapped_phase, mag, affine, mask=None, threshold=None, is_unwrapping
 
     fname_phase_unwrapped = glob.glob(os.path.join(path_tmp, 'rawPhase_unwrapped*'))[0]
 
-    return nib.load(fname_phase_unwrapped).get_fdata()
+    phase_unwrapped = nib.load(fname_phase_unwrapped).get_fdata()
+
+    for _ in range(wrapped_phase.ndim - phase_unwrapped.ndim):
+        phase_unwrapped = np.expand_dims(phase_unwrapped, -1)
+
+    return phase_unwrapped
--- a/shimmingtoolbox/unwrap/prelude.py
+++ b/shimmingtoolbox/unwrap/prelude.py
@@ -81,8 +81,9 @@ def prelude(wrapped_phase, mag, affine, mask=None, threshold=None, is_unwrapping
 
     fname_phase_unwrapped = glob.glob(os.path.join(path_tmp, 'rawPhase_unwrapped*'))[0]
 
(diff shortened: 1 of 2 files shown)


shimmingtoolbox/unwrap/prelude.py
```diff
@@ -80,4 +81,9 @@ def prelude(wrapped_phase, mag, affine, mask=None, threshold=None, is_unwrapping
 
     fname_phase_unwrapped = glob.glob(os.path.join(path_tmp, 'rawPhase_unwrapped*'))[0]
 
-    return nib.load(fname_phase_unwrapped).get_fdata()
+    phase_unwrapped = nib.load(fname_phase_unwrapped).get_fdata()
+
```
jcohenadad:
```suggestion
```

shimmingtoolbox/unwrap/prelude.py
```diff
@@ -80,4 +81,9 @@ def prelude(wrapped_phase, mag, affine, mask=None, threshold=None, is_unwrapping
 
     fname_phase_unwrapped = glob.glob(os.path.join(path_tmp, 'rawPhase_unwrapped*'))[0]
 
-    return nib.load(fname_phase_unwrapped).get_fdata()
+    phase_unwrapped = nib.load(fname_phase_unwrapped).get_fdata()
```
jcohenadad, line 87: i would add a comment to justify the added code below

shimmingtoolbox/unwrap/prelude.py
```diff
@@ -80,4 +81,9 @@ def prelude(wrapped_phase, mag, affine, mask=None, threshold=None, is_unwrapping
...
+
+    for _ in range(wrapped_phase.ndim - phase_unwrapped.ndim):
+        phase_unwrapped = np.expand_dims(phase_unwrapped, -1)
```
jcohenadad, line 89: are you sure you need to run this code systematically? ie: the issue we've encountered with the removal of the last dim might be due to the fact that we used single slice scans

jcohenadad: my comment suggest better unit test coverage (eg run code with single-slice and multi-slice, to make sure expected behaviour is met)

po09i: > my comment suggest better unit test coverage (eg run code with single-slice and multi-slice, to make sure expected behaviour is met)

All tests in test_prelude introduced before this PR data that have shapes: [128, 76, 10], so they cover the multi-slice scenario. The three tests introduced in this PR have shapes: [128, 76, 1], [128, 1], [128, 1, 1], which cover the single slice scenarios.

jcohenadad (approved):

PR #178, opened 2020-11-20
Change prelude mag parameter to be an optional argument

po09i:
## Description
Fsl_prelude called by our `prelude` wrapper requires a `mag` input argument. If `mag` is not available or we don't want to mask the fieldmap, `mag` can be a matrix full of ones. 

This PR changes the default behaviour of `prelude` to not require the `mag` argument and sets the default value of mag to be a matrix full of ones.

The tests and different calls to `prelude` were changed to call prelude appropriately.


po09i:
Relevant for #169

jcohenadad (approved):
i would just change the title of the PR to state that `mag` is now "optional" (vs. "default")

PR #183, opened 2020-11-29
Add GitHub Actions CI builds

### LINKED ISSUE #180: Add support for macOS on Travis CI
### Feature Description:

To render our tests more sensitive to other platform (see e.g. https://github.com/shimming-toolbox/shimming-toolbox/pull/179), we should extend our CI to other platforms. 

The low hanging fruit is to add support to macOs since it is available in Travis.

jidicula:
## Checklist

## Description

Travis doesn't offer great support for macOS and is also beginning to
charge for macOS build minutes.

**What this change does**
 - Refactors the Travis CI build config into a shell script for unit tests
 - Calls that shell script in a GitHub Actions Workflow after setting up dependencies
 - Adds a matrix build in the GH Actions Workflow for macOS 10.15
 Catalina and 11.0 Big Sur (on x86_64 Intel chips), and Ubuntu 18, 20, and
 16
 - Adds a matrix build in the GH Actions Workflow for Python 3.7 and
 3.8
 - Adds a build status badge to the README

**Additional context/notes/links**
https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

## Linked issues

Resolves #180


jidicula:
@jcohenadad I think it's good to go now. All the GH Actions runs are passing (love those green checkmarks 😁) and the Python version build matrix looks fine too!

~Note that I've set up Coveralls to only run on the Ubuntu 18.04 build. The tiny decrease in coverage looks like it's only on Python 3.8 (did some detective-work with which build finished first and matched it with the reports on coveralls.io). There might be a way to do some kind of averaging over all the build matrix OS options but I need to look into it further – I think it's best to handle that in a different PR.~

The Coveralls builds run over all matrix builds now and I believe it averages coverage over each build parameter. 3.8 seems to have fewer lines run under the hood, giving it the appearance of having lower coverage. However, since the build matrix will be running in every PR CI check from now on, it's really just setting a new baseline coverage percentage that future PRs will be measured against.

po09i (changes requested):
I looked more closely at the builds and even though they say they pass, they all have errors within the CI run. I think its related to prelude and dcm2niix not being installed.

jidicula:
> I looked more closely at the builds and even though they say they pass, they all have errors within the CI run. I think its related to prelude and dcm2niix not being installed.

Oh good catch - maybe a `ci.sh` isn't such a great idea since it obscures this stuff... I'll try to fix this up

jidicula:
@po09i @jcohenadad I narrowed down the failing build to the macOS build needing the macOS `prelude` binary. I would like to add the macOS `prelude` binary to the `binaries` repo, then conditionally set up some steps in the GH Actions workflow to use the correct binary depending on what build matrix OS option is running. Could you give me push access to the `binaries` repo?

po09i (approved):
LGTM!

ci.sh
```diff
@@ -2,28 +2,17 @@
```
kousu: Instead of the `|| exit`s everywhere, per a tip [from MIT's CS club](https://sipb.mit.edu/doc/safe-shell/), you can put

```
set -euf -o pipefail
```

at the top of this script.

I am also 99% sure that GH Actions has the equivalent set; I've seen it stop builds early lots of times when any of the intermediate steps fail.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,106 @@
...
+      - name: dcm2niix macOS
+        if: contains(matrix.os, 'macos')
+        run: brew install dcm2niix
+
+      - name: dcm2niix Ubuntu
+        if: contains(matrix.os, 'ubuntu')
+        run: |
+          curl -JLO https://github.com/rordenlab/dcm2niix/releases/latest/download/dcm2niix_lnx.zip || exit "$?"
+          unzip -o dcm2niix_lnx.zip || exit "$?"
+          sudo install dcm2nii* /usr/bin/ || exit "$?"
```
kousu: You tagged me to review the last version but my only comment on that would've been that you'd downloaded the linux app even on macOS, but you've got that covered here :)

ci.sh
```diff
@@ -2,28 +2,17 @@
...
-
-# Tests
-echo "Download prelude"
-st_download_data prelude
+# Dependencies
 echo "Set up prelude"
 prelude_path="$(pwd)"/prelude
-chmod 500 "${prelude_path}"/prelude
+chmod +x "${prelude_path}"/prelude
 PATH=${prelude_path}:${PATH}
```
kousu: Ideally, prelude would exist as a standalone package we could use as a dependency, and installing it would look like installing dcm2niix. Why do you need to `+x` prelude but not dcm2niix?

kousu: Can you use `install` on prelude instead, and do it in the Workflow instead of in here?

ci.sh
jidicula: Good call, thanks! Resolved in ed38c19 and 32679c4

ci.sh
jidicula: For sure, resolved in 32679c4

PR #182, opened 2020-11-29
[need to redo] Add GitHub Actions CI builds

jidicula:
## Checklist

- [x] I've given this PR a concise, self-descriptive, and meaningful title
- [x] I've linked relevant issues in the PR body
- [x] I've applied the relevant labels to this PR
- [ ] I've added relevant tests for my contribution
- [ ] I've updated the documentation and/or added correct docstrings
- [ ] I've assigned a reviewer

## Description

**Why this change was necessary**
Travis doesn't offer great support for macOS and is also beginning to
charge for macOS build minutes.

**What this change does**
 - Factors the Travis CI build config into a shell script
 - Calls that shell script in a GitHub Actions Workflow
 - Adds a matrix build in the GH Actions Workflow for macOS 10.15
 Catalina and 11.0 Big Sur (on Apple Silicon), Windows Server 2019,
 and Ubuntu 18, 20, and 16
 - Adds a matrix build in the GH Actions Workflow for Python 3.7 and
 3.8
 - Adds a build status badge to the README

**Additional context/notes/links**
https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

## Linked issues

Resolves #180

PR #184, opened 2020-11-29
Installation of the SpinalCord Toolbox on GH Actions

maheuss:
Installs the most recent version of SCT through Github Actions. This software will be necessary for functions testing APIs using SCT (sct_create_mask, get_centerline, etc).

Coverall checks fail since code coverage decreases a bit.


jcohenadad (changes requested):
CI has now moved to GH actions https://github.com/shimming-toolbox/shimming-toolbox/pull/182

jidicula:
GH Actions are in *#183 as soon as that's approved. ~You can add the steps into `ci.sh` now though and it'll be added to the CI workflow.~ @maheuss you'll have to add the install step into `.github/workflows/tests.yml` - you can follow the structuring of [this](https://github.com/shimming-toolbox/shimming-toolbox/blob/master/.github/workflows/tests.yml#L43) step, and insert the SCT installation steps just before it.

po09i:
The build is failing on travis, it can't recognize `./install_sct`. My guess is you have to `cd` into the folder you just unzipped first.

jcohenadad:
> My guess is you have to cd into the folder you just unzipped first.

👍

.github/workflows/tests.yml
```diff
@@ -56,6 +56,14 @@ jobs:
           unzip -o dcm2niix_lnx.zip
           sudo install dcm2nii* /usr/bin/
 
+      - name: SpinalCord Toolbox v5.0.1 installation
```
jcohenadad: i suggest to install the bleeding edge (master) version so we don’t need to update that link every time SCT updates

maheuss: I agree, that's what I originally intended to do. However, the name of the zip file containing the latest version of SCT is changed each time, and so is its URL. I don't know how to call it in the same way as dcm2niix, with /releases/latest/download/name.zip for example

jcohenadad: you can simply use git, which is installed in the travis runner (note the depth 1 makes the download faster):
~~~
git clone --depth 1 https://github.com/neuropoly/spinalcordtoolbox
cd spinalcordtoolbox
yes | ./install_sct
~~~

.github/workflows/tests.yml
```diff
@@ -56,6 +56,22 @@ jobs:
...
+          yes | ./install_sct
+
+      - name: SpinalCord Toolbox v5.0.1 installation Ubuntu
+        if: contains(matrix.os, 'ubuntu')
+        run: |
+          curl -JLO https://github.com/neuropoly/spinalcordtoolbox/archive/5.0.1.zip
+          unzip -o spinalcordtoolbox-5.0.1.zip
+          cd spinalcordtoolbox-5.0.1
+          yes | ./install_sct
+
```
jidicula: If the same zip src is being installed for both macOS and Ubuntu, you can collapse these into a single Workflow step and remove the `if: contains()` statement:
```suggestion
      - name: SpinalCord Toolbox v5.0.1 installation
        run: |
          curl -JLO https://github.com/neuropoly/spinalcordtoolbox/archive/5.0.1.zip
          unzip -o spinalcordtoolbox-5.0.1.zip
          cd spinalcordtoolbox-5.0.1
          yes | ./install_sct

```

jidicula (changes requested):
A suggestion to simplify workflow steps since SCT installation is from the same OS-independent source archive.

.github/workflows/tests.yml
maheuss: Much more efficient! I'll try that, thanks

.github/workflows/tests.yml
maheuss: I tried it yesterday, it worked for the Ubuntu tests but the installation of SCT on macOS was just cancelled after running for 6 hours...

jidicula: Hmm yeah, looks like the recent commit is also taking a long time to install SCT on macOS 🤔

maheuss: Yes, I think it will give the same result again...

jidicula: I interrupted the build to get a peek at the logs... looks like the macOS build never reaches [this line](https://github.com/shimming-toolbox/shimming-toolbox/pull/184/checks?check_run_id=1507641746#step:8:2128) that's reached in the Ubuntu build.

maheuss: Exactly the same line that wasn't building when I was implementing SCT the way you proposed!

jidicula: It's interesting that macOS doesn't error out where Ubuntu does with the `yes: standard output: Broken pipe`. This hints that either `yes` or piping works differently on macOS. I wonder if it's related to this [issue](https://stackoverflow.com/a/20624315/6310633).

.github/workflows/tests.yml
```diff
@@ -56,6 +56,12 @@ jobs:
           unzip -o dcm2niix_lnx.zip
           sudo install dcm2nii* /usr/bin/
 
+      - name: SpinalCord Toolbox installation
+        run: |
+          git clone --depth 1 https://github.com/neuropoly/spinalcordtoolbox
+          cd spinalcordtoolbox
+          yes | ./install_sct
```
jidicula:
```suggestion
          yes 2>/dev/null | ./install_sct
```

Let's try the [StackOverflow solution](https://stackoverflow.com/a/20624315/6310633) to see if it's a peculiarity of `yes` throwing an error....

jidicula: I'm also running this locally (without redirecting stderr) on my Mac to see what happens)

jidicula: image
Looks like I made it past the conda installation, so that's already better than the GH Actions build

po09i:
@jidicula Adding `2>/dev/null` did seem to do something. Looking at the logs the linux builds seem to have errors but when you added `2>/dev/null` it did not, still not sure why on MacOS still fails.

I narrowed down the problem to this [line](https://github.com/neuropoly/spinalcordtoolbox/blob/e07cd93cdd9f856b326f81efdd470d10ea642af2/install_sct#L525) in the SCT installation script. Maybe we should raise the error with SCT if they have ideas.

jidicula:
> @jidicula Adding `2>/dev/null` did seem to do something. Looking at the logs the linux builds seem to have errors but when you added `2>/dev/null` it did not, still not sure why on MacOS still fails.

Are you referring to the broken pipe errors? I think those are intentional - `yes` stops running when the process it's being piped into ends, which also breaks the pipe.

.github/workflows/tests.yml
```diff
@@ -56,6 +56,16 @@ jobs:
           unzip -o dcm2niix_lnx.zip
           sudo install dcm2nii* /usr/bin/
 
+      - name: SpinalCord Toolbox installation
+        run: |
+          bash
+          cd ../..
+          git clone --depth 1 https://github.com/neuropoly/spinalcordtoolbox
+          cd spinalcordtoolbox
+          yes | ASK_REPORT_QUESTION=false PIP_PROGRESS_BAR=off ./install_sct
```
joshuacwnewton: Could you try [cloning the branch](https://stackoverflow.com/questions/1911109/how-do-i-clone-a-specific-git-branch) from https://github.com/neuropoly/spinalcordtoolbox/pull/3102 temporarily to test how that PR affects this issue?

```suggestion
          git clone --depth 1 --single-branch --branch ng/strict-shell https://github.com/neuropoly/spinalcordtoolbox
          cd spinalcordtoolbox
          ./install_sct -y
```

joshuacwnewton (commented):
Suggestion from https://github.com/neuropoly/spinalcordtoolbox/issues/3112#issuecomment-748491446.

.github/workflows/tests.yml
jidicula: So far so good on macOS 10.15 🎉

jidicula: And we have a winner on macOS 11.0!

joshuacwnewton: Thanks much for testing this out! :D

Just to fill in the context, @kousu identified the underlying problem here: https://github.com/actions/runner/issues/884

`yes |` was indeed the issue, as you had identified. :sweat_smile:

kousu: Well I don't know if I'd say I *identified* it, but it's narrowed down. I'm pretty sure it's Github's bug. But they missed it because they don't expect people to be doing something fragile like `yes |` in this day and age.

kousu: But the same bug should show up if someone uses `seq | while read; do ... if [something]; then break; fi; done`?

jidicula: Yeah, it's not specifically `yes`, rather the piping, correct? Because the same hang still happened when I tried piping the exact number of `'y'`s from `printf()`.

.github/workflows/tests.yml
```diff
@@ -56,6 +56,16 @@ jobs:
           unzip -o dcm2niix_lnx.zip
           sudo install dcm2nii* /usr/bin/
 
+      - name: SpinalCord Toolbox installation
+        run: |
+          bash
```
kousu: Hi there!

I'm guessing this was trying to make sure the script runs in bash and not some other language? I'm pretty sure that's unnecessary, everything already runs on bash anyway.

Plus, as written, all this does is run bash with no script and exit, then continue with the rest of your script. So:

```suggestion
```

If you do need to specify it in the future, you can instead write:

```
 - name: SpinalCord Toolbox installation
   shell: bash
   run: |
     ....
```

There are some docs here: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell

.github/workflows/tests.yml
```diff
@@ -56,6 +56,16 @@ jobs:
...
+          cd ../..
+          git clone --depth 1 https://github.com/neuropoly/spinalcordtoolbox
```
kousu, line 62: You can pick a specific release like this:

```suggestion
          git clone --depth 1 -b 5.0.1 https://github.com/neuropoly/spinalcordtoolbox
```

5.0.1 [is a tag](https://github.com/neuropoly/spinalcordtoolbox/releases/tag/5.0.1) but this works with branches too!

jidicula: 5.0.1 doesn't have those changes in the installer to allow -y to request a non-interactive install, does it?

kousu: I'm sorry you need to do this by the way. I want you be able to just include SCT in your setup.py, but we're [not there yet](https://github.com/neuropoly/spinalcordtoolbox/issues/1526).

.github/workflows/tests.yml
```diff
@@ -56,6 +56,15 @@ jobs:
           unzip -o dcm2niix_lnx.zip
           sudo install dcm2nii* /usr/bin/
 
+      - name: SpinalCord Toolbox installation
+        run: |
+          cd ../..
+          git clone --depth 1 -b 5.0.1 https://github.com/neuropoly/spinalcordtoolbox
```
kousu: Ah, sorry, the fix for https://github.com/neuropoly/spinalcordtoolbox/issues/3112 isn't released yet. You'll have to use `-b master` here. I guess we'll try to get a 5.0.2 out that has the fix in in a day or two.

.github/workflows/tests.yml
```diff
@@ -56,6 +56,15 @@ jobs:
...
+          cd spinalcordtoolbox
+          ./install_sct -y
+          cd ..
+          cd shimming-toolbox
```
kousu: You can probably get rid of this!

```suggestion
```

if you trace the code starting at

https://github.com/shimming-toolbox/shimming-toolbox/blob/ng/ci-test-cd/.github/workflows/tests.yml#L66

which produced

https://github.com/shimming-toolbox/shimming-toolbox/runs/1630308550?check_suite_focus=true#step:8:33

> /home/runner/work/spinalcordtoolbox

the next step is reset back to

https://github.com/shimming-toolbox/shimming-toolbox/runs/1630308550?check_suite_focus=true#step:11:9
> /home/runner/work/shimming-toolbox/shimming-toolbox



each step runs in a separate subshell, so it resets its paths and everything. Actions makes it actually pretty difficult to communicate any information or state at all between steps and jobs; They want you to use this weird thing: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-output-parameter

.github/workflows/tests.yml
kousu: We've solved this and merged a fix to `master` today: https://github.com/neuropoly/spinalcordtoolbox/commit/d64af23b69dbc396322cdc963a90a899b73736e0

That SO thread is interesting. I'm going to forward that to https://stackoverflow.com/questions/20573282/hudson-yes-standard-output-broken-pipe/20624315#20624315

If you install from SCT's `master` it should be good to go.

.github/workflows/tests.yml
kousu: Actually, that would be in line with what Julien suggested doing already: https://github.com/shimming-toolbox/shimming-toolbox/pull/184#discussion_r536935477, so win-win:

```suggestion
          git clone --depth 1 -b master https://github.com/neuropoly/spinalcordtoolbox
```

.github/workflows/tests.yml
kousu: No, it doesn't, of course. You're right.

po09i (approved):

jidicula (approved):
LGTM! 🎸 

Coveralls checks are failing even though no Python code was added, so coverage hasn't actually changed.

jcohenadad:
the coveralls error is likely related to an upgrade in coveralls, see https://github.com/ivadomed/ivadomed/pull/628

jcohenadad (approved):
LGTM!

po09i:
The threshold update needs to be done on coveralls website so this only reruns the tests and adds visibility. The build fails.. I think this might be the issue julien was referencing.

jidicula:
> The threshold update needs to be done on coveralls website so this only reruns the tests and adds visibility. The build fails.. I think this might be the issue julien was referencing.

Ah that's a smart call!

PR #189, opened 2020-12-02
Update prepare fieldmap 

po09i:
## Description
This PR fixes a couple of problems with the prepare_fieldmap API and CLI:
- Outputs a json sidecar along with the fieldmap 
- Fixes inconsistencies when loading nibabel objects
- Fixes a bug where the default output directory would crash


jcohenadad (approved):
👍

PR #191, opened 2020-12-03
Rename phasediff image in dicom_to_nifti

### LINKED ISSUE #190: dicom_to_nifti should output a filename with phasediff if the image is a phasediff
`dicom_to_nifti` should output a filename phasediff filename if the image is a phasediff. The current implementation outputs phase2.

Look within fmap folder, look for phase images, look in json file for fields EchoTime1 and EchoTime2, if both are there, change the filename to phasediff.

po09i:
## Description
This PR fixes an issue in dicom_to_nifti where phasediff images were wrongly indentified in their filenames as phase images instead of phasediff images.

The function looks at the output of dcm2bids and if
- It's in fmap directory
- The json file is a phase image, contains both fields EchoTime1 and EchoTime2
- It is not already named phasediff
- It contains phase in the file name
- SequenceName field's last digit is 2 (Refers to the number of echoes when using dcm2bids)

It will rename the json and the nifti to phasediff.

A test was changed to reflect the new names.

## Linked issues
Fixes #190

### DIFF
--- a/shimmingtoolbox/dicom_to_nifti.py
+++ b/shimmingtoolbox/dicom_to_nifti.py
@@ -13,6 +13,7 @@ import shutil
 
 from shimmingtoolbox import __dir_config_dcm2bids__
 
+
 def dicom_to_nifti(path_dicom, path_nifti, subject_id='sub-01', path_config_dcm2bids=__dir_config_dcm2bids__, remove_tmp=False):
     """ Converts dicom files into nifti files by calling dcm2bids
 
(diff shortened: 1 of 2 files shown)


shimmingtoolbox/dicom_to_nifti.py
```diff
@@ -54,6 +55,44 @@ def dicom_to_nifti(path_dicom, path_nifti, subject_id='sub-01', path_config_dcm2
...
+                        if len(file_parts) == 2:
+                            # Stitch the filename back together making sure to remove any digits that could be after
+                            # 'phase'
+                            digits = '0123456789'
+                            fname_new_json = file_parts[0] + 'phasediff' + file_parts[1].lstrip(digits)
+                            is_renaming = True
+            # Rename the json file
+            if is_renaming:
+                os.rename(fname_json, fname_new_json)
+                fname_nifti_new = os.path.splitext(fname_new_json)[0] + '.nii.gz'
```
jcohenadad: what if the input file is a nii instead of nii.gz?

po09i: I can account for it, however I think dcm2bids always outputs the '.nii.gz' version. So I'm not sure how I could test it. It will be there if dcm2bids ever puts the option to output a .nii

jcohenadad: i see, well in that case let's leave it for now
but i know many people who work with nii, so in case they use ST bypassing dcm2bids, then it will fail

po09i: If they do bypass dicom_to_nifti, then the renaming part of the code won't be executed as well.

jcohenadad: ok in that case let's not worry too much with it then-- i see you did add some code supporting nii, but your arguments are solid enough to not add these lines of code (it might create confusion from someone external reading the code-- moreover, if we want fulll NII compatibility maybe more code needs to be changed)

jcohenadad (changes requested):
add test of output JSON file for the fieldmap diff case

po09i:
> Add test of output JSON file for the fieldmap diff case

I may have misled you in our internal meeting. This PR only changes the name of the output phasediff from sub-XX_phase2.nii.gz and sub-XX_phase2.json to sub-XX_phasediff.nii.gz and sub-XX_phasediff.json. The relevant function which creates the json file for  the fieldmap is `prepare_fieldmap`. 

I opened an issue to add the test to the relevant function #194 
@jcohenadad

jcohenadad (approved):

PR #200, opened 2020-12-17
eao/gaussian-filter

evaalonsoortiz:
In this PR I have added the option to use gaussian filtering when preparing fieldmaps. This gaussian filtering option is now used in the realtime z-shimming shell script.


examples/realtime_zshimming.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
```
jcohenadad: what is the purpose of having this file in addition to `demo_realtime_zshimming.sh`? (they seem to overlap)

evaalonsoortiz:
Indeed, they are very similar. In the `demo_realtime_zshimming.sh` script sample data are downloaded, so I see that script as a true "demo" to show a potential user how the realtime z-shimming procedure works. `realtime_zshimming.sh` is my actual script, intended to process acquired data. I suppose it makes more sense for me not to include it on the repo at all. I've removed it.

po09i:
This is the script used to present the shell script in the abstract since this is the one actually used for zshimming. I think I would leave it.

examples/demo_realtime_zshimming.sh
```diff
@@ -22,7 +22,7 @@ st_prepare_fieldmap "sub-example_phasediff.nii.gz" -mag "sub-example_magnitude1.
 st_mask box -input "../anat/sub-example_unshimmed_e1.nii.gz" -size 15 15 20 -output "sub-example_anat_mask.nii.gz" || exit
 
 # Shim
-st_realtime_zshim -fmap "sub-example_fieldmap.nii.gz" -anat "../anat/sub-example_unshimmed_e1.nii.gz" -resp "../../../realtime_zshimming_data/PMUresp_signal.resp" -mask "sub-example_anat_mask.nii.gz" -output "." || exit
+st_realtime_zshim -fmap "sub-example_fieldmap.nii.gz" -anat "../anat/sub-example_unshimmed_e1.nii.gz" -resp "../../../realtime_zshimming_data/PMUresp_signal.resp" -mask "sub-example_anat_mask.nii.gz" -output "."|| exit
```
po09i: Add the space back

shimmingtoolbox/prepare_fieldmap.py
```diff
@@ -1,13 +1,14 @@
-#!/usr/bin/python3
+ #!/usr/bin/python3
```
po09i: Remove the space

shimmingtoolbox/prepare_fieldmap.py
```diff
@@ -1,13 +1,14 @@
...
 
 import math
 import numpy as np
+from skimage.filters import gaussian
 
 from shimmingtoolbox.unwrap.unwrap_phase import unwrap_phase
 
 
-def prepare_fieldmap(phase, echo_times, affine, unwrapper='prelude', mag=None, mask=None, threshold=None):
+def prepare_fieldmap(phase, echo_times, affine, unwrapper='prelude', mag=None, mask=None, threshold=None, gaussian_filter=None):
```
po09i: Can you add an argument description in the `Args` section

shimmingtoolbox/prepare_fieldmap.py
```diff
@@ -81,4 +82,9 @@ def prepare_fieldmap(phase, echo_times, affine, unwrapper='prelude', mag=None, m
     fieldmap_rad = phasediff_unwrapped / echo_time_diff  # [rad / s]
     fieldmap_hz = fieldmap_rad / (2 * math.pi)  # [Hz]
 
-    return fieldmap_hz
+    # Gaussian blur the fieldmap
+    if gaussian_filter == True:
+        fieldmap_hz = gaussian(fieldmap_hz, sigma = 1, mode = 'constant', cval = 0.0)
+
+    # return fieldmap_hz_gaussian
+    return fieldmap_hz
```
po09i: Add a blank line after the last line of a file

po09i (commented):
Some small python syntax fixups.

po09i:
I made minor changes to follow PEP8 convention that I had not picked up on my last review and added a test to make sure the implementation works. I realized that sigma seems to be pretty important in the gaussian filtering, would it be a good idea to replace `gaussian_filter` to gaussian `gaussian_filter_sigma` and implement it accordingly?

Quick thought as well, I think the gaussian function used applies the filter on all dimensions. This means a fieldmap (X, Y, Z, T) gets filtered for the whole 4d volume. This might me intended but just making sure slice by slice or 3d was not the intention.

shimmingtoolbox/prepare_fieldmap.py
```diff
@@ -81,4 +84,14 @@ def prepare_fieldmap(phase, echo_times, affine, unwrapper='prelude', mag=None, m
...
     fieldmap_hz = fieldmap_rad / (2 * math.pi)  # [Hz]
 
+    # Gaussian blur the fieldmap
+<<<<<<< HEAD
+    if gaussian_filter == True:
+        fieldmap_hz = gaussian(fieldmap_hz, sigma = 2, mode = 'constant', cval = 0.0)
+=======
+    if gaussian_filter is True:
+        fieldmap_hz = gaussian(fieldmap_hz, sigma=0.1, mode='constant', cval=0.0)
+>>>>>>> 744ea9d8bc5c8926d3b3b1f898bbd44abc1af1f4
```
po09i: Merge issue

evaalonsoortiz:
> I made minor changes to follow PEP8 convention that I had not picked up on my last review and added a test to make sure the implementation works. I realized that sigma seems to be pretty important in the gaussian filtering, would it be a good idea to replace `gaussian_filter` to gaussian `gaussian_filter_sigma` and implement it accordingly?
> 
> Quick thought as well, I think the gaussian function used applies the filter on all dimensions. This means a fieldmap (X, Y, Z, T) gets filtered for the whole 4d volume. This might me intended but just making sure slice by slice or 3d was not the intention.

Hum ... I had assumed 3d would be the maximum, but giving it some thought, I think 4d filtering makes sense. 

BTW, is there any way the other changes you made could have impacted my running the code? I haven't been able to run it lately. No errors, just nothing, it stalls and starts taking up all my computer's memory.

po09i:
@evaalonsoortiz 
> BTW, is there any way the other changes you made could have impacted my running the code? I haven't been able to run it lately. No errors, just nothing, it stalls and starts taking up all my computer's memory.

I did change sigma to see what it would do. (1 --> 0.1) This might be why it's taking longer to run. There is currently a merge issue which might be the problem as well. (at least the tests fail because of that)

evaalonsoortiz:
> @evaalonsoortiz
> 
> > BTW, is there any way the other changes you made could have impacted my running the code? I haven't been able to run it lately. No errors, just nothing, it stalls and starts taking up all my computer's memory.
> 
> I did change sigma to see what it would do. (1 --> 0.1) This might be why it's taking longer to run. There is currently a merge issue which might be the problem as well. (at least the tests fail because of that)

I found the problem. 

Call to run the script: `realtime_zshimming.sh run`, where `run` contains the fieldmap (mag/phase), GRE mag and PMU data. 

In order to keep track of different analyses I run, I save the `rt_zshim_nifti` output folder in the `run` folder. But, since the original dicom data in `run` is copied to `rt_zshim_nifti`, this was creating a situation where the origial data was being recursively copied with every call to `realtime_zshimming.sh`, effectively taking up my entire hard drive 🤦

evaalonsoortiz:
@po09i I'm ready to merge, however, I re-requested your review specifically because some checks have failed and I am not sure what it is I need to fix. Could you point me in the right direction to get them sorted?

shimmingtoolbox/cli/prepare_fieldmap.py
```diff
@@ -24,7 +24,9 @@
               help="Output filename for the fieldmap, supported types : '.nii', '.nii.gz'")
 @click.option('-mask', 'fname_mask', type=click.Path(exists=True), help="Input path for a mask. Used for PRELUDE")
 @click.option('-threshold', 'threshold', type=float, help="Threshold for masking. Used for: PRELUDE")
-def prepare_fieldmap_cli(phase, fname_mag, unwrapper, fname_output, fname_mask, threshold):
+@click.option('-gaussian_filter', 'gaussian_filter', type=bool, help="Gaussian filter for B0 map")
```
po09i:
```suggestion
@click.option('-gaussian-filter', 'gaussian_filter', type=bool, help="Gaussian filter for B0 map")
```

shimmingtoolbox/cli/prepare_fieldmap.py
```diff
@@ -24,7 +24,9 @@
...
+@click.option('-sigma', 'sigma', type=float, help="Standard deviation of gaussian filter. Used for: gaussian_filter")
```
po09i:
```suggestion
@click.option('-sigma', type=float, help="Standard deviation of gaussian filter. Used for: gaussian_filter")
```

FYI, if the name of the variable is the same as the name of the option, you don't need to specify it, it will automatically use that name.

shimmingtoolbox/prepare_fieldmap.py
```diff
@@ -3,11 +3,13 @@
...
 import math
 import numpy as np
+from skimage.filters import gaussian
 
 from shimmingtoolbox.unwrap.unwrap_phase import unwrap_phase
 
 
-def prepare_fieldmap(phase, echo_times, affine, unwrapper='prelude', mag=None, mask=None, threshold=None):
+def prepare_fieldmap(phase, echo_times, affine, unwrapper='prelude', mag=None, mask=None, threshold=None,
+                     gaussian_filter=None, sigma=None):
```
po09i:
```suggestion
                     gaussian_filter=False, sigma=None):
```
I would suggest going with a default value of `False` so that it's always a boolean

```suggestion
                     gaussian_filter=None, sigma=1):
```
I think having a default value for sigma would be useful, I have not played with the code to know which default is best.

shimmingtoolbox/prepare_fieldmap.py
```diff
@@ -81,4 +84,9 @@ def prepare_fieldmap(phase, echo_times, affine, unwrapper='prelude', mag=None, m
     fieldmap_rad = phasediff_unwrapped / echo_time_diff  # [rad / s]
     fieldmap_hz = fieldmap_rad / (2 * math.pi)  # [Hz]
 
+    # Gaussian blur the fieldmap
+    if gaussian_filter == True:
```
po09i:
```suggestion
    if gaussian_filter:
```
This can be simplified since the default value will be `False`

shimmingtoolbox/prepare_fieldmap.py
```diff
@@ -22,7 +24,8 @@ def prepare_fieldmap(phase, echo_times, affine, unwrapper='prelude', mag=None, m
         mag (numpy.ndarray): Array containing magnitude data relevant for ``phase`` input. Shape must match phase[echo].
         mask (numpy.ndarray): Mask for masking output fieldmap. Must match shape of phase[echo].
         threshold: Prelude parameter used for masking.
-
+        gaussian_filter: Option of using a Gaussian filter to smooth the fieldmaps (boolean)
```
po09i:
```suggestion
        gaussian_filter (bool): Option of using a Gaussian filter to smooth the fieldmaps (boolean)
```

shimmingtoolbox/prepare_fieldmap.py
```diff
@@ -22,7 +24,8 @@ def prepare_fieldmap(phase, echo_times, affine, unwrapper='prelude', mag=None, m
...
+        sigma: Standard deviation of gaussian filter.
```
po09i:
```suggestion
        sigma (float): Standard deviation of gaussian filter.
```

test/test_prepare_fieldmap.py
```diff
@@ -149,3 +149,13 @@ def test_prepare_fieldmap_3_echoes(self):
...
+
+    def test_prepare_fieldmap_gaussian_filter(self):
+        """ Test output of gaussian filter optional argument"""
+
+        fieldmap = prepare_fieldmap([self.phase], self.echo_times, self.affine, gaussian_filter=True)
+
+        assert fieldmap.shape == self.phase.shape
+        # If the behaviour of the called function is modified, this assertion below should capture it:
+        # assert np.all(np.isclose(fieldmap[30:35, 40, 0, 0],
+        #                          np.array([18.51355514, 13.84794053,  9.48013154,  5.11232207,  0.64524454])))
```
po09i: We should update this test a bit to make it more robust (Uncommenting these lines and updating the values being checked)

po09i (commented):
Some small suggestions to avoid errors.

> @po09i I'm ready to merge, however, I re-requested your review specifically because some checks have failed and I am not sure what it is I need to fix. Could you point me in the right direction to get them sorted?

One test failed: test/test_prepare_fieldmap.py:test_prepare_fieldmap_gaussian_filter
The error is related to the fact that the call to `prepare_fieldmap` does not include a sigma. Adding a default value as suggested in the review would avoid the error.

Lastly, I would suggest creating a test for the CLI as well since we have this new functionality in the CLI file as well.

evaalonsoortiz:
@po09i I believe this PR should be ready to merge now.

po09i (approved):
I added a CLI test as well, other than that LGTM! 👌🏼

PR #207, opened 2021-01-13
Fix coveralls 3.0.0 bug and github actions macos 11.0 bug

### LINKED ISSUE #206: Bug in coveralls release 3.0.0
The coveralls error is likely related to an upgrade in coveralls, see https://github.com/ivadomed/ivadomed/pull/628

_Originally posted by @jcohenadad in https://github.com/shimming-toolbox/shimming-toolbox/issues/184#issuecomment-758837739_

po09i:
## Description
As mentioned in [ivadomed](https://github.com/ivadomed/ivadomed/pull/628), coveralls introduced a bug with release 3.0.0. This fixes the error by:

```
run: |
  coveralls
to

run: |
  coveralls --service=github
```

While doing the checks, Github actions CI on MacOS 11.0 also fail. See this [issue](https://github.com/actions/virtual-environments/issues/2381). Since no timeframe is provided for the resolution. The build will be commented out until a fix is done.

## Linked issues
Fixes #206
https://github.com/actions/virtual-environments/issues/2381

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,7 +20,7 @@ jobs:
       matrix:
         os:
           - 'macos-10.15'
-          - 'macos-11.0'
+          # - 'macos-11.0'
           - 'ubuntu-18.04'
           - 'ubuntu-20.04'
           - 'ubuntu-16.04'
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -108,6 +108,6 @@ jobs:
     - name: Finished
       run: |
         pip3 install --upgrade coveralls
-        coveralls --service=github --finish
+        coveralls --finish
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -108,6 +108,6 @@ jobs:
     - name: Finished
       run: |
         pip3 install --upgrade coveralls
-        coveralls --finish
+        coveralls --service=github --finish
```
po09i: @jidicula Any reason why we would not want to put `--service=github` here? I noticed ivadomed only added it to the one above. Ivadomed's PR is linked in the description of this PR.

jidicula: I *think* it should be fine. Honestly coveralls is a bit of voodoo for me - I only remember having a lot of trouble getting this set up with the matrix build and just luckily coming across the right package to use, as the GitHub Actions use-case wasn't well-documented at the time.

jidicula: In any case, checks are passing now so at least we have that confirmation 🎉

jidicula (approved):
LGTM! Love seeing those green checkmarks ✅

jidicula:
Failing macos-11.0 checks are related to actions/virtual-environments#2381

PR #209, opened 2021-01-14
Implement Spatial Regularization in deltaB0 vs. Pressure Least Squares Fitting 

evaalonsoortiz:
This PR is a draft for the #eao/SpatialRegularization branch, in which the least squares fitting for deltaB0 vs. pressure is converted to matrix form and spatial regularization is included. 

In order to include a spatial regularizer, we need to implement the regression in matrix form (regression now requires joint estimation over all voxels). 

The code block below (adapted from [here](https://github.com/shimming-toolbox/shimming-toolbox/blob/a532152a873e4e7e305edcb119d864c3a9f93dd7/shimmingtoolbox/shim/realtime_zshim.py#L171)) does this by computing the design matrix **A** and the response vector **Bt**, so that we can find the vector of parameters x, such that x = min || **A***x - **Bt** ||^2. Note: the matrix **R** is the regularization term, which can be set to zero at this point.
```
    for g_axis in range(3):
        #linear operator: A
        A = sparse.hstack((I,p[0]*I))
            
        # solution vector: Bt
        Bt = np.array([-gradient[g_axis][:, :, :, 0]])
        Bt = np.reshape(Bt, (nVoxels, 1))

        for iT in range(1, nt):
            A = sparse.vstack((A,sparse.hstack((I,p[iT]*I))))

            tmpB = np.array([-gradient[g_axis][:, :, :, iT]])
            tmpB = np.reshape(tmpB, (nVoxels, 1))

            Bt = np.concatenate([Bt, tmpB])

        A = sparse.vstack((A, R))
        Bt = np.concatenate([Bt, np.zeros((2*nVoxels,1))])

        x, istop, itn, r1norm = lsqr(A,Bt)[:4]

        betas = np.reshape(x, (2, nx, ny, nz))

        static[g_axis][:, :, :] = betas[0, :, :, :]
        riro[g_axis][:, :, :] = betas[1, :, :, :] * pressure_rms
```

This implementation was successful, in that it reproduces the same results as those obtained by performing voxel-wise linear least-squares. 

The next step is to include regularization. We now want to find x, such that x = min ( || **A***x - **Bt** ||^2 + lamda * || **L**x ||^2 ), where lamda is a regularization parameter and the matrix **L** is a regularization matrix. **L** can be chosen to be a scaled finite difference approximation of the first derivative. 

The regularization matrix is computed as follows:

```
nVoxels = nx * ny * nz
nVoxPerSlice = nx * ny

I = sparse.identity(nVoxels,format='csc')

# lamda[0]: regularization parameter for the static field component
# lamda[1]: regularization parameter for the RIRO component
lamda = np.array([20, 30000]) # regularization parameters selected by trial and error

# first order finite difference operator: D
# 1st dimension
e = np.ones(nx)
diags = np.array([-1, 1])
Dx = spdiags([-e, e], diags, nx, nx).toarray()

while Dx.shape[1] < nVoxels :
    Dx = block_diag(Dx, Dx)
    Dx = Dx[0:nVoxels,0:nVoxels]     

# 2nd dimension
e = np.ones(nVoxPerSlice)
diags = np.array([-nx, nx])
Dy = spdiags([-e, e], diags, nVoxPerSlice, nVoxPerSlice).toarray() 

while Dy.shape[1] < nVoxels :
     Dy = block_diag(Dy, Dy)  
Dy = Dy[0:nVoxels,0:nVoxels]

# 3rd dimension
e = np.ones(nVoxels)
diags = np.array([-nVoxPerSlice, nVoxPerSlice])
Dz = spdiags([-e, e], diags, nVoxels, nVoxels).toarray() 

Dx = Dx/(2*delx) 
Dy = Dy/(2*dely) 
Dz = Dz/(2*delz) 
L = coo_matrix(Dx + Dy + Dz)

# R: regularization term
R = sparse.vstack([sparse.hstack([lamda[0]*L,coo_matrix((L.shape[0],L.shape[1]))]),sparse.hstack([coo_matrix((L.shape[0],L.shape[1])),lamda[1]*L])])
```

This code was adapted from the matlab shimming toolbox code. The spatial regularization functionality in the matlab code has been enabled in the [eao/SpatialRegularization](https://github.com/shimming-toolbox/shimming-toolbox-matlab/tree/eao/SpatialRegularization). 

The results obtained for the following dataset and inputs, using that branch are shown below:

**Dataset:** 
  acdc_121p/run_2
**Inputs:** 
  code call: realtime_zshim('phantom')
  field map magnitude data: 'GRE_FIELD_MAPPING_PMULOG_MEDBOTTLE_0032/'
  field map phase data: 'GRE_FIELD_MAPPING_PMULOG_MEDBOTTLE_0033/'
  MGRE magnitude data:  'GRE_REALTIME_ZSHIM_NOSHIM_0034/'
  ROI size: 25

Screen Shot 2021-01-14 at 11 15 54 AM

The results obtained with the python code:
Screen Shot 2021-01-14 at 2 41 27 PM

I am a bit stuck trying to debug this. Is it an array indexing issue or was an undetected error introduced when converting the regularization term from matlab to python? 
@rtopfer Do the regularization terms seem correct to you?  
@po09i Since you have more experience at this point with python, would you mind giving your input RE indexing, or if something else strikes you as wrong?


shimmingtoolbox/shim/realtime_zshim.py
```diff
@@ -113,21 +122,88 @@ def realtime_zshim(nii_fieldmap, nii_anat, pmu, json_fmap, nii_mask_anat=None, p
...
+
+    # 3rd dimension
+    e = np.ones(nVoxels)
+    diags = np.array([-nVoxPerSlice, nVoxPerSlice])
+    Dz = spdiags([-e, e], diags, nVoxels, nVoxels).toarray() 
+
+    Dx = Dx/(2*delx) 
+    Dy = Dy/(2*dely) 
+    Dz = Dz/(2*delz) 
+    L = coo_matrix(Dx + Dy + Dz)
```
po09i: I need to spend more time on this (will do monday) but I think the error is introduced around this line, more specifically `Dy` seems to create the pattern in the image. (I tried only running with `Dy` instead of `Dx+Dy+Dz` and the error is similar). I would revisit the calculation of those terms (or how they are used). I don't see any obvious mistake though.

evaalonsoortiz: oooh, good find! The error is gone without Dy. I'll narrow down on Dy then ...

evaalonsoortiz: Ok, I've figured out what's going on here. 

Previously I defined the following:
`nx, ny, nz, nt = nii_fieldmap.shape`, which returns **64 x 96 x 1 x 60**
numpy.shape presumably gives dimensions in the following order: **(metaframes, frames, rows, columns)**

In the case of `nii_fieldmap.shape`, that is not the order (I do not know why, and I haven't found any documentation giving information on this). I had _assumed_ the order was: **(rows, columns, frames, metaframes)**.

In the matlab code [(createdifferenceoperators.m)](https://github.com/shimming-toolbox/shimming-toolbox-matlab/blob/eao/SpatialRegularization/misc/createdifferenceoperators.m), it is stated that:
`Dx, Dy, & Dz that act as partial differential operators in row, column, and slice directions respectively.`

In matlab, the image matrix dimensions are: **96 x 64 x 1 x 1 x 60**, which corresponds to **rows x columns x slices x echoes x frames**. 

So, it looks like the matrix dimensions are flipped for python vs. matlab (?) 

What I've now done is changed the following:
`nCol, nRow, nSlice, nt = nii_fieldmap.shape`. This goes against the assumption I made above (that  `nii_fieldmap.shape` returns  **(rows, columns, frames, metaframes)**. 

The code now outputs static and riro correction data that make more sense (no more artifact), but don't exactly match the matlab-code output. However, the least-squares algorithm is not the same, so we can't expect the results to be (making it hard to validate whether or not the indexing is correct now).

po09i: > So, it looks like the matrix dimensions are flipped for python vs. matlab (?)

Either Numpy or Nibabel does make the orientation different from Matlab as you have seen:
- Matlab: 96 x 64 x 1 x 1 x 60
- Python: 64 x 96 x 1 x 60

I used the default for nibabel and numpy and they return nCol, nRow, nSlice, nt and I use different variables for the different echoes. The indexes for either the column or row are also indexed the other way (i.e. One of these is true:  `Row[0]` in Python == `Row[96]` in Matlab OR `Col[0]` in Python == `Col[64]` in Matlab but I never remember which one is true). I like to think of it as if the first 2 dimensions got rotated. (This is why all plots in python are rotated when compared to the plots in matlab)

PR #225, opened 2021-02-06
Cache SCT in GitHub Actions CI workflow

### LINKED ISSUE #216: GitHub Actions: Cache SCT in our CI tests workflow
### Checklist

- [x] I've given this issue a self-descriptive, meaningful title (e.g. *Document the use of config/dcm2bids.json*)

Currently, Spinal Cord Toolbox installation is the bottleneck in our CI tests that are run on each PR. Even on the fastest build matrix option, Ubuntu 20.04 on Python 3.8, [SCT installation step takes around 2m16s](https://github.com/shimming-toolbox/shimming-toolbox/pull/215/checks?check_run_id=1798130049#step:8:1).

This can easily be fixed by caching its install location, then only running its install step in the GH Actions workflow if the cache hit does not occur. However, before I start on that enhancement, I would like to understand how SCT is being used:

1. What version of SCT needs to be installed in CI? Does it need to be pinned, or should we continue always building the HEAD version of SCT on every PR in shimming-toolbox?
2. Is SCT a soft dependency? If so, should we create a build matrix to run tests with and without it?

### Additional Context / Screenshots:

N/A

jidicula:
## Checklist

## Description

Clones SCT on every run and stores its HEAD commit hash as the cache
key. If this key doesn't match the previously stored cache key, then
SCT is reinstalled and its install location is added to the cache.

The end goal of caching SCT in this workflow is to speed up CI runs -
SCT doesn't need to be reinstalled from scratch if it hasn't changed.
This speeds up development iteration on PRs with multiple cycles of
commit, request changes, commit.

Overall, these caching rules ensure that SCT updates are still
respected, but on average, a PR with a 1-day lifetime only ever requires
a single installation of SCT in GitHub Actions. As SCT stabilizes, we can consider
pinning a specific version of SCT and caching that.

The first time the cache is built (which is equivalent to our CI without caching SCT):

* [9m33s total](https://github.com/shimming-toolbox/shimming-toolbox/actions/runs/544250520) for the full build matrix
* [5m8s](https://github.com/shimming-toolbox/shimming-toolbox/runs/1847046938?check_suite_focus=true) for the fastest matrix option, Ubuntu 20.04 with Python 3.8:
image

Subsequent CI runs that use cached SCT: 
* [4m6s total](https://github.com/shimming-toolbox/shimming-toolbox/actions/runs/544267629) for the full build matrix
* [3m7s](https://github.com/shimming-toolbox/shimming-toolbox/runs/1847108707?check_suite_focus=true) for the fastest matrix option, Ubuntu 20.04 with Python 3.8:
image

Note that macOS runners have extremely variable runtime - there's no way around this, as those runners seem to be slammed by many GitHub users trying to run their daily iOS and macOS app builds using GitHub Actions.

## Linked issues

Resolves #216

See also https://github.com/actions/cache

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -58,11 +58,24 @@ jobs:
           unzip -o dcm2niix_lnx.zip
           sudo install dcm2nii* /usr/bin/
 
-      - name: SpinalCord Toolbox installation
+      - name: Spinal Cord Toolbox clone
         run: |
-          cd ../..
+          cd $HOME/work
           git clone --depth 1 https://github.com/neuropoly/spinalcordtoolbox
           cd spinalcordtoolbox
+          git rev-parse HEAD > $HOME/work/.sct_HEAD_hash
+
+      - name: Cache SCT
+        id: cache-sct
+        uses: actions/cache@v2
+        with:
+          path: ~/work/spinalcordtoolbox
+          key: ${{ matrix.os }}-${{ matrix.python-version }}-${{ hashFiles('~/work/.sct_HEAD_hash')}}
+
+      - name: Spinal Cord Toolbox installation
+        if: steps.cache-sct.outputs.cache-hit != 'true'
+        run: |
+          cd ../../spinalcordtoolbox
           ./install_sct -y
 
       - name: prelude macOS
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -74,7 +74,7 @@ jobs:
         with:
           path: ~/work/spinalcordtoolbox
           # Use commit hash file from previous step as key
-          key: ${{ matrix.os }}-${{ matrix.python-version }}-${{ hashFiles('~/work/.sct_HEAD_hash') }}
+          key: ${{ matrix.os }}-python-${{ matrix.python-version }}-${{ hashFiles('~/work/.sct_HEAD_hash') }}
 
       - name: Spinal Cord Toolbox installation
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -56,11 +60,27 @@ jobs:
...
+      - name: Cache SCT
+        id: cache-sct
+        uses: actions/cache@v2
+        with:
+          path: ~/work/spinalcordtoolbox
+          # Use commit hash file from previous step as key
+          key: ${{ matrix.os }}-${{ hashFiles('**/.sct_HEAD_hash') }}
+
+      - name: Spinal Cord Toolbox installation
+        if: steps.cache-sct.outputs.cache-hit != 'true'
```
jcohenadad, line 81: i'm not familiar with the syntax but i assume you tested it 😊

jidicula:
In a future PR, we can probably also permacache prelude and dcm2niix since those remain largely unchanged - what do you think @po09i @jcohenadad ?

edit: nvm, it only takes 7 seconds max (for dcm2niix installation) on the slowest runner, `macOS-10.15`. Not worth consideration at this time 😛

jcohenadad (approved):
i haven't tested it (no time) but i assume you did and it worked. In that case great contribution and thank you 😊

.github/workflows/tests.yml
jidicula: With lots of trial and error 😄  - there's no way around working with GH Actions (or any other CI) other than pushing a bunch of commits and seeing what sticks.

Issue #205, opened 2021-01-13
Consider using FSLEYES plugin as wrapper for the Shimming Toolbox


po09i:
### Feature Description:
Shimming Toolbox will eventually have a GUI. We could leverage FSLEYES to display the functionalities of the toolbox and make the CLIs available from the FSLEYES' GUI. Here is a screenshot from SCT.

![Screen Shot 2021-01-12 at 1 49 59 PM](https://user-images.githubusercontent.com/47249340/104411073-c1ec8e80-5537-11eb-85bc-a6f3617b1782.png)

Pros:
- Easier implementation than most options
- Already deals with niftis
- Also in python

Cons:
- Requires the end user to have fsleyes installed to use the GUI
- Hidden feature in fsleyes (maybe)

Some considerations if we go ahead with FSLeyes:
- have a design (put it in a platform where it's easy to give feedback/iterate-- draw.io, gslide, etc) before starting implementing
- could be either a script (like in [SCT](https://github.com/neuropoly/spinalcordtoolbox/blob/master/contrib/fsl_integration/sct_plugin.py)) or a plugin (like in [ADS](https://github.com/neuropoly/axondeepseg/blob/master/ads_plugin.py))
- installation: should we add fsleyes as a dependency in the requirements?
- license compat: https://git.fmrib.ox.ac.uk/fsl/fsleyes/fsleyes/-/blob/master/LICENSE

Feel free to add your pros and cons

charjl:
-ADS has a plugin for this (FSL-Eyes), SCT has a script. Why the difference?
-Is FSL-Eyes the correct tool for this ? There are installation problems. <2020-01-26: https://docs.google.com/document/d/1CKpGLPcQ133hNdVa-Hzk5AINkJiQ13-uBEiDogpE8sg/edit#heading=h.cuol6iqlzhv8 > 
-Different cli have tabs and images on the SCT one
-Implement it in such a way where you don't have to load the script into fsleyes, that it starts loaded from command line
-Run FSLEyes in SCT to get used to it
-Look at ADS implementation. Take what we want from each of the implementations

po09i:
The main CLIs we want to implement are:
- st_mask
- st_dicom_to_nifti
- st_prepare_fieldmap
- st_realtime_zshim

po09i:
### How should the main layout be? 
ADS and SCT's implementation is included for reference and to show the different tabs:

[...]

PR #232, opened 2021-02-25
Add GUI Integration into FSLeyes

ahill187:
## Checklist

## Description
We are integrating the CLI tools from `shimming-toolbox` into the `FSLeyes` package, as a plugin.

> ⚠️ When testing this PR, you should say "NO" when FSLeyes is asking to install this plugin permanently. Otherwise, you end us with this issue: https://github.com/shimming-toolbox/shimming-toolbox/pull/232#issuecomment-800697168

## Linked issues
https://github.com/shimming-toolbox/shimming-toolbox/issues/205


jcohenadad (commented):
I do prefer scripts, but open to discussion -- see https://github.com/shimming-toolbox/shimming-toolbox/issues/205#issuecomment-790006679

jcohenadad:
Few review comments (from high priority to low priority):
- Installing fsleyes and the plugin is cumbersome. Creating launchers instead of expecting the user to have the proper venv might be a better alternative: https://github.com/shimming-toolbox/shimming-toolbox/issues/237
- consider changing the plugin for a script: https://github.com/shimming-toolbox/shimming-toolbox/issues/205#issuecomment-790006679
- implement feature of selecting "input files" that are listed in the overlay section of fsleyes
- "mask" card: i got this error: `no such option: -i`
- "fieldmap": for unwrapper: create a combo box listing the options 
- "fieldmap": we should be able to only select the folder containing the fieldmaps-- the function will figure out each phase files and the number of echos
- "dcm2niix" card: if output folder does not exist, it will crash.
- add more verbose in the Terminal

shimmingtoolbox/utils.py
```diff
@@ -100,3 +101,18 @@ def st_progress_bar(*args, **kwargs):
         kwargs['disable'] = True
 
     return tqdm.tqdm(*args, **kwargs)
+
+
+def create_output_dir(path_output, output_folder_name="output"):
```
po09i:
```suggestion
def create_output_dir(path_output):
```

shimmingtoolbox/utils.py
```diff
@@ -100,3 +101,18 @@ def st_progress_bar(*args, **kwargs):
...
+    """Given a path, create the directory if it doesn't exist.
+
+    Args:
+        path_output (str): Full path to either a folder or a file.
+        output_folder_name (str): Name of sub-folder.
```
po09i:
```suggestion

```

po09i:
`FSLEYES_PLUGIN_PATH` shell variable can be used to automatically load the newest version of the plugin. Only need to activate it to show it up.

jcohenadad:
hum, why is it looking there? i'm using https://github.com/shimming-toolbox/shimming-toolbox/pull/232/commits/c5775938811eb1dfc308ad571649ab8bbdcd8d68

Screen Shot 2021-03-16 at 8 12 47 PM

po09i:
> hum, why is it looking there? i'm using [c577593](https://github.com/shimming-toolbox/shimming-toolbox/commit/c5775938811eb1dfc308ad571649ab8bbdcd8d68)

Since we can't import any shimming toolbox packages, we also can't import `__dir_shimmingtoolbox__`. Two things rely on this: the value for default paths and the logo. 

We defined `__dir_shimmingtoolbox__` like so:
`__dir_shimmingtoolbox__ = os.path.dirname(os.path.dirname(os.path.dirname(os.path.realpath(__file__))))` which works if the gui file is in the shimming toolbox repo. I'm guessing it tried to use the GUI file that was in .fsleyes so it could not find the logo since it was the wrong shimming toolbox path.

The only solution I see is having a shell variable for Shimming Toolbox so that we can find the logo. However, I don't think this is relevant to fix since we want to move to a fsleyes package in #242 anyways and the package will include the logo and change the default paths to the current directory.

PS: To test, I would remove the files in the .fsleyes directory, relaunch fsleyes, load the plugin but select "no" when prompted to permanently install since we are still in the dev stage. Hopefully it should work after that.

po09i:
It looks good to me but I was a big contributor so I'll wait for others to have a look.

jcohenadad:
> > hum, why is it looking there? i'm using [c577593](https://github.com/shimming-toolbox/shimming-toolbox/commit/c5775938811eb1dfc308ad571649ab8bbdcd8d68)
> 
> Since we can't import any shimming toolbox packages, we also can't import `__dir_shimmingtoolbox__`. Two things rely on this: the value for default paths and the logo.
> 
> We defined `__dir_shimmingtoolbox__` like so:
> `__dir_shimmingtoolbox__ = os.path.dirname(os.path.dirname(os.path.dirname(os.path.realpath(__file__))))` which works if the gui file is in the shimming toolbox repo. I'm guessing it tried to use the GUI file that was in .fsleyes so it could not find the logo since it was the wrong shimming toolbox path.
> 
> The only solution I see is having a shell variable for Shimming Toolbox so that we can find the logo. However, I don't think this is relevant to fix since we want to move to a fsleyes package in #242 anyways and the package will include the logo and change the default paths to the current directory.
> 
> PS: To test, I would remove the files in the .fsleyes directory, relaunch fsleyes, load the plugin but select "no" when prompted to permanently install since we are still in the dev stage. Hopefully it should work after that.

what i did:
- removed the file `~/.fsleyes/plugins/st_plugin.py`
- activated ST venv
- opened fsleyes
- re-loaded this plugin
- i get the same error

gaspardcereza (approved):
We tested everything from scratch and it worked (MacOs Catalina & Big Sur). Installation process will be updated once the new plugin repo will be working.

po09i (changes requested):
Before merging in master, we should merge #242 in this branch and port to the new repo the gui files, then remove those files from this repo.

PR #247, opened 2021-03-22
Update dcm2bids to handle all TFL B1 maps

### LINKED ISSUE #246: Change in TFL B1 mapping "SequenceName" tag
### Environment details

Mac OS Catalina. Followed the installation process. 

### Bug Description

When converted to nifti, the `SequenceName` tag of the TFL B1 mapping dicoms isn't "tfl12d1_16" as it used to be but “*tfl12d1_16". The star at the beginning wasn't there during the older conversions. 

The star might be due to a wrong version of `dcm2niix` or simply to ho the scanner writes the tags directly. I am currently investigating that problem.

Updating the `dcm2bids.json` file so that it handles both cases should be enough to correct this bug.

gaspardcereza:
Resolves #246

### DIFF
--- a/config/dcm2bids.json
+++ b/config/dcm2bids.json
@@ -224,7 +224,7 @@
             "dataType": "rfmap",
             "modalityLabel": "TB1map",
             "criteria": {
-                "SequenceName": "tfl2d1_16"
+                "SequenceName": "*tfl2d1_16"
             }
         }
     ]


gaspardcereza:
@po09i reviewing this should not be very time consuming 😉

po09i (approved):
Looks good!

PR #248, opened 2021-03-24
Support fsleyes as a GUI

### LINKED ISSUE #205: Consider using FSLEYES plugin as wrapper for the Shimming Toolbox

### LINKED ISSUE #241: Consider converting fsleyes plugin to a package
### Motivation

Currently, the `fsleyes` plugin implementation is a script (WIP), but I think converting it to a package might be the way to go. The `FSLeyes` documentation describes using a plugin package here: https://users.fmrib.ox.ac.uk/~paulmc/fsleyes/userdoc/latest/plugins.html

### Pros

1. If we package it, we can have multiple files/folders. This will allow us to not have one giant file, and also to import the logos etc. If we don't package it, we would have to set an environment variable to find where the logos/images are installed via `shimming-toolbox`. This seems like it could get confusing.
2. If we have a package called `fsleyes-plugin-your-package-name` here will automatically be loaded, and will update each time you update the plugin and restart `fsleyes`. This means we don't have to do `File -> Load plugin` each time it updates.
3. Since we want the plugin to be independent of `shimming-toolbox`, it doesn't really make sense to have it as part of the `shimming-toolbox` package.

### Cons

- The package needs to be installed within the venv that `fsleyes` is installed, this has implications:
  - All dependencies from `fsleyes`, `fsleyes-plugin-shimming-toolbox` and `shimming-toolbox` are merged
    - FIX: `shimming-toolbox` dependencies can be separated with launchers
  - We either have to install `fsleyes-plugin-shimming-toolbox` in `fslpython` venv (seems wrong) or create a new environment and download `fsleyes` within it and `fsleyes-plugin-shimming-toolbox`. Not too bad
- We need a way to find where the repo is installed for: the default `shimming-toolbox` config file for dcm2bids.
  - Possibly a SHELL variable to find the default `shimming-toolbox` config file for dcm2bids (The variable could be SHIMTOOL_DIR)

### Examples

https://bitbucket.org/gaunab/fsleyes-plugin-roitool/src/master/

### Draft Implementation

If you want to test out using/installing the package yourself, I have a draft pull request here:

https://github.com/shimming-toolbox/shimming-toolbox/pull/242

You can install by doing:

```
pip install -e fsleyes-plugin-shimming-toolbox
```

po09i:
## Description
Implement the relevant changes to support fsleyes as a GUI. This PR is used in parallel with [fsleyes-plugin-shimming-toolbox PR #1](https://github.com/shimming-toolbox/fsleyes-plugin-shimming-toolbox/pull/1)  and [fsleyes-plugin-shimming-toolbox PR #2](https://github.com/shimming-toolbox/fsleyes-plugin-shimming-toolbox/pull/2)  to implement the GUI.

## Linked issues
Fixes #205
Fixes #241

## Linked PR
This PR reopens #232 due to a move of the main GUI files to https://github.com/shimming-toolbox/fsleyes-plugin-shimming-toolbox


setup.py
```diff
@@ -44,6 +44,9 @@
         "pytest~=4.6.3",
         "pytest-cov~=2.5.1",
         "sklearn~=0.0",
+        "pillow~=8.0",
+        "dataclasses",
+        "raven",
```
po09i, line 51: @ahill187 Do you know if this is still required?

ahill187 (approved):

PR #250, opened 2021-03-26
Add ability to average across files

### LINKED ISSUE #245: Add ability to average different echoes for better SNR for masking
When running `st_mask sct` on a file with low SNR. SCT has a hard time with the segmentation, we are already averaging across time points if the input is 4d. However, we could further improve by averaging across echoes. The reason we are not currently doing so is that the BIDS convention separates the echoes in different files. This means that it is not trivial to average across echoes.

I see  ways of going with this:
- Create `st_image concat file1.nii.gz file2.nii.gz --axis 4` which would concatenate the files into one file
- Modify `st_mask sct` to search within the folder of the input file for other echoes
- Modify `st_mask sct --input` to allow for multiple input files (the different echoes)

I'm leaning towards the first option as it is more robust and allows to have the same benefit for multiple masking methods (BET, SCT, threshold)

po09i:
## Description
This PR introduces 2 new CLIs:
- `st_image` allows to make modifications on images
   -  Adds the `concat` argument
-  `st_maths` allows to perform math operations on the image
   - Adds the `mean` argument

Only those 2 arguments will be available right now but the CLI framework will be set up to include more math/image operations/manipulations.

st_image concat -h
``` 
Usage: st_image concat [OPTIONS] INPUT...

  Concatenate NIfTIs along the specified dimension

  input: Input paths for the files to concatenate. Separate the files by a
  space.

Options:
  --output PATH       Output filename, supported extensions: .nii, .nii.gz
  --axis [0|1|2|3|4]  Dimension to concatenate  [required]
  --pixdim FLOAT      Pixel resolution to join to image header
  -h, --help          Show this message and exit.
```

st_maths mean -h
```
Usage: st_maths mean [OPTIONS]

  Average data across dimension.

Options:
  --input PATH        Input filename, supported extensions: .nii, .nii.gz
                      [required]

  --output PATH       Output filename, supported extensions: .nii, .nii.gz
  --axis [0|1|2|3|4]  Axis of the array to calculate the average  [required]
  -h, --help          Show this message and exit.
```

## To test the PR 
Make sure `testing_data` is in the root of shimming-toolbox, if not, fix the path to `testing_data`, or use your own files
```
cd shimming-toolbox
pip install -e .
mkdir temp_folder
st_image concat testing_data/realtime_zshimming_data/nifti/sub-example/anat/sub-example_unshimmed_e1.nii.gz testing_data/realtime_zshimming_data/nifti/sub-example/anat/sub-example_unshimmed_e2.nii.gz testing_data/realtime_zshimming_data/nifti/sub-example/anat/sub-example_unshimmed_e3.nii.gz --axis 4 --output temp_folder/concat.nii.gz
st_maths mean --input temp_folder/concat.nii.gz --axis 4 --output temp_folder/mean.nii.gz
```

You should then have a file `temp_folder/mean.nii.gz` which contains the average of the input files

## Linked issues
Fixes #245


evaalonsoortiz:
I just tested this out using acdc_129 and I still end up with an ROI that is has missing slices. Are you reproducing the same steps that were taken in the matlab code?

po09i:
What are the steps taken in the Matlab code? The steps taken [here](https://github.com/shimming-toolbox/shimming-toolbox/pull/250#issue-601880118) are:
1. Concatenating the echoes in a single file 
2. Averaging over that dimension

jcohenadad (commented):
I tested both functions and they work nicely 👌

shimmingtoolbox/cli/image.py
```diff
@@ -0,0 +1,43 @@
...
+# -*- coding: utf-8 -*
+
+import click
+import os
+import nibabel as nib
+
+from shimmingtoolbox.image import concat_data
+
+CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
+AXES = ['0', '1', '2', '3', '4']
```
jcohenadad, line 11: is "4" really necessary?

shimmingtoolbox/cli/maths.py
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*
+
+import click
+import os
+import nibabel as nib
+import numpy as np
+
+CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
+AXES = ['0', '1', '2', '3', '4']
```
jcohenadad, line 13: same comment about '4'

shimmingtoolbox/cli/image.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+@click.group(context_settings=CONTEXT_SETTINGS,
```
jcohenadad, line 14: I know it's a "click" thing, but the usage shows up like this:
~~~
Usage: st_image concat [OPTIONS] INPUT...
~~~

whereas in fact, INPUT could go _before_ the options (which i find more intuitive). I would prefer having the INPUT listed before, but it's not a big deal...

shimmingtoolbox/cli/maths.py
jcohenadad, line 13: in 99% of the case, axis will be 3 (time). I suggest to put it in default. Same comment for the math CLI

shimmingtoolbox/cli/maths.py
```diff
@@ -0,0 +1,44 @@
...
+@click.group(context_settings=CONTEXT_SETTINGS,
+             help="Perform mathematical operations on images.")
+def maths_cli():
+    pass
+
+
+@maths_cli.command(context_settings=CONTEXT_SETTINGS)
+@click.option('--input', 'fname_input', type=click.Path(exists=True), required=True,
+              help="Input filename, supported extensions: .nii, .nii.gz")
+@click.option('--output', 'fname_output', type=click.Path(), default=os.path.join(os.curdir, 'mean.nii.gz'),
```
jcohenadad, line 22:
```suggestion
@click.option('--output', 'fname_output', type=click.Path(), default=os.path.join(os.curdir, '_mean.nii.gz'),
```

jcohenadad: a more intuitive default output would be to add "_mean" suffix to the input file

shimmingtoolbox/cli/image.py
```diff
@@ -0,0 +1,43 @@
...
+             help="Perform manipulations on images")
+def image_cli():
+    pass
+
+
+@image_cli.command(context_settings=CONTEXT_SETTINGS)
+@click.argument('input', nargs=-1, type=click.Path(exists=True), required=True)
+@click.option('--output', 'fname_output', type=click.Path(), default=os.path.join(os.curdir, 'concat.nii.gz'),
```
jcohenadad, line 22: if users don't read the usage (users never read the doc) and fail to provide an output name, the file concat is created and they don't know about this. So either force output flag to be used (not great if ST will mostly be used with the GUI) or indicate in the terminal what will be the output.

alternatively, you could create a file name where you concatenate the file names of the input eg:
~~~
st_image concat fmri1.nii.gz fmri2.nii
--> Created: fmri1_fmri2.nii
~~~

shimmingtoolbox/cli/image.py
po09i: My thought is if the user has data that is (x,y,z,volumes) and wants to concatenate echoes, then they would need a 5th dimension.

shimmingtoolbox/cli/image.py
po09i: I had noticed the same thing, however, the layout does not seem to be customizable yet according to the [doc](https://click.palletsprojects.com/en/7.x/documentation/)

> Click makes it very easy to document your command line tools. First of all, it automatically generates help pages for you. While these are currently not customizable in terms of their layout, all of the text can be changed.

shimmingtoolbox/cli/image.py
jcohenadad: ah yes, i forgot that in ST we put the echoes in the 5th dim. OK, i guess it makes sense then

evaalonsoortiz:
> I just tested this out using acdc_129 and I still end up with an ROI that is has missing slices. Are you reproducing the same steps that were taken in the matlab code?

Ah, sorry I should have been more specific. st_image and st_maths work as expected. 

The problem I was having is that when running st_mask sct, the resulting mask is missing in some slices (as described in issue #245). I noticed however, that st_mask sct is running sct_deepseg_sc, whereas in the matlab repo we are creating a cylindrical mask using the centerline:
`system( ['sct_create_mask -i ' Params.tmpSaveDir 't2s.nii.gz -p centerline,' ...`
`    Params.tmpSaveDir 't2s_centerline.nii.gz -size ' ...`
`    num2str(Params.cylinderSize) 'mm -f cylinder -o ' ...`
[https://github.com/shimming-toolbox/shimming-toolbox-matlab/blob/7f0ba2d480a697a16a0d4b20154c61171f7e13ca/Img/MaRdI.m#L2518](url)

So, to summarize this PR does improve masking when using sct_create_mask with a cylinder (I tested this as well), so its goal is achieved. 

Going forward we may want to add the option of choosing between sct_deepseg_sc and the cylindrical mask (I think this has already been discussed in [#164](https://github.com/shimming-toolbox/shimming-toolbox/pull/164)). For my particular protocol, the spinal cord in the MGRE images can not be reliably segmented using sct_deepseg_sc (see https://github.com/evaalonsoortiz/dynamic_realtime_shimming/issues/5 for further discussion).

ahill187 (changes requested):
> When testing this locally, I used the testing examples as follows:
> 
> ```
> st_image concat /Users/work/Coding/NeuroPoly/shimming-toolbox/testing_data/realtime_zshimming_data/nifti/sub-example/anat/sub-example_unshimmed_e1.nii.gz /Users/work/Coding/NeuroPoly/shimming-toolbox/testing_data/realtime_zshimming_data/nifti/sub-example/anat/sub-example_unshimmed_e2.nii.gz --output my_test_folder/
> ```
> 
> When using the `--output` flag, it creates a file called `.nii` in `my_test_folder`. If I remove the `--output` flag, it creates a file called `concat.nii.gz`, which is expected. I think the file name has a bug when using the `--output` flag.

**Update:** I re-read the instructions, it has to be an output file name. My bad! Seems to be working 😄

ahill187 (approved):

PR #253, opened 2021-04-23
Refactor and add features to Optimizers

### LINKED ISSUE #152: Bounds option in LSQ_Optimizer throws error if the value is None
Project version:

1058ca2

### Bug Description
Our optimizer has a bounds parameter which defaults to None if it is not provided. The `scipy.optimize.least_squares()` function expects a tuple as an input and defaults to `bounds=(-np.inf, np.inf)`. This means it crashes whenever we input a default value or when we set `bounds=None`.

The error was first seen while failing a travis build : https://travis-ci.com/github/shimming-toolbox/shimming-toolbox/builds/191295723 

### Solution

```
if bounds == None:
    bounds=(-np.inf, np.inf)
```

### LINKED ISSUE #278: multiprocessing sometimes hanging
### Environment details

- I followed the [installation instructions](https://shimming-toolbox.org/en/latest/2_getting_started/1_installation.html)

Project version:

Branch: `ad/optimizer`

Operating System and version:

```
(st_venv) alex@Alexandres-MacBook-Pro shimming-toolbox % st_dump_env_info 
ENVIRONMENT INFO:
posix x86_64
Darwin 20.4.0
Darwin Kernel Version 20.4.0: Fri Mar  5 01:14:14 PST 2021; root:xnu-7195.101.1~3/RELEASE_X86_64
CPython 3.7.9

CPU cores: Available: 8, Used by ITK functions: 0
RAM: Total: 16384MB, Used: 8490MB, Available: 5352MB

PACKAGE INFO:
0.1.0
```

### Bug Description
When using multiprocessing (specifically the multiprocessing in the coordinates.py file:resample_from_to()) it sometimes hangs and a keyboard interrupt (ctrl + c) needs to happen by the user to stop the process. The behaviour did not occur when launching test_sequential_zslice from the IDE but occured when launching pytest from the command line on my computer (BigSur). The bug can also be seen in these 2 commits on MacOs Catalina python 3.7: https://github.com/shimming-toolbox/shimming-toolbox/pull/253/commits/35ab6178cdf50e51313602c2ab22b462ea30a2c2 and https://github.com/shimming-toolbox/shimming-toolbox/pull/253/commits/a45ff8a3d9792b4676cc281666adf0d95d726417.

Screen Shot 2021-06-09 at 12 49 08

https://stackoverflow.com/questions/55812213/multiprocessing-pool-hangs-on-running-program
Note that there is a ref to the following link in the comments which explains what I think is the bug but was fixed in python 3.8
https://bugs.python.org/issue33725

If this is really the case, we would need to drop support fo 3.7... 
For now multiprocessing is not necessary so I just commented the code out but this is something to think about later if we want to add multiprocessing.

Other threads I looked at:
https://github.com/gammapy/gammapy/issues/2453

po09i:
## Description
This PR refactors and improves the Optimizer classes previously developed to allow for use with custom coils.

More precisely, the PR adds:
- A Coil class to store coil profiles, affine and constraints
- A feature that resamples coil profiles to the target unshimmed image
- The ability to select the optimization algorithm while dynamic shimming
- The ability to respect the constraints for the least-squares method (only bounds were previously respected, now total coef as well)

## Test this PR
To test the PR, you can use the `test_sequential_zslice` test that uses the dynamic shimming optimizer.

## Test the speed of the optimizer/resampling
https://github.com/shimming-toolbox/shimming-toolbox/pull/253#issuecomment-852448288

## Linked issues
Fixes #152 
Fixes #278


shimmingtoolbox/coils/coil.py
```diff
@@ -0,0 +1,73 @@
...
+                                         is used to specify no bound.
+                Example:
+                    constraints = {
+                        "coef_sum_max": 40,
+                            # 8 channel coil
+                        "coef_channel_minmax": [(-2, 2), (-2, 2), (-2, 2), (-2, 2), (-3, 3), (-3, 3), (-3, 3), (-3, 3)]
+                    }
+        """
+
+        self.x = self.y = self.z = self.n = -1
```
jcohenadad:
```suggestion
        self.x = self.y = self.z = self.n = np.nan
```

shimmingtoolbox/coils/coil.py
```diff
@@ -0,0 +1,73 @@
...
+
+class Coil(object):
+    """
+    Coil profile object that stores coil profiles and there constraints
+
+    Attributes:
+        x (int): Amount of pixels in the X direction
+        y (int): Amount of pixels in the Y direction
+        z (int): Amount of pixels in the Z direction
+        n (int): Amount of channels in the coil profile
```
jcohenadad:
```suggestion
        dim (tuple int): Dimension along specific axis. dim:0,1,2 are spatial axes, while dim:3 corresponds to the coil channel.
```

shimmingtoolbox/coils/coil.py
```diff
@@ -0,0 +1,73 @@
...
+        profiles (numpy.ndarray): (X, Y, Z, N) 4d array of N 3d coil profiles
```
jcohenadad:
```suggestion
        profile (numpy.ndarray): (X, Y, Z, N) 4d array of N 3d coil profiles
```

shimmingtoolbox/optimizer/basic_optimizer.py
```diff
@@ -13,110 +18,84 @@ class Optimizer(object):
     For basic optimizer, uses unbounded pseudo-inverse.
 
     Attributes:
-        X (int): Amount of pixels in the X direction
-        Y (int): Amount of pixels in the Y direction
-        Z (int): Amount of pixels in the Z direction
-        N (int): Amount of channels in the coil profile
-        coils (numpy.ndarray): (X, Y, Z, N) 4d array of N 3d coil profiles
+        coil (Coil): Coil object containing the coil profiles and related constraints
```
jcohenadad: list of Coil

shimmingtoolbox/coils/coil.py
```diff
@@ -0,0 +1,73 @@
...
+        coef_sum_max (float): Contains the maximum value for the sum of the coefficients
```
jcohenadad, line 24: we might want to reconsider the naming of the variable, given that the ACDC coil constraint unit is "amp"

shimmingtoolbox/coils/coil.py
```diff
@@ -0,0 +1,77 @@
...
+        coef_channel_minmax (list): Contains the maximum coefficient for each channel
+    """
+
+    def __init__(self, profile, affine, constraints):
+        """
+
+        Args:
+            profile (np.ndarray): Coil profile (dim1, dim2, dim3, channels) 4d array of N 3d coil profiles
+            affine (np.ndarray): 4x4 array containing the affine transformation for the coil profiles
```
jcohenadad: maybe check or specify in the docstring if you are referring to the qform or sform-- we might also want to add a check to make sure that both sform (used by FSL/SPM) and qform (used by ITK/SCT) are the **same**. They should, because no processing (transformation) should be run on the coil profile.

shimmingtoolbox/coils/coil.py
po09i: I agree that `coef` might not be the best name however the `Coil` class does not only support "amps" as a unit. We can use it with spherical harmonics which don't have "amps" as units. The `coef` really only tells us how many times we can multiply the `profile`, regardless of the unit. 

So from a design perspective I feel like we are restricting the use case by saying amps but from a user's perspective it gives a lot more intuition for what I think will be the scenario used by people most of the time.

shimmingtoolbox/coils/coil.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+    Coil profile object that stores coil profiles and there constraints
+
+    Attributes:
+        dim (Tuple[int]): Dimension along specific axis. dim: 0,1,2 are spatial axes, while dim: 3 corresponds to the
+                          coil channel.
+        profile (np.ndarray): (dim1, dim2, dim3, channels) 4d array of N 3d coil profiles
+        affine (np.ndarray): 4x4 array containing the qform affine transformation for the coil profiles
+        coef_sum_max (float): Contains the maximum value for the sum of the coefficients
+        coef_channel_minmax (list): Contains the maximum coefficient for each channel
```
gaspardcereza, line 25: Isn't it weird to call it "minmax" if it contains the maximum coefficient ?

gaspardcereza:
Comment from Jason: 
"One minor detail:  For joint optimizations with the scanner shims, we might want to define separate “total current” limits for the scanner shims and the local multi coil shims"

po09i:
> "One minor detail: For joint optimizations with the scanner shims, we might want to define separate “total current” limits for the scanner shims and the local multi coil shims"

Yes, this is implemented! We can use multiple coils for the optimization with each coil having their respective max total current.

po09i:
### Speed test
For speed concerns, I implemented a test that I have since commented out since it takes around 20s on my machine to run to check the speed of the whole optimizing pipeline for a typical heavy scenario: 
* unshimmed maps: (100x100x60) created using NumericalModel. 3 different maps are generated and are repeated in the different slices if the number of slices is greater than 3. 
  * The first is a linear b0 field in one direction. 
  * This field is then used to create a second field, `(np.rot90(unshimmed[:, :, 0]) + unshimmed[:, :, 0]) / 2`
  * The third is the first one squared `unshimmed[:, :, 0] ** 2`
  * I used an affine with scaling factors, changes in orientation and translation (no shearing).
* Coil profile (150x150x70x32) created using siemens basis which returns 1st order and 2nd order spherical harmonics coil profiles (8 channels). This used siemens basis concatenated 4 times (32 channels).
  * I used an identity affine with scaling factors.

### To reproduce: 
* Uncomment the test in `test_sequential_zslice` called `test_speed_huge_matrix`
* Comment out everything after the line `# Calculate theoretical shimmed map` since we don't need to calculate this
* Above the line `@pytest.mark.parametrize(`, change nz for 60 to change the number of slices to 60 instead of 3 (line 78 at this moment)
* Set up a tic() toc() in `basic_optimizer`'s init

### My setup

```
(st_venv) alex@Alexandres-MacBook-Pro shimming-toolbox % st_dump_env_info 
ENVIRONMENT INFO:
posix x86_64
Darwin 20.4.0
Darwin Kernel Version 20.4.0: Fri Mar  5 01:14:14 PST 2021; root:xnu-7195.101.1~3/RELEASE_X86_64
CPython 3.7.9

CPU cores: Available: 8, Used by ITK functions: 0
RAM: Total: 16384MB, Used: 8146MB, Available: 6284MB

PACKAGE INFO:
0.1.0
```
To be clear the unused RAM is while not processing

### Results
hash when test occured: https://github.com/shimming-toolbox/shimming-toolbox/pull/253/commits/cad331b27f52aa7bc3fbaad5bacee23f293706d9
#### Optimization aka `test_speed_huge_matrix()`
I ran the whole test 4-5 times and averaged around 20s on my machine, this includes the time to run `test_speed_huge_matrix()` which includes optimizing each slice independently and resampling the coil profiles.
* (150x150x70x8) : ~6s
* 2 * (150x150x70x8) : ~12.5s
* (150x150x70x32) : ~20s

#### Resampling
For the actual **resampling** of the coil profiles, they took around **3s** for 32 channels:
* (150x150x70x8) --> (100x100x60x8) : ~0.9s
* 2 * (150x150x70x8) --> (100x100x60x16) : ~1.8s
* (150x150x70x32) --> (100x100x60x32) : **~3.2s**

Note: there is some multi processing done to make it faster. See `shimmingtoolbox/coils/coordinates.py` file where the resampling is implemented.

shimmingtoolbox/coils/coil.py
po09i: Yes.. I wanted to find a way to include the "max" negative current/coefficient

shimmingtoolbox/coils/coil.py
```diff
@@ -0,0 +1,84 @@
...
+    def profile(self, profile):
+        if profile.ndim != 4:
+            raise ValueError(f"Coil profile has {profile.ndim} dimensions, expected 4 (dim1, dim2, dim3, channel)")
+        self.dim = profile.shape
+        self._profile = profile
+
+    def load_constraints(self, constraints):
+        """Loads the constraints named in required_constraints as attribute to this class"""
+
+        required_constraints = [
```
jcohenadad: you could put it as a global at the top and call this variable in the docstrings, so the doc and code are not duplicated

shimmingtoolbox/coils/coil.py
```diff
@@ -0,0 +1,84 @@
...
+                }
+        """
+
+        self.dim = (np.nan,) * 4
+        self.profile = profile
+
+        if affine.shape != (4, 4):
+            raise ValueError("Shape of affine matrix should be 4x4")
+        self.affine = affine
+
```
jcohenadad, line 60: suggestion: make "required_constraints" an attribute so it can be exposed when inspecting the class

test/test_sequential_zslice.py
```diff
@@ -7,54 +7,151 @@
...
-    z_slices = np.array(range(nz))
-    currents = sequential_zslice(unshimmed, coils, full_mask, z_slices, bounds=bounds)
-
-    # Calculate theoretical shimmed map
-    shimmed = unshimmed + np.sum(currents * coils, axis=3, keepdims=False)
-
-    for i_slice in z_slices:
-        assert(np.sum(np.abs(full_mask[:, :, i_slice] * shimmed[:, :, i_slice])) <
-               np.sum(np.abs(full_mask[:, :, i_slice] * unshimmed[:, :, i_slice])))
+class TestSequentialZSlice(object):
```
jcohenadad, line 97: you might want to add a pytest session decorator, otherwise this `setup(self)` is run each time you run a new test in this file. Example [here](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/master/unit_testing/test_image.py#L18)

shimmingtoolbox/optimizer/sequential.py
```diff
@@ -2,32 +2,86 @@
...
+    Args:
+        method (str): Supported optimizer: 'least_squares', 'pseudo_inverse'
+        unshimmed (numpy.ndarray): 3D B0 map
+        affine (np.ndarray): 4x4 array containing the affine transformation for the unshimmed array
+        coils (ListCoil): List of Coils containing the coil profiles
+
+    Returns:
+        Optimizer: Initialized Optimizer object
     """
+    supported_optimizer = {
```
jcohenadad: you could put it as a global at the top and call this variable in the docstrings, so the doc and code are not duplicated. Alternatively we could create a function (or method?) where we hard-code the authorized values. @ahill187 do you have any tips for good practice here?

jcohenadad (approved):

shimmingtoolbox/optimizer/sequential.py
po09i: Having variables in the docstring is not as straightforward as I thought and not really recommended (according to stack overflow): https://stackoverflow.com/questions/10307696/how-to-put-a-variable-into-python-docstring

There were a couple of solutions like a decorator but they all don't work really well for various reasons (sphinx, we have an example with {}, the doc is not clear when just looking at the code).. It seems more hassle than what we gain.

I will still put it on top since it is a constant and it is closer to the doc.

test/test_sequential_zslice.py
po09i: Pytest classes are set up to run the `setup` method every time to isolate each tests. I used the parametrize decorator to achieve this and pass arguments to the class methods.

PR #270, opened 2021-05-27
Output in-plane (x and y) shim coefficients

evaalonsoortiz:
In this PR realtime_zshim.py is modified so that in-plane (x and y) shim coefficients are output alongside with z-shim coefficients. Filenames are also changed form *zshim* to *shim*.


jcohenadad:
Looks like this branch includes modifications from previous PRs

po09i (approved):
I updated some minor syntax warnings: 120 character limit, minor typos, etc. I also updated the description in the example scripts, demo script and the website. I renamed and updated the demo script since it was failing due to the renaming of realtime_zshimming to realtime_shimming.

Otherwise it looks good to me!

PR #283, opened 2021-07-06
Handling of different slice orientation when reading b1 maps

### LINKED ISSUE #281: Problem with reordering the B1 maps after nifti conversion
### Environment details

Project version: Master branch (2021/06/10)

Operating System and version: Catalina (Mac OS)

Python version: 3.7

### Bug Description
I tried to scaled B1 maps from our last acquisition at the MNI (acdc_spine_7t_007) and realized they are not reshuffled correctly by `load_nifti` (slices are still shuffled).  

### Reproduction Steps
Need access to dicoms from our last acquisition (acdc_7t_007). After converting dicoms into nifti, run `shimming-toolbox.load_nifti` on a  `b1map_standard` nifti. 

Probably because some B1 mapping acquisitions are interleaved and others are not not --> Currently checking this

gaspardcereza:
## Description
I found out that dcm2niix's inconsistent shuffling of B1 maps (mentioned in #281 )was due to the orientation of the slices:
* axial imaging --> Coils are in correct order 
* sagittal imaging --> Coils are inverted

I added a `if` statement based on the nifti tag `ImageOrientationPatientDICOM` in order to apply the correct reordering. However, I don't have any coronal acquisitions yet so this case will have to be handled later. 

## Next steps
- [ ] Implement handling of coronal slices case
- [ ] Find a way to assert masking consistency within the slices
- [ ] Add nifti files in [testing_data](https://github.com/shimming-toolbox/data-testing) to test sagittal and coronal cases (we need to perform a new acquisition for coronal images)
- [ ] Test all scenarios

## Linked issues
Resolves #281

### DIFF
--- a/shimmingtoolbox/load_nifti.py
+++ b/shimmingtoolbox/load_nifti.py
@@ -216,37 +216,57 @@ def scale_tfl_b1(image, json_data):
         warnings.warn("Missing json tag: 'SliceTiming', slices number cannot be checked.")
         n_slices = image.shape[2]
 
-    # Calculate B1 efficiency (1ms, pi-pulse) and scale by the ratio of the measured FA to the saturation FA.
-    # Get the Transmission amplifier reference amplitude
-    amplifier_voltage = json_data['TxRefAmp']  # [V]
-    socket_voltage = amplifier_voltage * 10 ** -0.095  # -1.9dB voltage loss from amplifier to coil socket
     # Magnitude values are stored in the first half of the 4th dimension
-    b1_mag = image[:, :, :, :image.shape[3] // 2] / 10  # Siemens magnitude values are stored in degrees x10
+    b1_mag = image[:, :, :, :image.shape[3] // 2]
 
     if b1_mag.min() < 0:
         raise ValueError("Unexpected negative magnitude values")
 
+    # Phase values are stored in the second half of the 4th dimension. Siemens phase range: [248 - 3848]
+    b1_phase = image[:, :, :, image.shape[3] // 2:]
+
+    # Reorder data shuffled by dm2niix into shape (x, y , n_slices*n_coils)
+    b1_mag_vector = np.zeros((image.shape[0], image.shape[1], n_slices * n_coils))
+    b1_phase_vector = np.zeros((image.shape[0], image.shape[1], n_slices * n_coils))
+    if 'ImageOrientationPatientDICOM' in json_data:
+        # If axial slices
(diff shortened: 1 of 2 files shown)


gaspardcereza:
I have a problem with asserting the masking consistency within a slice, because of a very few null values encountered from time to time as can be seen below: 

image
 
What I want to do is basically to make sure that within a slice, the overall masking is the same for all coils (thus I know that the different slices are no more mixed up after reordering the B1 maps). Obviously, there are still a few values that might be masked oud for certain coils and not for others (as we can see on the picture above), be we still observe that the masks of the different coils are very similar and indeed correspond to a same slice.

If anyone knows how to overcome this, it would be awesome (I was thinking about thresholding the number of different pixels allowed but it it not very rigorous and might fail in case of numerous null values...)

In any case, this does not prevent my code from working, it is just a way to easily spot future potential bugs (e.g. if I don't cover all scenarios).

shimmingtoolbox/load_nifti.py
```diff
@@ -216,37 +216,58 @@ def scale_tfl_b1(image, json_data):
...
 
+    # Phase values are stored in the second half of the 4th dimension. Siemens phase range: [248 - 3848]
+    b1_phase = image[:, :, :, image.shape[3] // 2:]
+
+    # Reorder data shuffled by dm2niix into shape (x, y , n_slices*n_coils)
+    b1_mag_vector = np.zeros((image.shape[0], image.shape[1], n_slices * n_coils))
+    b1_phase_vector = np.zeros((image.shape[0], image.shape[1], n_slices * n_coils))
+    if 'ImageOrientationPatientDICOM' in json_data:
+        # If axial slices
+        if json_data['ImageOrientationPatientDICOM'] == [1, 0, 0, 0, 1, 0]:
```
po09i, line 233: Is this always going to be the case that axial = [1, 0, 0, 0, 1, 0]? What does the vector mean?

gaspardcereza: It should be, as it correspond to this DICOM tag: https://dicom.innolitics.com/ciods/mr-image/image-plane/00200037

po09i: So it's a direction, would that mean it could be -1?

gaspardcereza: Yes I think for coronal images one value is -1. But from what I found on forums/doc there seem to be 3 cases that cover 99% of the scenarios. If I ever have an unexpected value I will just raise an error.

po09i (approved):
Looks good! 👌🏼

PR #280, opened 2021-06-16
Add realtime sequencing ability

po09i:
## Checklist

- [x] I've linked relevant issues in the PR body
- [x] I've added relevant tests for my contribution
- [x] I've updated the documentation and/or added correct docstrings
- [x] I've assigned a reviewer

## Description
We currently have a sequencer to perform slicewise shimming. This PR introduces the following:

* Better selection of slices
  * Multi-slice: [(0, 3), (1, 4), (2, 5)] → slices 0 and 3 are optimized at the same time
  * Slab: [(0, 1, 2, 3, 4, 5, 6)] → slab with all slices
  * Slicewise: [(0,), (1,), (2,), (3,), (4,), (5,), (6,)] → slice by slice
  * API to select slices: define_slices(...)
* Refactoring of the sequencer files (shim_sequencer())
* Realtime API to perform realtime shimming (shim_realtime_pmu_sequencer())
* Dilation of masks
* Added min/max attributes to the pmu

## How to test
Debug `test_sequencer.test_realtime_sequencer()`

PR #289, opened 2021-07-09
Fix realtime shimming mask resampling modes

### LINKED ISSUE #288: Realtime shimming anat masks can create unwanted resampled masks
### Bug Description
When inputting an anatomical mask in realtime shimming that masks True for the whole range of slices as the anat image, the resampling currently used (`nearest`) makes it so that values outside the mask are specified as True in the "fieldmap mask" which is not a desired behaviour. However, since we are resampling back on the anatomical image, the behaviour does not seem to have any effect on the results since the "bigger mask" is removed when resampling back on the anatomical image.

### Reproduction Steps
Run any test in test_realtime_shim where we have masks and look at the fmap_mask vs anat_masks

Originally, `nearest` was used because it would output 0 all of the time. Since then, I have found that nearest is only required when going from nii_static/nii_riro to the resampled_static/resampled_riro.

More investigation is required to figure out why this is the case and if that leads to potential problems.

po09i:
## Description
This PR addresses issue #288 where the resampled mask could include a bigger region than masked in the anatomical image. The reason is that the mode `nearest` was used when resampling to the fieldmap as it is the default value in `shimming_toolbox/coils/coordinates.resample_from_to()`. The mode `nearest` is required for the resampling back to the anatomical so it is added to be more explicit.

This does not change any output as described in the issue since resampling back on the anatomical image removes this part added. This PR addresses this for consistency and so that the masked fieldmap can be relevant.

## To reproduce
In test/shim/test_realtime_shim:test_output_figure
Run the test in debug and look at the `tmp` folder at the last line of the test
Compare `fig_anat`, `fig_mask_static` and `fig_mask_fmap_static`

## More context

- realtime_zshim requires a mask in the anat space
- the mask is resampled in the fieldmap space
  - the resampled mask is _currently_ only used for the purpose of visualization.

However, beyond the visualization purpose, our plan is to _also_ use the mask in the fieldmap space for performing shimming experiments #280. This is the motivation to get this done properly.

## Linked issues
Fixes #288

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,7 @@ setup(
         "phantominator~=0.6.4",
         "nibabel~=3.1.1",
         "requests",
-        "scipy~=1.5.0",
+        "scipy~=1.6.0",
         "tqdm",
         "matplotlib~=3.1.2",
         "psutil~=5.7.3",
--- a/shimmingtoolbox/shim/realtime_shim.py
+++ b/shimmingtoolbox/shim/realtime_shim.py
@@ -74,7 +74,7 @@ def realtime_shim(nii_fieldmap, nii_anat, pmu, json_fmap, nii_mask_anat_riro=Non
                 not np.all(nii_mask_anat_riro.shape == nii_anat.shape):
             raise RuntimeError("Mask must have the same shape and affine transformation as anat")
         nii_fmap_3d_temp = nib.Nifti1Image(fieldmap[..., 0], nii_fieldmap.affine)
-        nii_mask_fmap_riro = resample_from_to(nii_mask_anat_riro, nii_fmap_3d_temp)
+        nii_mask_fmap_riro = resample_from_to(nii_mask_anat_riro, nii_fmap_3d_temp, mode='constant')
         mask_fmap_riro = nii_mask_fmap_riro.get_fdata()
     else:
         mask_fmap_riro = np.ones_like(fieldmap[..., 0])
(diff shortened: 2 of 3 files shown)


po09i:
fsleyes where we see the fieldmap (biggest), anatomical (smaller), anat mask (smallest):
Screen Shot 2021-07-09 at 16 22 03

Previous mask (using `nearest` for resampling from anat to fmap):
Screen Shot 2021-07-09 at 16 27 42


Current mask (using `constant` for resampling from anat to fmap):
Screen Shot 2021-07-09 at 16 24 17

gaspardcereza (approved):
Everything looks good to me and as the masks are the same in the comparison you did, I guess we can assume that it works as intended now !

gaspardcereza:
@po09i I notice small lines above and below the square mask (the last one) that differ from the first one, is it something you would expect ?

po09i:
> @po09i I notice small lines above and below the square mask (the last one) that differ from the first one, is it something you would expect ?

Ideally, the 3rd figure (fmap mask) would match exactly the first one (anat mask), however it is usually not possible since the fieldmap and anat usually have different dimensions. In this case, the fieldmap mask looks "short". This is a border effect which seems to happen when we input a mask that has the first and/or last slice included in the mask. This is not perfect but definitely an improvement to what we previously had. 

Note: This is only for visual assessment, we do not use the fieldmap mask to do any computing, we instead take the results, resample back into the anatomical space and use the original anatomical mask.
I tried a couple of options to fix this as well as nilearn and got the same results.

gaspardcereza:
> Ideally, the 3rd figure (fmap mask) would match exactly the first one (anat mask), however it is usually not possible since the fieldmap and anat usually have different dimensions. In this case, the fieldmap mask looks "short". This is a border effect which seems to happen when we input a mask that has the first and/or last slice included in the mask. This is not perfect but definitely an improvement to what we previously had.

In that case I think it is ready to be merged 🙂

jcohenadad:
The problem _could_ be related to https://github.com/shimming-toolbox/shimming-toolbox/issues/291

po09i:
## Update
I found out scipy has new modes that can be used during resampling. The one relevant to us is `grid-constant` which will interpolate outside the roi, this allows to include the last row of pixels. See more info [here](https://docs.scipy.org/doc/scipy/tutorial/ndimage.html#ndimage-interpolation-modes).

The second thing I added was to interpolate according to order 0 since we are masking. Again, more info in scipy's doc page [here](https://docs.scipy.org/doc/scipy/tutorial/ndimage.html#ndimage-interpolation-modes).

## Output
Here are the new outputs:
Original mask in anat resolution:
Screen Shot 2021-07-23 at 12 45 59

Mask is resampled to fieldmap resolution using grid-constant and 0th order. (Using anything other than 0th order leaves a pixel in the corner for this dataset.
Screen Shot 2021-07-23 at 12 47 16

gaspardcereza:
Looks great ! Didn't find any problem so I think you can merge now 🙂

PR #295, opened 2021-08-10
Automate the install process

ahill187:
## Checklist

## Description
In `fsleyes-plugin-shimming-toolbox`, we automate the install of the plugin. It would be nice to do that for `shimming-toolbox` as well. 

Currently, the install entails installing `FSL`, `dcm2niix`, `conda`, and `shimming-toolbox`. We can bundle the install of `dcm2niix`, `conda`, and `shimming-toolbox`, leaving just the `FSL` install to the user.

Since `shimming-toolbox` is installed as a dependency for `fsleyes-plugin-shimming-toolbox`, we will also need to update the install procedure there as well: https://github.com/shimming-toolbox/fsleyes-plugin-shimming-toolbox/pull/19

As a side effect, the dcm2niix installation check was removed from st_check_dependencies as it is now installed in the venv and does not need to be in the path anymore.

## Test
```
git fetch
git checkout ah/installer
make install
# Make sure you can call the shimming toolbox's CLIs and that they come from your home folder
which st_prepare_fieldmap
```


Makefile
```diff
@@ -0,0 +1,21 @@
+SHELL := /bin/bash
+ST_DIR := $(HOME)/shimming_toolbox
+PYTHON_DIR := python
+CLEAN := false
+
+# Put it first so that "make" without argument is like "make help".
+help:
+	@egrep -h '\s##\s' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m  %-30s\033[0m %s\n", $$1, $$2}'
+
+install: ## Run 'make install' to install the plugin
```
po09i:
```suggestion
install: ## Run 'make install' to install Shimming Toolbox
```

Makefile
```diff
@@ -0,0 +1,21 @@
...
+	if [[ $(CLEAN) == true || ! -d $(ST_DIR) ]]; then \
+		echo "Clean install, deleting $(ST_DIR). Warning: This will delete fsleyes-plugin-shimming-toolbox."; \
```
po09i: Is this true? The plugin might not be installed

po09i:
We probably don't want to see the code...
Screen Shot 2021-08-25 at 15 43 22

ahill187:
> We probably don't want to see the code...
> Screen Shot 2021-08-25 at 15 43 22

Ah yes good point. I need to figure out how to get the Makefile to stop printing out all the lines (that's the default).

**Update** - Fixed here: https://github.com/shimming-toolbox/shimming-toolbox/pull/295/commits/93e712949c414d575cc7c44482ee1c7e061d17d9

Makefile
ahill187: Ah yes. I meant it will delete the plugin if it has been installed. I should clarify that.

ahill187: Fixed here: https://github.com/shimming-toolbox/shimming-toolbox/pull/295/commits/a087cf89eb3b2b2c7318054a78cc7cdddfc26c65

po09i:
There is a prompt when downloading dcm2niix, can we use something like `yes |` to automate this?

Screen Shot 2021-08-26 at 13 06 49

po09i:
I also ran make install CLEAN=true and got this error:

Screen Shot 2021-08-26 at 13 17 09

Makefile
```diff
@@ -0,0 +1,23 @@
...
+install: ## Run 'make install' to install Shimming Toolbox
+	if [[ $(CLEAN) == true || ! -d $(ST_DIR) ]]; then \
+		echo "Clean install, deleting $(ST_DIR). Warning: If you have previously installed fsleyes-plugin-shimming-toolbox, this will delete it."; \
+		bash installer/create_st_dir.sh; \
+	elif [[ ! -f $(ST_DIR)/$(PYTHON_DIR)/etc/profile.d/conda.sh ]]; then \
+		echo "Conda install not found in $(ST_DIR), installing conda"; \
+		bash installer/install_conda.sh; \
+	else \
+		echo "$(ST_DIR) and conda install found, skipping install"; \
+    fi
```
po09i:
```suggestion
        fi
```

ahill187:
> There is a prompt when downloading dcm2niix, can we use something like `yes |` to automate this?
> 
> Screen Shot 2021-08-26 at 13 06 49

There is a `yes` in the command so I need to see why it's being ignored...

**Update** - Fixed here: https://github.com/shimming-toolbox/shimming-toolbox/pull/295/commits/8f3184f79b6ff73ed36b462c22bc6b72c7c730e7

ahill187:
> I also ran make install CLEAN=true and got this error:
> 
> Screen Shot 2021-08-26 at 13 17 09

Ah yes, same for me. Fixed it here: https://github.com/shimming-toolbox/shimming-toolbox/pull/295#issuecomment-906595226

po09i (approved):
Looks good!

gaspardcereza:
Great work @ahill187 ! I used both `make install` and `make install CLEAN=true` and it seems to be working perfectly. Only problem: with this install, I'm unable to run `pytest` to make sure the install actually worked (`zsh: command not found: pytest`). Shouldn't pytest comme along the shimming-toolbox ? @po09i any thoughts ? Maybe I missed some talks about this

ahill187:
> Great work @ahill187 ! I used both `make install` and `make install CLEAN=true` and it seems to be working perfectly. Only problem: with this install, I'm unable to run `pytest` to make sure the install actually worked (`zsh: command not found: pytest`). Shouldn't pytest comme along the shimming-toolbox ? @po09i any thoughts ? Maybe I missed some talks about this

Oh good catch! That makes sense to me. `pytest` should be installed in the `st_venv`. In the previous install process, you had to create `st_venv` and activate it, so you could just run `pytest`. Since `st_venv` is created automatically, it isn't activated for the user. I will update the install instructions to fix that 😃 

**Update** - Should work here: https://github.com/shimming-toolbox/shimming-toolbox/pull/295/commits/abc6c83b9c48f63f56cc290a7914d1312c48b056

gaspardcereza:
In the instructions, it is written that ` conda activate st_venv` will activate the environment. On my end it does not work because st_venv has no name (it is referred to only with a path). In order to activate it, I had to type: `conda activate /Users/gaspard/shimming_toolbox/python/envs/st_venv`. 
Otherwise, everything seems to be working normally once its activated
image

jcohenadad:
> In the instructions, it is written that ` conda activate st_venv` will activate the environment. On my end it does not work because st_venv has no name (it is referred to only with a path). In order to activate it, I had to type: `conda activate /Users/gaspard/shimming_toolbox/python/envs/st_venv`.
> Otherwise, everything seems to be working normally once its activated
> image

Sourcing the conda.sh should be run before running conda activate. See [SCT's wiki](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Testing%3A-Running-the-test-suite#running-unitfunctional-tests)

this line should be added in ST documentation

po09i:
We should test the new installation process using github actions as opposed to using `pip install .`

ahill187:
> > In the instructions, it is written that ` conda activate st_venv` will activate the environment. On my end it does not work because st_venv has no name (it is referred to only with a path). In order to activate it, I had to type: `conda activate /Users/gaspard/shimming_toolbox/python/envs/st_venv`.
> > Otherwise, everything seems to be working normally once its activated
> > image
> 
> Sourcing the conda.sh should be run before running conda activate. See [SCT's wiki](https://github.com/spinalcordtoolbox/spinalcordtoolbox/wiki/Testing%3A-Running-the-test-suite#running-unitfunctional-tests)
> 
> this line should be added in ST documentation

Yes, that is in the documentation already: https://shimming-toolbox-py--295.org.readthedocs.build/en/295/getting_started/installation_st/mac.html

ahill187:
> We should test the new installation process using github actions as opposed to using `pip install .`

Should we change it to `make install` in `tests.yml`, for example?

po09i:
Changed the PATH to shimming-toolbox from shimming_toolbox to shimming-toolbox for consistency with the git clone.

po09i:
There seems to be a problem with the `$HOME` shell variable in Github CI depending on whether we are in a script or on the shell.

What I'm trying to do: Use the github actions to use our install script. When trying to add the launchers to the path, the $HOME differs from the one in the script which seems odd.

So far, it appears in ubuntu 20
- Echo $HOME in shell : `/root`
- Echo $HOME in script: `/home/runner`

Do you have some insights? @kousu @joshuacwnewton @taowa

.github/workflows/tests.yml
```diff
@@ -47,18 +47,12 @@ jobs:
       - name: Package installation
         run: |
           echo "shimming-toolbox installation"
-          pip install .
+          sudo make install
```
joshuacwnewton: > When trying to add the launchers to the path, the $HOME differs from the one in the script which seems odd.
> - Echo $HOME in shell : `/root`
> - Echo $HOME in script: `/home/runner`

`/home/runner` should be the correct default `$HOME` value for GH Actions. 

I think the reason why the shell script differs is because the shell script is invoked via the Makefile, which here uses `sudo` (see above), and [`$HOME` is `/root` for the superuser/root account](http://www.linfo.org/slash_root.html).

The latter steps don't use `sudo`, so `$HOME` remains the default value.

po09i: Thanks for the info! I found sudo's `-E` [option](https://www.sudo.ws/man/1.8.13/sudo.man.html#E) which solved my problem by giving superuser privileges but preserves the environment variables of the current user.

gaspardcereza (approved):
Everything looks good now @po09i ! Thanks for the help @joshuacwnewton ! And thanks again for the amazing work @ahill187 🙂

PR #296, opened 2021-08-25
Add shim CLI and sequencer

### LINKED ISSUE #101: Create an API and  a CLI for realtime_zshimming sequencer and optimizer

po09i:
## Description
Adds the ability to shim using the sequencer through `st_b0shim`. More precisely, the PR adds:

- `st_b0shim` CLI
- `st_b0shim` argument to use the sequencer (`st_b0shim static`, `st_b0shim realtime`, `st_b0shim gradient_realtime`)
- `st_b0shim` argument to define slices (`st_b0shim define_slices`) (Removed)
- Ability to define the slices using the "--slices" and "--slice-factor" options
- Define coil names in the constraints when defining a `Coil`
- Displays the R2 score of the linear fit in debug mode for rt sequencer
- Ability to output shim coefficients in different file formats
  - For realtime, available file formats: `slicewise-ch`, `chronological-ch`, `eva`
  - This allows for custom coils and scanner coils.
- Function to convert from physical CS to Gradient CS

## To test
Run the tests in test/cli/test_cli_b0shim

## Related
Fixes #101 
Related to https://github.com/shimming-toolbox/shimming-toolbox/issues/297


gaspardcereza (approved):
Tested it locally and it works just fine. I went through the code and didn't spot any mistake/typo. LGTM !

shimmingtoolbox/coils/siemens_basis.py
```diff
@@ -151,4 +153,16 @@ def siemens_basis(x, y, z):
     for i_channel in range(0, spher_harm.shape[3]):
         scaled[:, :, :, i_channel] = scaling_factors[i_channel] * reordered_spher[:, :, :, i_channel]
 
-    return scaled
+    # Patch to make orders work. A better implementation would be to refactor _get_scaling_factors and
+    # _reorder_to_siemens
```
gaspardcereza, line 155: You could add a comment explaining what's wrong with the orders (or refactor `_get_scaling_factors` and `_reorder_to_siemens` as suggested)

shimmingtoolbox/coils/siemens_basis.py
```diff
@@ -5,11 +5,12 @@
 from shimmingtoolbox.coils.spherical_harmonics import spherical_harmonics
 
 
+X = 0
 Y = 1
 Z = 2
 ZX = 4
 XY = 7
-GYROMAGNETIC_RATIO = 42.576
+GYROMAGNETIC_RATIO = 42.5774785178325552
```
gaspardcereza, line 13: Its a lot of precision for something that changes depending on the surroundings of the protons. Do we need to be that precise ? Also the value might differ slightly depending on the source...

config/coil_config.json
```diff
@@ -0,0 +1,15 @@
+{
+    "name": "siemens_gradient_coil",
+    "coef_channel_minmax": [
```
gaspardcereza, line 3: Are these values provided by Siemens ? Will the user that want to use a different coil setup have to create a similar file ?

shimmingtoolbox/utils.py
```diff
@@ -147,3 +147,30 @@ def create_fname_from_path(path, file_default):
         fname = path
 
     return os.path.abspath(fname)
+
+
+def set_all_loggers(verbose, list_exclude=('matplotlib',)):
+    """ Set all loggers in the root manager to the verbosity level. Exclude any logger with the name in list_exclude
+
+    Args:
+        verbose: Verbosity level: info, debug, warning, critical, error
```
gaspardcereza: Should it be a string (with '') or just the word alone (i.e. `'info'` or `info`)?

test/shim/test_sequencer.py
```diff
@@ -209,6 +213,13 @@ def test_shim_sequencer_wrong_mask_dim(self, nii_fieldmap, nii_anat, nii_mask, s
         with pytest.raises(ValueError, match="Mask image must be in 3d"):
             shim_sequencer(nii_fieldmap, nii_anat, nii_wrong_mask, slices, [sph_coil])
 
+    # def test_shim_sequencer_wrong_units(self, nii_fieldmap, nii_anat, nii_mask, sph_coil, sph_coil2, caplog):
```
gaspardcereza, line 216: Try to have 100% test coverage before merging (this way we won't have to do it later)

shimmingtoolbox/coils/siemens_basis.py
po09i: It seems better to add more precision than not having that precision at all even if it might differ slightly

config/coil_config.json
po09i: The value are currently arbitrary. The idea is indeed to have the user fill out this example file when they set up a new coil to reflect the min/max parameters.

shimmingtoolbox/utils.py
po09i: It should be a string. I updated the doc.

shimmingtoolbox/cli/b0shim.py
```diff
@@ -0,0 +1,732 @@
...
+from shimmingtoolbox.shim.shim_utils import phys_to_gradient_cs
+
+
+CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+@click.group(context_settings=CONTEXT_SETTINGS,
+             help="Shim according to the specified algorithm as an argument e.g. st_shim xxxxx")
```
jcohenadad:
```suggestion
             help="Shim according to the specified algorithm as an argument e.g. st_b0shim xxxxx")
```

shimmingtoolbox/cli/b0shim.py
```diff
@@ -0,0 +1,732 @@
...
+@click.command(context_settings=CONTEXT_SETTINGS)
+@click.option('--coil', 'coils', nargs=2, multiple=True, type=(click.Path(exists=True), click.Path(exists=True)),
+              help="Pair of filenames containing the coil profiles followed by the filename to the constraints "
+                   "e.g. --coil a.nii cons.json. If you have more than one coil, use this option more than once. "
+                   "The coil profiles and the fieldmaps (--fmap) must have matching units (if fmap is in Hz, the coil "
+                   "profiles must be in Hz/unit_shim). If you only want to shim using the scanner's gradient/shim "
+                   "coils, use the `--scanner-coil-order` option. For an example of a constraint file, "
+                   f"see: {__dir_config_scanner_constraints__}")
+@click.option('--fmap', 'fname_fmap', required=True, type=click.Path(exists=True),
+              help="Nifti filename of the B0 fieldmap. This file should contain a 3d array.")
```
jcohenadad:
```suggestion
              help="Static B0 fieldmap.")
```

shimmingtoolbox/cli/b0shim.py
```diff
@@ -0,0 +1,732 @@
...
+@click.option('--anat', 'fname_anat', type=click.Path(exists=True), required=True,
+              help="Nifti filename of the anatomical image to apply the correction onto."
```
jcohenadad: simplify wording as in the comment above

shimmingtoolbox/cli/b0shim.py
```diff
@@ -0,0 +1,732 @@
...
+                   "This file should contain a 3d array.")
+@click.option('--mask', 'fname_mask_anat', type=click.Path(exists=True), required=False,
+              help="Nifti file used to define the spatial region to shim. This file should contain a 3d array."
```
jcohenadad: should work if using 2D data

shimmingtoolbox/cli/b0shim.py
```diff
@@ -0,0 +1,732 @@
...
+                   "The coordinate system should be the same as ``anat``'s coordinate system.")
+@click.option('--scanner-coil-order', type=click.INT, default=-1, show_default=True,
+              help="Maximum order of the shim system, allowed values are 0, 1, 2. Note that specifying 2 will return "
```
jcohenadad: allowed values should be defined in the click arguments

jcohenadad: and remove choice from usage

shimmingtoolbox/cli/b0shim.py
```diff
@@ -0,0 +1,732 @@
...
+                   "orders 0, 1 and 2. The 0th order is the f0 frequency.")
+@click.option('--scanner-coil-constraints', 'fname_sph_constr', type=click.Path(exists=True),
+              default=__dir_config_scanner_constraints__, show_default=True,
+              help="Constraints for the 1st and 2nd order scanner coils")
```
jcohenadad: fragile when >2 orders will be implemented

shimmingtoolbox/cli/b0shim.py
```diff
@@ -0,0 +1,732 @@
...
+@click.option('--slices', type=click.Choice(['interleaved', 'sequential', 'volume']), required=False,
+              default='sequential', show_default=True, help="Defines the slice ordering")
+@click.option('--slice-factor', 'slice_factor', type=click.INT, required=False, default=1, show_default=True,
+              help="Number of slices per shim for 'interleaved' and 'sequential'")
```
jcohenadad:
```suggestion
              help="Number of slices per shimmed group. For example, if the value is '3', then with the 'sequential' mode, shimming will be performed independently on the following groups: {0,1,2}, {3,4,5}, etc. With the mode 'interleaved', it will be: {0,2,4}, {6,8,10}, etc.")
```

jcohenadad: possible issue: if total number of slice is not divisible by the nb of slice_factor --> issue to open-- more investigations needed

shimmingtoolbox/cli/b0shim.py
```diff
@@ -0,0 +1,732 @@
...
+@click.option('--optimizer-method', 'method', type=click.Choice(['least_squares', 'pseudo_inverse']), required=False,
+              default='least_squares', show_default=True, help="Method used by the optimizer. LS will respect the "
+                                                               "constraints, PS will not respect the constraints")
+@click.option('--mask-dilation-kernel', 'dilation_kernel',
+              type=click.Choice(['sphere', 'cross', 'line', 'cube', 'None']), required=False, default='sphere',
+              show_default=True, help="Kernel used to dilate the mask to expand the roi")
```
jcohenadad:
```suggestion
              show_default=True, help="Kernel used to dilate the mask to expand the ROI")
```

shimmingtoolbox/cli/b0shim.py
```diff
@@ -0,0 +1,732 @@
...
+              default=__dir_config_scanner_constraints__, show_default=True,
+              help="Constraints for the 1st and 2nd order scanner coils")
+@click.option('--slices', type=click.Choice(['interleaved', 'sequential', 'volume']), required=False,
+              default='sequential', show_default=True, help="Defines the slice ordering")
+@click.option('--slice-factor', 'slice_factor', type=click.INT, required=False, default=1, show_default=True,
+              help="Number of slices per shim for 'interleaved' and 'sequential'")
+@click.option('--optimizer-method', 'method', type=click.Choice(['least_squares', 'pseudo_inverse']), required=False,
+              default='least_squares', show_default=True, help="Method used by the optimizer. LS will respect the "
+                                                               "constraints, PS will not respect the constraints")
+@click.option('--mask-dilation-kernel', 'dilation_kernel',
```
jcohenadad: i would remove this option and just hard-code sphere

shimmingtoolbox/cli/b0shim.py
```diff
@@ -0,0 +1,732 @@
...
+              type=click.Choice(['sphere', 'cross', 'line', 'cube', 'None']), required=False, default='sphere',
+              show_default=True, help="Kernel used to dilate the mask to expand the roi")
+@click.option('--mask-dilation-kernel-size', 'dilation_kernel_size', type=click.INT, required=False, default='3',
+              show_default=True,
+              help="Length of a side of the 3d kernel to dilate the mask. Must be odd. For example, a kernel of size 3"
+                   "will dilate the mask by 1 pixel, 5->2 pixels")
```
jcohenadad:
```suggestion
              help="Number of voxels to consider outside of the masked area. For example, when doing dynamic shimming with a linear gradient, the coefficient corresponding to the gradient orthogonal to a single slice cannot be estimated: there must be at least 2 (ideally 3) points to properly estimate the linear term. When using 2nd order or more, more dilation is necessary.")
```

shimmingtoolbox/cli/b0shim.py
```diff
@@ -0,0 +1,732 @@
...
+@click.option('-o', '--output', 'path_output', type=click.Path(), default=os.path.abspath(os.curdir),
+              show_default=True, help="Directory to output coil text file(s).")
+@click.option('--output-file-format-coil', 'o_format_coil',
+              type=click.Choice(['slicewise-ch', 'slicewise-coil', 'chronological-ch', 'chronological-coil']),
+              default='slicewise-coil',
+              show_default=True, help="Format of the output txt file(s) for the custom coils. slicewise will output "
+                                      "one slice per row in the txt file, chronological will output one set of shim "
+                                      "per row in the order that the shim will be performed. Use 'ch' or 'coil' to "
+                                      "specify whether to output one txt file per coil system or coil channel.")
```
jcohenadad:
```suggestion
              show_default=True, help="Syntax used to describe the sequence of shim events. Use 'slicewise' to output in row 1, 2, 3, etc. the shim coefficients for slice 1, 2, 3, etc. Use 'chronological' to output in row 1, 2, 3, etc. the shim value for trigger 1, 2, 3, etc. The trigger is an event sent by the scanner and captured by the controller of the shim amplifier. Use 'ch' to output one file per coil channel (coil1_ch1.txt, coil1_ch2.txt, etc.). Use 'coil' to output one file per coil system (coil1.txt, coil2.txt). In the latter case, all coil channels are encoded across multiple columns in the text file.")
```

po09i:
- Name of method: throw away fieldmap
- Happy with st_b0shim?

shimmingtoolbox/coils/coil.py
```diff
@@ -4,9 +4,13 @@
...
 from typing import Tuple
 
+from shimmingtoolbox.coils.siemens_basis import siemens_basis
+from shimmingtoolbox.coils.coordinates import generate_meshgrid
+
 required_constraints = [
+    "name",
     "coef_channel_minmax",
-    "coef_sum_max"
+    "coef_sum_max",
```
gaspardcereza: Is this comma necessary ?

shimmingtoolbox/masking/mask_utils.py
```diff
@@ -3,15 +3,17 @@
 
 import nibabel as nib
 import numpy as np
-from scipy.ndimage import binary_dilation
-from scipy.ndimage import binary_opening
-from scipy.ndimage import generate_binary_structure
-from scipy.ndimage import iterate_structure
+import logging
+from scipy.ndimage import binary_dilation, binary_opening, generate_binary_structure, iterate_structure
+import os
```
gaspardcereza, line 7: "import" should come before "from"

shimmingtoolbox/shim/realtime_shim.py
```diff
@@ -109,6 +109,7 @@ def realtime_shim(nii_fieldmap, nii_anat, pmu, json_fmap, nii_mask_anat_riro=Non
     for it in range(nt):
         gradient[:][..., it] = phys_gradient(g * fieldmap[:, :, :, it], nii_fieldmap.affine)  # [mT / mm]
     gradient *= 1000  # [mT / m]
+    # gradient *= 1000  # [uT / m]
```
gaspardcereza: Why the same multiplicator for mT and uT ? Do you want to keep the second line ?

test/shim/test_sequencer.py
```diff
@@ -16,14 +16,17 @@
...
 from shimmingtoolbox.shim.sequencer import update_affine_for_ap_slices
+from shimmingtoolbox.utils import set_all_loggers
 
 import numpy as np
 import pytest
 import os
 import nibabel as nib
 import json
-from matplotlib.figure import Figure
+import logging
```
gaspardcereza: library imports should come before individual functions are imported

test/shim/test_sequencer.py
```diff
@@ -69,8 +72,9 @@ def create_constraints(max_coef, min_coef, sum_coef, n_channels=8):
         bounds.append((min_coef, max_coef))
 
     constraints = {
+                "name": "test",
                 "coef_sum_max": sum_coef,
-                "coef_channel_minmax": bounds
+                "coef_channel_minmax": bounds,
```
gaspardcereza: Same question as for previous comma

test/shim/test_sequencer.py
```diff
@@ -209,6 +213,13 @@ def test_shim_sequencer_wrong_mask_dim(self, nii_fieldmap, nii_anat, nii_mask, s
         with pytest.raises(ValueError, match="Mask image must be in 3d"):
             shim_sequencer(nii_fieldmap, nii_anat, nii_wrong_mask, slices, [sph_coil])
 
+    # def test_shim_sequencer_wrong_units(self, nii_fieldmap, nii_anat, nii_mask, sph_coil, sph_coil2, caplog):
+    #     # Change the name of the units
+    #     sph_coil2.units = "T"
+    #     slices = [(0, 2), (1,)]
+    #     shim_sequencer(nii_fieldmap, nii_anat, nii_mask, slices, [sph_coil, sph_coil2])
+    #     assert "The coils don't have matching units:" in caplog.text
+
```
gaspardcereza, line 222: Do you plan to implement this test later ? Otherwise I would just remove it

test/test_siemens_basis.py
```diff
@@ -5,8 +5,10 @@
 import numpy as np
 import pytest
 import math
+import pathlib
+import tempfile
```
gaspardcereza: Why the line jump ?

shimmingtoolbox/shim/realtime_shim.py
po09i: Uncommenting the second line will output in uT/m, probably not necessary anymore.

test/shim/test_sequencer.py
po09i: Yes, this will eventually be added

shimmingtoolbox/coils/coil.py
po09i: Resolved in 1a5ce6c.

shimmingtoolbox/masking/mask_utils.py
po09i: Resolved in 1a5ce6c.

shimmingtoolbox/shim/realtime_shim.py
po09i: Resolved in 1a5ce6c.

test/shim/test_sequencer.py
po09i: Resolved in 1a5ce6c.

test/shim/test_sequencer.py
po09i: Resolved in 1a5ce6c.

test/test_siemens_basis.py
po09i: Resolved in 1a5ce6c.

shimmingtoolbox/optimizer/lsq_optimizer.py
```diff
@@ -75,12 +76,20 @@ def _apply_sum_constraint(inputs, indexes, coef_sum_max):
...
-                                   bounds=self.merged_bounds,
-                                   constraints=tuple(constraints),
-                                   options={'maxiter': 500})
+        # When clipping to bounds, scipy raises a warning. Since this can be frequent for our purposes, we ignore that
+        # warning
+        with warnings.catch_warnings():
+            warnings.filterwarnings('ignore', message="Values in x were outside bounds during a "
+                                                      "minimize step, clipping to bounds",
+                                    category=RuntimeWarning,
+                                    module='scipy')
```
po09i, line 179: When optimizing, scipy raises a warning if the optimization reaches the bounds. This causes 2 problem, one is that this raises warnings for the tests. It also clutters the output on the terminal for what is an "expected" behaviour (That's why we have bounds). I'm wondering if this is acceptable practice to silence this warning. @jcohenadad @joshuacwnewton

jcohenadad: usually warnings are there for a reason, but in this case I understand that it has to do with the optimization process and that it _could_ be ignored as long as the information is always irrelevant for the user.

For example: could we imagine a scenario where an optimization method would reach the bounds for many channels, and that solution would be less optimal than with another optimization method, in which case the user might want to know that they should maybe try using another optimization method?  Alternatively, as part of the output figures, you could output a graphs with all channels and their shim voltage, and showing the bounds on the graphs, so the user would have a quick overview about the number of maxed out channels (and then decide whether they want to try another method)

joshuacwnewton: My apologies, I missed this ping! I don't have prior experience with scipy's optimization tools, so it's tricky for me to comment on what practices are typical. (I also don't have much prior exposure to section of ST's codebase.)

But, if you still need another set of eyes on this, please feel free to re-ping me and I can try to dig into this warning a little more.

PR #298, opened 2021-09-01
Change command that checks for sct installation

po09i:
## Description
SCT can be installed at different places in the OS and removed which can lead to `which sct_scheck_dependencies` to display a path but actually using sct_check_dependencies will return an error since sct is not installed anymore. This outputs an error on the terminal which is not the correct behaviour. This can be solved by using `sct_check_dependencies` instead of the `which`.

### DIFF
--- a/shimmingtoolbox/cli/check_env.py
+++ b/shimmingtoolbox/cli/check_env.py
@@ -122,7 +122,8 @@ def check_sct_installation():
         bool: True if sct is installed, False if not.
     """
     try:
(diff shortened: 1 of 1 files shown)


gaspardcereza:
LGTM now that we solved the problem with the SCT install

gaspardcereza (approved):
LGTM

po09i:
Our CI would not pass, we opened and [issue](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3511) on SCT as there was a bug in the install process. This was fixed [here](https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3512). Everything looks good and passes now.

PR #299, opened 2021-09-08
Implement RF shimming code

### LINKED ISSUE #208: Add RF-shimming functionalities to the toolbox
### Feature Description:
RF-shimming corresponds to the process of homogenizing the RF-field by optimizing the "shim-weights" (magnitude and phase) of every transmit elements. 

The main steps of the RF-shimming process are: 
1) Acquire a B<sub>1</sub> map for each transmit coil over the desired volume on the scanner,
2) Load the B<sub>1</sub> maps on the shimming-software (here `shimming-toolbox`),
3) Load SAR matrices that have been simulated on a software,
4) Run an optimization algorithm (e.g. reducing the [coefficient of variation](https://en.wikipedia.org/wiki/Coefficient_of_variation) while using the SAR  matrices as a constraint,
5) Send the obtained shim-weights to the scanner,

Many types of RF-shimming algorithms can be found in literature (e.g. slice-by-slice, dynamic, SAR efficient...). Here, we would first focus on a simple static volume-wise RF-shimming (one set of shim-weights is calculated to homogenize over the volume of interest). 

Here are suggestions of how we could organise our implementation to go through all these steps (**feedback is more than welcome !**) : 

- The B<sub>1</sub> maps are stored in a `DICOM` format. We would first have to convert them into `NIFTI` to be consistent with the B<sub>0</sub> mapping process. (This matter has already been discussed in #58 but we are still facing some issues due to the type of `DICOM` we have).

- Once stored into `NIFTI` files, the B<sub>1</sub> maps could be loaded using the [load_nifti](https://github.com/shimming-toolbox/shimming-toolbox/blob/master/shimmingtoolbox/load_nifti.py) function (some adaptation might be needed)

- Regarding the SAR matrices: The Specific Absorption Rate corresponds to the energy received by the patient during the scan. For safety concerns, it must not exceed a specific threshold. SAR cannot be directly measured in the patient so it has to be estimated based on electro-magnetic simulations. 
The loading of the SAR matrices will mainly depend on the format we use to store them. In the Matlab code, they were just stored as `.mat` file containing a `nCoils x nCoils x nMeshes` matrix. This was obtained by using a MATLAB script on the CST (electromagnetic simulation software) raw output files to shape it into a `.mat` object. For our python code, I can think of several possibilities: 
    - Building a script that build a `numpy array` from the CST file (would not necessarily work for other softwares),
    - Considering that the user already has a SAR file for his coil and just load it (I don't know how most people store their SAR matrices but we would have to adapt to as many cases as we can imagine).

- Regrid the B<sub>1</sub> maps (based on the B<sub>0</sub> maps) in order to apply the same masking (we already have regridding functions that we could use/adapt).

- For the constrained optimization, we could chose to go with specific Python modules or try to adapt our own optimizer. 

- Sending the shim-weights to the scanner could be done manually (easy but time consuming and prone to human error) or automatically (might be hard to implement but worth it on a long term basis --> might not work with different scanners)

- It could be useful to display the expected  B<sub>1</sub> distribution  and the corresponding coefficient of variation once the optimization has been performed.

- I think it could be interesting to reuse/adapt some functions that we already have in the toolbox rather than developing slightly different ones (I think the code would thus remain clearer but I may be wrong).
 
- The `CLI `/`API` organisation of the code would be close to what we use for B<sub>0</sub> shimming in order to be logical for the user.

**This issue intends to be as clear as possible to everybody involved in the project so if blurry areas remain, please tell me and I will do my best to clarify things.**

### LINKED ISSUE #282: Implement code for CP phase computation
### Feature Description:
In order to RF-shim, it can be useful to start with initial shim-weights based on the CP-mode (it usually yields a better homogeneity during the optimization process).

The actual weights used during the B1 mapping are stored in the `dicom` header and can be retrieved by parsing it. They correspond to CP weights present in the coil files during the acquisition (if online RF-shim is applied, they might not be the same > to check).
They can also be computed from the B1 maps by playing with the shim weights in order to get the same phase in a specified voxel (in the center of the coil)
Finally, for circular volume coils with regularly spaced elements, a fairly good estimation of the CP weights it obtained with regularly spaced phase applied to each element between 0 and 2pi.

Ideally, these three solutions should be implemented, in order to cover all possible scenarios.

gaspardcereza:
## Description
We want to include RF shimming functionalities as part of the `shimming-toolbox` project.

What we want is a way to input B1 maps acquired at the scanner and to compute complex shim weights that will mitigate the RF field inhomogeneities.

## How to test it
So far, this code is only compatible with Siemens standard RF mapping sequence. 
The user has to run `st_dicom_to_nifti` directly on the dicoms from the scanner, and then run `st_b1_shim` while providing the path to the created NIfTI file containing the B1 data (the file should have "TB1map" in its name, according to the BIDS convention). 
The code will then output:
- A text file containing the shim weights (magnitude and phase, as required in `options/adjustments/b1 shim` on Siemens scanners)
- A figure with the B1 distribution (single pulse, cp mode, shimmed, ROI mask)

image

image

## What remains to be done
Use same masking/resampling as B0 shimming (see #311)
Find a way to send the shim weights automatically to the scanner

## Linked issues
RF-shimming: #208 
CP phase computation: #282


shimmingtoolbox/b1/b1_shim.py
```diff
@@ -0,0 +1,270 @@
...
+        pass
+    else:
+        raise ValueError(f"The number of shim weights does not match the number of channels.\n"
+                         f"Number of shim weights: {len(weights)}\n"
+                         f"Number of channels: {b1_maps.shape[-1]}")
+
+    return np.abs(np.matmul(b1_maps, weights))
+
+
+def cov(array):
```
jcohenadad: probably something that can be used for b0 shimming as well

shimmingtoolbox/b1/b1_shim.py
```diff
@@ -0,0 +1,270 @@
...
+    """
+    Computes static optimized shim weights that minimize the B1 field coefficient of variation over the masked region.
+
+    Args:
+        b1_maps (numpy.ndarray): 4D array  corresponding to the measured B1 field. (x, y, n_slices, n_channels)
+        mask (numpy.ndarray): 3D array corresponding to the region where shimming will be performed. (x, y, n_slices)
+        cp_weights (numpy.ndarray): 1D vector of length n_channels of complex weights corresponding to the CP mode of
+        the coil. Must be normalized.
+        algo (int): Number from 1 to 5 specifying which algorithm to use for B1 optimization:
+                    1 - Optimization aiming to reduce the coefficient of variation (CoV) of the resulting B1+ field.
```
jcohenadad: maybe consider merging the optim functions for usage with B0 shimming. Also tagging @po09i

shimmingtoolbox/b1/b1_shim.py
po09i: scipy.stats.variation does something similar

config/cp_mode.json
```diff
@@ -0,0 +1,10 @@
+{
+  "weights": ["0.353553391+0j",
+              "0.25-0.25j",
+              "0-0.353553391j",
+              "-0.25-0.25j",
+              "-0.353553391-0j",
+              "-0.25+0.25j",
+              "0+0.353553391j",
+              "0.25+0.25j"]
+}
```
po09i: Add a new line

shimmingtoolbox/b1/b1_shim.py
```diff
@@ -0,0 +1,293 @@
...
+    """
+    if b1_maps.ndim == 4:
+        x, y, n_slices, n_channels = b1_maps.shape
+    else:
+        raise ValueError(f"The provided B1 maps have an unexpected number of dimensions.\nExpected: 4\nActual: "
+                         f"{b1_maps.ndim}")
+
+    if mask is None:
+        # If no mask is provided, shimming is performed over all non-zero values
+        mask = b1_maps.sum(axis=-1) != 0
```
po09i: This seems odd to me. Maybe using one of the masking apis? `shimmingtoolbox.masking.threshold` might be useful

shimmingtoolbox/b1/b1_shim.py
```diff
@@ -0,0 +1,293 @@
...
+
+    if b1_maps.shape[:-1] == mask.shape:
+        b1_roi = np.reshape(b1_maps * mask[:, :, :, np.newaxis], [x * y * n_slices, n_channels])
+        b1_roi = b1_roi[b1_roi[:, 0] != 0, :]
```
po09i: Explain what you are doing here please

shimmingtoolbox/b1/b1_shim.py
```diff
@@ -0,0 +1,293 @@
...
+            raise ValueError(f"The number of CP weights does not match the number of channels.\n"
+                             f"Number of CP weights: {len(cp_weights)}\n"
+                             f"Number of channels: {b1_maps.shape[-1]}")
+    else:
+        cp_weights = calc_cp(b1_maps)
+
+    # Initial weights for optimization
+    weights_init = complex_to_vector(cp_weights)
+
+    if algorithm == 1:
```
po09i: You could map your different functions to a dictionary so that each algorithm could stand on its own. There is an example in `shimmingtoolbox.masking.shapes:shapes`. `shape_mask` is the dictionnary.

shimmingtoolbox/b1/b1_shim.py
```diff
@@ -0,0 +1,293 @@
...
+    elif algorithm == 3:
+        # Maximizing the minimum B1+ value to get better RF efficiency
+        def cost(weights):
+            b1_abs = combine_maps(b1_roi, vector_to_complex(weights))
+            return 1 / np.min(b1_abs)
+
+    else:
+        raise ValueError(f"The specified algorithm does not exist. It must be an integer between 1 and 5.")
+
+    if q_matrix is not None:
```
po09i: Comments as to why this is necessary

shimmingtoolbox/b1/load_vop.py
```diff
@@ -0,0 +1,30 @@
...
+    if os.path.exists(path_sar_file):
+        sar_data = scipy.io.loadmat(path_sar_file)
+    else:
+        raise FileNotFoundError('The SarDataUser.mat file could not be found.')
+    # Assert file exists
+
+    if 'ZZ' in sar_data:
+        return sar_data['ZZ']
+    else:
+        raise ValueError('The SAR data does not contain the expected VOP values.')
```
po09i:
```suggestion
    if 'ZZ' not in sar_data:
        raise ValueError('The SAR data does not contain the expected VOP values.')

    return sar_data['ZZ']
```

shimmingtoolbox/cli/b1_shim.py
```diff
@@ -0,0 +1,77 @@
...
+    if fname_mask is not None:
+        nii_mask = nib.load(fname_mask)
+    else:
+        nii_mask = None
+
+    # If a path to a cp json file is provided, read it and store the values as complex numbers
+    # See example of json file in config/cp_mode.json (Do not add spaces within the complex values)
+    if cp_weights is not None:
+        f = open(cp_weights)
+        cp_json = json.load(f)
```
po09i:
```suggestion
        with open(fname_cp_weights) as json_file:
            jcp_json= json.load(json_file)
```

shimmingtoolbox/cli/b1_shim.py
```diff
@@ -0,0 +1,77 @@
...
+CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
+
+
+@click.command(context_settings=CONTEXT_SETTINGS)
+@click.option('--b1map', 'fname_b1_map', required=True, type=click.Path(exists=True),
+              help="Path to B1 nifti as returned by st_dicom_to_nifti.")  # TODO consider using pre scaled B1 maps here
+@click.option('--mask', 'fname_mask', type=click.Path(exists=True), required=False,
+              help="3D nifti file used to define the static spatial region to shim. "
+                   "The coordinate system should be the same as ``b1map``'s coordinate system.")
+@click.option('--cp', 'cp_weights', type=click.Path(exists=True), required=False,
```
po09i:
```suggestion
@click.option('--cp', 'fname_cp_weights', type=click.Path(exists=True), required=False,
```

shimmingtoolbox/cli/b1_shim.py
```diff
@@ -0,0 +1,77 @@
...
+                   "3 - Maximizes the minimum B1+ value for better efficiency.")
+@click.option('--target', 'target', type=float, required=False,
+              help="Target B1+ value used by algorithm 2 in nT/V")
+@click.option('--SED', 'SED', type=float, required=False,
+              help="Factor (=> 1) to which the local SAR after optimization can exceed the CP mode local SAR."
+                   "SED between 1 and 1.5 usually work with Siemens scanners. Higher SED allows more liberty for RF"
+                   "shimming but might result in SAR excess at the scanner.")
+@click.option('-o', '--output', 'path_output', type=click.Path(), default=os.path.join(os.curdir),
+              show_default=True, help="Directory to output shim weights text file and figures.")
+def b1_shim_cli(fname_b1_map, fname_mask, cp_weights=None, algorithm=1, target=None, q_matrix=None, SED=1.5,
```
po09i:
```suggestion
def b1_shim_cli(fname_b1_map, fname_mask, fname_cp_weights =None, algorithm=1, target=None, q_matrix=None, SED=1.5,
```

shimmingtoolbox/utils.py
```diff
@@ -112,7 +112,7 @@ def st_progress_bar(*args, **kwargs):
     return tqdm.tqdm(*args, **kwargs)
 
 
-def create_output_dir(path_output, is_file=False, output_folder_name="output"):
+def create_output_dir(path_output: object, is_file: object = False, output_folder_name: object = "output") -> object:
```
po09i: What does this do/why is it necessary?

test/b1/test_b1_shim.py
```diff
@@ -0,0 +1,170 @@
+#!usr/bin/env python3
+# -*- coding: utf-8
+
+import pathlib
+import pytest
+import tempfile
+from shimmingtoolbox import __dir_testing__
```
po09i: Separate with a newline

test/b1/test_load_vop.py
```diff
@@ -0,0 +1,34 @@
+#!usr/bin/env python3
+# -*- coding: utf-8
+
+import numpy as np
+import os
+import pytest
+import scipy
+from shimmingtoolbox import __dir_testing__
```
po09i, line 9: Newline

test/cli/test_cli_b1_shim.py
```diff
@@ -0,0 +1,46 @@
+#!usr/bin/env python3
+# coding: utf-8
+
+import os
+import pathlib
+import tempfile
+
+from click.testing import CliRunner
```
po09i:
```suggestion
from click.testing import CliRunner

```

shimmingtoolbox/b1/b1_shim.py
gaspardcereza: I just want my mask to be zeros and ones so the threshold function does not work for me. We will discuss that further when homogenizing masking between B0 and B1 but I prefer to let it as is right now.

shimmingtoolbox/b1/b1_shim.py
gaspardcereza: That's elegant indeed but it makes the code much more complicated (e.g. if someone wants to improve it in the future, they might get a bit confused). So that's something we can discuss further

shimmingtoolbox/utils.py
gaspardcereza: Weird... I don't remember changing that... I'll change it back to what it was

shimmingtoolbox/cli/b1_shim.py
```diff
@@ -0,0 +1,91 @@
...
+    containing shim weights for each transmit element.
+    """
+
+    # Load B1 map
+    nii_b1, json_b1, b1_map = read_nii(fname_b1_map)
+
+    create_output_dir(path_output)
+
+    # Save uncombined B1 map as nifti
+    nii_b1.header["datatype"] = 32  # 32 corresponds to complex data
```
po09i: If the rescaling is done in `read_nii()`, then the datatype should also be changed in `read_nii()`

shimmingtoolbox/cli/b1_shim.py
```diff
@@ -0,0 +1,91 @@
...
+    # s_form are bogus with tfl_rfmap so use qform instead when creating NIfTI image (makes it readable in fsleyes)
```
po09i: How do you know this is a `tfl_rfmap` sequence at this point? Should this be something we check here or maybe define the affine in `read_nii()`?

shimmingtoolbox/masking/threshold.py
```diff
@@ -25,4 +16,4 @@ def threshold(data, thr=30):
     Returns:
         numpy.ndarray: Boolean mask with same dimensions as data
     """
-    return data > thr
+    return abs(data) > thr
```
po09i: As discussed, this might lead to unwanted results if there are negative inputs. (I would expect -35 to be False).

shimmingtoolbox/cli/b1_shim.py
gaspardcereza: Right now that's the only sequence that works in shimming toolbox. If it's not a `tfl_rfmap` image, an error will be returned during `read_nii`'s exectution specifying that the sequence is unknown.

shimmingtoolbox/masking/threshold.py
gaspardcereza: Should be fixed by [3cf8dd4](https://github.com/shimming-toolbox/shimming-toolbox/pull/299/commits/3cf8dd4def3aa4239ca5b838a2ea95f7a3f3923f)

shimmingtoolbox/cli/b1_shim.py
gaspardcereza: Fixed in [3636a0e](https://github.com/shimming-toolbox/shimming-toolbox/pull/299/commits/3636a0e767ce00370df7c8185d448bbd5d6dcd3f)

shimmingtoolbox/masking/threshold.py
gaspardcereza: Idea: simply check if data are complex instead of passing an argument

config/cp_mode.json
```diff
@@ -0,0 +1,10 @@
+{
```
jcohenadad, line 1: de we want to create a subfolder b1/

shimmingtoolbox/cli/b1_shim.py
```diff
@@ -0,0 +1,88 @@
...
+@click.option('--cp', 'fname_cp_weights', type=click.Path(exists=True), required=False,
+              help="json file containing CP weights. See config/cp_mode.json for example.")
+@click.option('--algo', 'algorithm', type=int, default=1, show_default=True,
+              help="Number from 1 to 3 specifying which algorithm to use for B1 optimization"
+                   "1 - Optimization aiming to reduce the coefficient of variation (CoV) of the resulting B1+ field."
+                   "2 - Magnitude least square optimization targeting a specific B1+ value. Target value required."
+                   "3 - Maximizes the minimum B1+ value for better efficiency.")
+@click.option('--target', 'target', type=float, required=False,
+              help="Target B1+ value used by algorithm 2 in nT/V")
+@click.option('--SED', 'SED', type=float, required=False,
```
jcohenadad: i'd rather we keep small caps for flags

setup.py
```diff
@@ -24,6 +24,8 @@
             "st_mask=shimmingtoolbox.cli.mask:mask_cli",
             "st_dicom_to_nifti=shimmingtoolbox.cli.dicom_to_nifti:dicom_to_nifti_cli",
             "st_prepare_fieldmap=shimmingtoolbox.cli.prepare_fieldmap:prepare_fieldmap_cli",
+            "st_prepare_b1_map=shimmingtoolbox.cli.prepare_b1_map:prepare_b1_map_cli",
```
jcohenadad: should not be there

config/cp_mode.json
gaspardcereza: @po09i What are your thoughts about this ? I recall at some point we were talking about having b1/b0 folders but I guess both `b1shim` and `load_vop` could go directly in `shimmingtoolbox` (I even think about putting `load_vop` as a simple function in `b1shim`)

po09i: Moved this conversation to #208

po09i (approved):
Looks good!

PR #304, opened 2021-09-22
Remove the need for sudo in the installer

kousu:
## Checklist

## Description

The install script shouldn't demand root privileges. One of the big reasons for using conda is to make it easy to install dependencies without root; but even if this wasn't using conda it's very possible to package it without root.

## Linked issues

* https://github.com/neuropoly/computers/issues/201#issue-1004356523

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -46,8 +46,7 @@ jobs:
       - name: Package installation
         run: |
           echo "shimming-toolbox installation"
-          # Use -E to preserve the environment variables of the GA CI runner
-          sudo -E make install
+          make install
 
       - name: Set path for Shimming Toolbox
         run: |
--- a/docs/source/getting_started/installation_st/linux.rst
+++ b/docs/source/getting_started/installation_st/linux.rst
@@ -12,6 +12,18 @@ Install FSL
 You will need to install `FSL <https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/FslInstallation>`__ if you want to use ``prelude``.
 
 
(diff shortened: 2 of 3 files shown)


po09i:
We needed to use sudo in the CI when installing with the script, I'll remove that dependency now so we can test the install in the CI as well

kousu:
In SCT we just assume that `curl` is so basic it's probably already there, and error out if it's not:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/b745313dcf5a342de9eda4393d3c102195704993/install_sct#L180

If the CI machine doesn't have it installed, instead of _removing_ it you could _move_ it from the install script to directly in the CI script, as an extra step, like

```
      - name: Package installation
        run: |
          echo "shimming-toolbox installation"
          sudo apt install -y curl

      - name: Package installation
        run: |
          echo "shimming-toolbox installation"
          make install
```

kousu:
I haven't yet found a way to 100% bootstrap from nothing. You either need to:

- manually install Ubuntu
- manually install Windows
- manually install curl
- manually install docker
- manually install chocolatey
- etc etc

in order to get something going. It's just a matter of choosing where the line is for your project and tbh I think this is a reasonable split.

po09i:
I think this is reasonable as well. I tested it on Joplin and it works! Thanks @kousu !

kousu:
I added a note to the Linux install instructions about `curl`. I guess we're assuming Macs come with curl pre-installed always so I left that part alone?

po09i (approved):
LGTM!

PR #309, opened 2021-10-13
Add output of phase for anat images in dcm2bids

po09i:
## Description
This PR adds the ability to output phase anatomical images when using `st_dicom_to_nifti`. This required to add lines in the dcm2bids config file. 

Reason: The change was required since outputting both the phase and magnitude from the scanner would convert both mag and phase to the same file names. Now they have different names.

### DIFF
--- a/config/dcm2bids.json
+++ b/config/dcm2bids.json
@@ -194,6 +194,7 @@
             "modalityLabel": "unshimmed_e1",
             "criteria": {
                 "SeriesDescription": "*gre*realtime*",
+                "ImageType": ["*", "*", "M", "*"],
                 "EchoNumber": "1"
             }
         },
@@ -202,6 +203,7 @@
             "modalityLabel": "unshimmed_e2",
             "criteria": {
                 "SeriesDescription": "*gre*realtime*",
+                "ImageType": ["*", "*", "M", "*"],
                 "EchoNumber": "2"
             }
         },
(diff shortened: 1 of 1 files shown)


config/dcm2bids.json
```diff
@@ -210,6 +212,34 @@
             "modalityLabel": "unshimmed_e3",
```
gaspardcereza, line 212: Don't you want to add "magnitude" in the `modalityLabel` to be consistent with the phase ?

config/dcm2bids.json
```diff
@@ -210,6 +212,34 @@
...
+                "ImageType": ["*", "*", "P", "*", "*"],
+                "EchoNumber": "1"
+            }
+        },
+        {
+            "dataType": "anat",
+            "modalityLabel": "unshimmed_phase_e2",
+            "criteria": {
+                "SeriesDescription": "*gre*realtime*",
+                "ImageType": ["*", "*", "P", "*", "*"],
```
gaspardcereza, line 233: For magnitude images there are 4 elements in `ImageType` and for phase images there are 5. Is it correct ?

po09i: Yes

config/dcm2bids.json
po09i: From what I have seen in the [bids specification](https://bids-specification.readthedocs.io/en/stable/02-common-principles.html#single-session-example) and sample datasets, the anat files do not add a suffix for the magnitude images, only phase images.

gaspardcereza (approved):
LGTM

PR #316, opened 2021-11-09
Improved RF shimming output figure

gaspardcereza:
## Description
I corrected the displaying of the RF-shimming output figures based on feedback I received. 

The set the maximum range value, I use the 99th percentile of B1 so that the rare intense noisy pixels don't make the whole image too dark. 

I also removed the single pulse excitation as we deemed it irrelevant in a previous meeting,

### Old version; 
image

### New version:
image


po09i (approved):
Approved once conflicts are resolved

PR #318, opened 2021-11-10
Add norm constraint to shim-weights when not SAR constrained

gaspardcereza:
## Description
I encountered a RF-shimming scenario where performing SAR unconstrained optimization resulted in very large shim-weights magnitude (which is not surprising if the optimization process finds a local minimum corresponding to large shim weights value). In this particular case, this makes the RF-shimming functionality useless at the scanner as the odds for these shim weights to not excess the SAR limit are very low. 

I thus decided to add a constraint on the norm of the shim weights (it must always remain 1) during the optimization process. This will make sure that the total RF power used by the scanner is the same as for the CP mode (but the power distribution among the different channels may vary). This solution does not assert that we won't excess the SAR limit, but it highly reduces the odds.


po09i (commented):
Is this standard to restrict the norm to 1 is no SAR limit is available?

gaspardcereza:
> Is this standard to restrict the norm to 1 is no SAR limit is available?

I would not say it's a standard because in literature everybody presents SAR constrained shimming (because that's the way it should be done). But in the cases where you can't apply SAR constraints (e.g. because you don't have the VOP files), that's an interesting way to increase the chances that your shimming will be accepted by the scanner. I know that Kyle often works with normalized shim weights to make sure he always uses the same transmit power.

po09i (approved):

PR #321, opened 2021-12-03
Improve B1 shimming: implement resampling of masks into B1 maps + phase only shimming and add documentation

### LINKED ISSUE #311: Use same masking/resampling pipeline for B1 and B0 shimming
### Description
The shimming-toolbox aims to perform B0 and B1 shimming. We want to shim both B0 and B1 in the same region of interest, so we need to use the same mask.

Using same resampling/masking functionalities for both B0 and B1 shimming.

gaspardcereza:
## Description
In order to perform RF-shimming in a segmented region, the segmented mask needs to be resampled to have the same coordinates as the B1 maps.

On another end, CP computation, which either required additional input from the user or was prone to be done in a masked out region, has been replaced with a phase-only shimming step prior to the complex RF-shimming optimization, so that these phases may be used as a starting point for the optimization.

Corresponding unit tests are also added.

Documentation on B1+ shimming is also added.

## Details

In order to perform the resampling, the affine matrices of the B1 maps need to be reconstructed by shimming-toolbox because values in the 'ImageOrientationPatient" are not correctly saved in the `tfl_rfmap` DICOM tags. 

Note: This PR makes RF-shimming only compatible with `dcm2niix v1.0.20211006` and newer. This should not be a problem at it is the current default version installed by conda.

## Linked issues
Closes #311


shimmingtoolbox/load_nifti.py
```diff
@@ -243,22 +260,22 @@ def scale_tfl_b1(image, json_data):
     # Reorder data shuffled by dm2niix into shape (x, y , n_slices*n_channels)
     b1_mag_vector = np.zeros((image.shape[0], image.shape[1], n_slices * n_channels))
     b1_phase_vector = np.zeros((image.shape[0], image.shape[1], n_slices * n_channels))
-    if 'ImageOrientationPatientDICOM' in json_data:
-        orientation = json_data['ImageOrientationPatientDICOM']
-        # Axial or Coronal cases
-        if orientation[:3] == [1, 0, 0]:
+    if 'ImageOrientationText' in json_data:
```
po09i, line 267: I added a function to find the orientation of an acquisition, since the `ImageOrientationText` field comes from a private dice header field that is not always populated, I suggest having a fallback method that uses the function get_main_orientation() that was implemented in #323

gaspardcereza: @po09i I updated everything as we discussed earlier

shimmingtoolbox/load_nifti.py
```diff
@@ -181,11 +181,35 @@ def read_nii(fname_nifti, auto_scale=True):
...
-            nii = nib.Nifti1Image(image, nii.header.get_qform(), header=nii.header)
+            nii.header['datatype'] = 32  # 32 corresponds to complex data
+            nii.header['aux_file'] = 'Uncombined B1+ maps'
+            # Affine matrices are bogus with tfl_rfmap so we rebuild them from scratch
+            qfac = nii.header['pixdim'][0]
+
+            if 'ImageOrientationPatientDICOM' not in json_data:
+                raise KeyError("Missing json tag: 'ImageOrientationPatientDICOM'. Check dcm2niix version.")
+
+            print(json_data['ImageOrientationPatientDICOM'][4])
```
po09i: What is the purpose of this print?

shimmingtoolbox/cli/b1shim.py
```diff
@@ -52,12 +53,15 @@ def b1shim_cli(fname_b1_map, fname_mask, fname_cp_weights, algorithm, target, fn
     nib.save(nii_b1, fname_nii_b1)
     file_json_b1 = open(os.path.join(path_output, 'TB1maps_uncombined.json'), mode='w')
     json.dump(json_b1, file_json_b1)
+    b1_map_combined = b1_map.sum(axis=-1)
+    nii_b1_map_combined = nib.Nifti1Image(b1_map_combined, nii_b1.affine, header=nii_b1.header)
 
     # Load static anatomical mask
     if fname_mask is not None:
         nii_mask = nib.load(fname_mask)
+        mask_resampled = resample_from_to(nii_mask, nii_b1_map_combined, order=1, mode='grid-constant').get_fdata()
```
po09i, line 60: I would suggest having a figure or a debug option to look at the resampled mask.

shimmingtoolbox/load_nifti.py
gaspardcereza: That should not be there thanks I probably used it for debugging thanks ! I'll change that

shimmingtoolbox/cli/b1shim.py
gaspardcereza: The mask is visible on the output figure
image

shimmingtoolbox/load_nifti.py
gaspardcereza: Removed by commit 3a0dbd1

jcohenadad (changes requested):
there is no added documentation for this new feature-- i would strongly advise to add the appropriate documentation along with this PR

gaspardcereza:
> there is no added documentation for this new feature-- i would strongly advise to add the appropriate documentation along with this PR

I integrated the RF-shimming functions into the documentation and corrected the docstrings that were not correctly displayed.

jcohenadad (dismissed):
- add mention of B1 shimming in the [intro](https://shimming-toolbox-py--321.org.readthedocs.build/en/321/#welcome-to-shimming-toolbox)
- same in the [feature](https://shimming-toolbox-py--321.org.readthedocs.build/en/321/#features)
- add scenario B1 shimming in [getting started](https://shimming-toolbox-py--321.org.readthedocs.build/en/321/getting_started/getting_started.html#getting-started)

jcohenadad:
> I integrated the RF-shimming functions into the documentation

I don't see it-- can you please point to a link

gaspardcereza:
> add mention of B1 shimming in the intro
> same in the feature
> add scenario B1 shimming in getting started

This should be covered by c16f6db.

> I don't see it-- can you please point to a link

It should be visible in [this build](https://shimming-toolbox-py--321.org.readthedocs.build/en/321/).

README.rst
```diff
@@ -48,6 +48,7 @@ ________
 * Masking tools for the brain and spinal cord
 * Supports the creation and usage of a variety of coil profiles: Spherical harmonics, shim-only arrays, "AC-DC" multi-coil, etc.
 * Supports different shimming scenarios: dynamic (slicewise), realtime (shim modulation with respiration), gradient z-shimming, two-region shimming (e.g., fat and brain)
+* Supports static B1+ shimming scenarios
```
jcohenadad:
```suggestion
* RF shimming for parallel transmit systems (a.k.a. B1+ shimming)
```

README.rst
```diff
@@ -48,6 +48,7 @@ ________
 * Masking tools for the brain and spinal cord
 * Supports the creation and usage of a variety of coil profiles: Spherical harmonics, shim-only arrays, "AC-DC" multi-coil, etc.
```
jcohenadad:
```suggestion
* Create and use a variety of B0 shimming coil profiles: Spherical harmonics, shim-only arrays, "AC-DC" multi-coil, etc.
```

README.rst
```diff
@@ -48,6 +48,7 @@ ________
...
 * Supports different shimming scenarios: dynamic (slicewise), realtime (shim modulation with respiration), gradient z-shimming, two-region shimming (e.g., fat and brain)
```
jcohenadad:
```suggestion
* Supports different B0 shimming scenarios: dynamic (slicewise), realtime (shim modulation with respiration), gradient z-shimming, two-region shimming (e.g., fat and brain)
```

examples/README.rst
```diff
@@ -13,3 +13,14 @@ series (magnitude and phase images) and respiratory trace information obtained f
...
+demo_b1_shimming.sh
+--------------------------
+
+This  `shell script <https://github.com/shimming-toolbox/shimming-toolbox/blob/master/examples/demo_b1_shimming.sh>`__
+calls different shimming-toolbox command lines functions to perform a B1+ shimming scenario. Text files, scaled B1 maps
+and results figures are generated in the `testing-data` folder.
+
+This script shows how to perform a B1+ shimming experiment using the command lines functions. In this example, a
+magnitude least square algorithm is used to target a B1+ value specified by the user. This scenario assumes that the
+B1 maps have already been converted into NIfTI files.
```
jcohenadad: could you add some figures to show what the user is expected to see? this feel a bit abstracts and it is unclear what files users need, what sequence, etc. A lot of background is missing here to be able to understand how to use this script.

examples/README.rst
```diff
@@ -13,3 +13,14 @@ series (magnitude and phase images) and respiratory trace information obtained f
 multi-gradient echo (MGRE) magnitude image is used to generate an ROI and resample the static and dynamic Gx, Gy, Gz
 component maps to match the MGRE image. Lastly the average Gx, Gy, Gz values within the ROI are computed for each
 slice.
+
```
jcohenadad, line 16: Rename this file to getting_started --> do this in another PR, pls open an issue

examples/README.rst
```diff
@@ -13,3 +13,14 @@ series (magnitude and phase images) and respiratory trace information obtained f
...
+demo_b1_shimming.sh
+--------------------------
```
jcohenadad: fix numebr of dash

examples/README.rst
```diff
@@ -13,3 +13,14 @@ series (magnitude and phase images) and respiratory trace information obtained f
...
+
+This  `shell script <https://github.com/shimming-toolbox/shimming-toolbox/blob/master/examples/demo_b1_shimming.sh>`__
```
jcohenadad, line 20: link to script should be relative so it works for branches

examples/README.rst
```diff
@@ -13,3 +13,14 @@ series (magnitude and phase images) and respiratory trace information obtained f
 multi-gradient echo (MGRE) magnitude image is used to generate an ROI and resample the static and dynamic Gx, Gy, Gz
 component maps to match the MGRE image. Lastly the average Gx, Gy, Gz values within the ROI are computed for each
 slice.
+
+demo_b1_shimming.sh
```
jcohenadad: rename for something more intuitive-- eg: B1 shimming

examples/README.rst
jcohenadad:
```suggestion
calls different command lines functions to perform static B1+ shimming. Text files, scaled B1 maps
and result figures are generated in the `testing-data` folder.

In this example, a magnitude least square algorithm is used to target a B1+ value specified by the user. This scenario assumes that the B1 maps have already been converted into NIfTI files.
```

shimmingtoolbox/cli/b1shim.py
```diff
@@ -15,28 +16,24 @@
...
-              help="Number from 1 to 3 specifying which algorithm to use for B1 optimization.\n"
-                   "1 - Optimization aiming to reduce the coefficient of variation (CoV) of the resulting B1+ field.\n"
-                   "2 - Magnitude least square optimization targeting a specific B1+ value. Target value required.\n"
-                   "3 - Maximizes the minimum B1+ value for better efficiency.")
-@click.option('--target', 'target', type=float, required=False, help="Target B1+ value used by algorithm 2 in nT/V.")
+              help="Number specifying the B1+ shimming algorithm:\n"
+                   "1 - Reduce B1+ coefficient of variation.\n"
+                   "2 - Target a specified B1+ value. Target value required.\n"
+                   "3 - Maximize minimum B1+ for higher signal.")
+@click.option('--target', 'target', type=float, required=False, help="B1+ value (nT/V) targeted by algorithm 2.")
```
po09i, line 28: Type should either be click.int or click.choice. I usually go for 'choice' when you have specific options the user needs to input.

gaspardcereza: In the [click documentation](https://click.palletsprojects.com/en/8.0.x/options/) they use type=float or type=int in their examples. I don't understand why we should avoid it here @po09i

po09i: Maybe I was not clear sorry. We can use `float` when it makes sense. However, in this case, we are choosing between 3 options (`choice` makes more sense since it restricts what the user can input and we don't need to error check, click does it). `Int` also makes more sense than `float` but we need to manually do the error checking.

gaspardcereza: So you're referring to the `algo` option not the `target` one right ? In this case it makes more sense

po09i: Oups! yes I'm referring to algo 😅

examples/README.rst
gaspardcereza: I tried to use relative paths in a .rst file but it did not work... Maybe I missed something but from what I understand, Sphinx builds the doc from the GitHub repo and puts it under the `readthedocs.org` server. So by using relative paths, one can only point to html pages that have been added to this RTD server and not to files located on GitHub... 
Maybe a solution would be to fetch the branch name during the documentation build and insert it in the .rst file links. 
@kousu @joshuacwnewton do you have any thoughts ?

examples/README.rst
gaspardcereza: Will do in #326

examples/README.rst
joshuacwnewton: A couple of notes here:

* The following syntax:
   ```   
   ` <>`_
   ```
   Is meant for [external hyperlinks only](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#hyperlinks), i.e. weblinks to be opened in the browser. So, directly changing this to a relative link wouldn't work as-is -- we would need to change the syntax, too.
* But, the internal link syntax (e.g. [`:ref:`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#internal-links)) won't work either, for the reasons that @gaspardcereza highlighted -- the internal links are meant to point to documentation pages (i.e. reST/`.rst` files) that get converted to html files.
* We could still use the [`:download:`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-download) role! But, in that case, the functionality would be different (clicking the link would download the script, rather than opening a page on GitHub).
* So, it depends on the purpose of the link... 

Just to check, though: Why do we want the link to work for branches anyway? Isn't this link only going to be viewed by the user on master? (In which case, I would think the original solution is probably OK.)

po09i (approved):
Looks good!

examples/README.rst
gaspardcereza: The idea behind having the links pointing to the branch was mainly for review purpose but once merged it does not matter anymore so I think the `:download:` solution would be a bit overkill here

Issue #322, opened 2021-12-09
Figuring out gradient coordinate system (GCS) from NIfTIs


po09i:
### Context
Siemens uses multiple coordinate systems to acquire images. From the Device Coordinate System, (DCS) to the Shim coordinate system (ShimCS) to the Gradient Coordinate System (GCS). Some conventions are used to "facilitate" the understanding of coordinate systems.

They are defined by their orientation and origin. Nibabel describes in more details the different conventions [here](https://nipy.org/nibabel/coordinate_systems.html#naming-reference-spaces) but the gist of it is that the 3d space is defined by either "R" or "L" to designate the positive end of the Right-Left axis, "A" or "P" to designate the positive end of the Anterior-Posterior axis and "I" or "S" to designate the positive end of the Inferior to posterior axis. We can use this to describe the different coordinate systems.

For example, syngo MR E11C describes the following coordinate systems:
* DCS: LAI
* ShimCS: LAI

Software conventions also have coordinate systems:
* Dicom: LPS
* NIfTI: RAS

Other coordinate systems use other axis since they don't represent coordinates in the real world. The gradient coordinate system is one of them. It uses the Frequency Encode direction (ReadOut, RO, FE), the Phase Encode direction (PE) and Slice Encode direction (SL).

### Goal
Our realtime-shim experiment requires to output the coefficients in the Gradient Coordinate System (GCS). There does not seem to be any real way to figure out the Gradient coordinate system. If you find a way let me know! 

So for the moment, we applied a gradient in a direction (x, y or z in the ShimCS), acquired a field map and a dummy anat to figure out the anat orientation and tried to shim. We acquired in different anat orientations without any angle (TRA, COR, SAG, AP/PA, RL/LR...). From this we can find the orientation of the GCS for the different orientations.

### Results
<img width="462" alt="Screen Shot 2021-12-09 at 5 44 49 PM" src="https://user-images.githubusercontent.com/47249340/145487593-f483111c-bf7e-4931-9e50-c58acfb135d0.png">

* Col1: Acquisition number where the raw data can be found
* Col2: Anatomical orientation that we acquired
* Col3: Phase encode direction
* Col4: Sign that the correction coefficient should have to correct for a positive gradient in the ShimCS
* Col5: Axis that the positive gradient was applied in the ShimCS (LAI)
* Col6: Axis that the applied gradient affects in the Gradient CS

### Discussion
[...]

PR #323, opened 2021-12-10
Gradient realtime_shim orientation and nan problems

### LINKED ISSUE #320: Gradient realtime shimming can output nans as coefficients which bricks the sequence
### Overview
The sequence which runs the gradient realtime/static shimming can crash when nans are in the file used to send the coefficients to the scanner. 

### Reproduce the bug
The nans were introduced when we plugged the pmu but did not use it (pressure is constant) so that we have a respiratory signal to use for shimming toolbox. The nans were all in the riro component.

### Potential reason
Since the pmu is constant, the pressure_rms is 0, we then use the pressure_rms to scale the coefficients. While doing so we divide by 0 therefore introducing the nans.

po09i:
## Description
This PR fixes a problem with the output gradient coordinate system. During acquisition acdc_151p we realized that the orientation of the anatomical image changed the expected output of the coefficient. The issue is described in issue #322. There was also an issue which bricked the scanner when there were nans in the output text files.

More specifically, this PR
* adds a function `get_main_orientation()` to figure out the orientation and flips some axis accordingly in `realtime_shim_cli()`.
* prevents the output of nans in the text files by avoiding a division by 0.

### Tests
The PR was tested locally and outputs the correct coefficients for gradient realtime_shim for all 18 different orientations of acdc_144p and acdc_151p anat: ascending, TRA/SAG/COR, PE/-PE

To complete the tests, we should acquire in descending orientation for all TRA/SAG/COR, PE/-PE. Then do some rotations.

## Linked issues
#322 , fixes #320

### DIFF
--- a/test/cli/test_cli_realtime_shim.py
+++ b/test/cli/test_cli_realtime_shim.py
@@ -140,7 +140,7 @@ def test_cli_realtime_shim_sag_anat():
         # Specify output for text file and figures
         path_output = os.path.join(tmp, 'test_realtime_shim')
 
-        # Change json to have an orientation that is SAGITAL
+        # Change json to have an orientation that is SAGITTAL
         nii = nib.load(fname_anat)
         with open(fname_json) as json_file:
             json_data = json.load(json_file)
--- a/test/cli/test_cli_realtime_shim.py
+++ b/test/cli/test_cli_realtime_shim.py
@@ -132,3 +132,59 @@ def test_phase_encode_wrong_tag_value():
         with pytest.raises(ValueError,
                            match="Unexpected value for PhaseEncodingDirection:"):
             _get_phase_encode_direction_sign(fname_nii)
+
+
+def test_cli_realtime_shim_sag_anat():
+    runner = CliRunner()
+    with tempfile.TemporaryDirectory(prefix='st_' + pathlib.Path(__file__).stem) as tmp:
+        # Specify output for text file and figures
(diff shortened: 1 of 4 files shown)


test/cli/test_cli_realtime_shim.py
```diff
@@ -132,3 +132,87 @@ def test_phase_encode_wrong_tag_value():
         with pytest.raises(ValueError,
                            match="Unexpected value for PhaseEncodingDirection:"):
             _get_phase_encode_direction_sign(fname_nii)
+
+
+def test_cli_realtime_shim_sag_anat():
+    runner = CliRunner()
```
po09i, line 139: Add comment

test/cli/test_cli_realtime_shim.py
```diff
@@ -132,3 +132,87 @@ def test_phase_encode_wrong_tag_value():
...
+    with tempfile.TemporaryDirectory(prefix='st_' + pathlib.Path(__file__).stem) as tmp:
+        # Specify output for text file and figures
+        path_output = os.path.join(tmp, 'test_realtime_shim')
+
+        # Change json to have an orientation that is SAGITAL
```
po09i:
```suggestion
        # Change json to have an orientation that is SAGITTAL
```

gaspardcereza (approved):
We reviewed everything together. LGTM !

PR #330, opened 2022-01-26
Correct for possible 2*pi offsets across time points when unwrapping

### LINKED ISSUE #327: Phase unwrapping having 2pi offset across timepoints within a time series
### Bug Description
When taking a time series of phase difference maps of a static phantom, some field maps within the time series sometimes have a constant offset. This should not be the case since the field map should be really similar since the phantom is not moving.

### Reproduction Steps
dataset: acdc144p --> 01_Gx_tra_ap

```
st_dicom_to_nifti
st_prepare_fieldmap --unwrapper prelude
```

During the unwrapping process, a 2pi shift is added to some field maps and not others. These seems to be because there is no way to know where the 0pi is located. It therefore chooses 0s that are different for different field maps.

### Screenshots:
The following screenshots are from the same time series (index 1 and index 2) and we can see a ~400Hz (2pi) shift across the whole image.
Screen Shot 2022-01-24 at 11 20 21 PM Screen Shot 2022-01-24 at 11 20 05 PM

po09i:
## Description
As described in #327, there can be a n*2*pi offset when unwrapping a time series (n is an integer). This PR fixes that by:
- Calculating a shared roi where timepoints x and x+1 are well defined.
- Calculating the mean of both volumes in the roi and looking for n in n*2*pi offset.
- Corrects it by adding the proper offset to timepoints x+1.
- Restarts this algorithm but with the next time point

## Linked issues
Fixes #327

### DIFF
--- a/shimmingtoolbox/prepare_fieldmap.py
+++ b/shimmingtoolbox/prepare_fieldmap.py
@@ -119,7 +119,7 @@ def prepare_fieldmap(list_nii_phase, echo_times, mag, unwrapper='prelude', mask=
 
 
 def correct_2pi_offset(unwrapped, mag, mask, validity_threshold):
-    """ Removes 2*pi offsets from unwrapped for a time series. If there is not offset, it returns the same array.
+    """ Removes 2*pi offsets from `unwrapped` for a time series. If there is no offset, it returns the same array.
 
     Args:
         unwrapped (numpy.ndarray): 4d array of the spatially unwrapped phase
--- a/shimmingtoolbox/prepare_fieldmap.py
+++ b/shimmingtoolbox/prepare_fieldmap.py
@@ -1,15 +1,21 @@
 #!/usr/bin/python3
 # -*- coding: utf-8 -*
 
+import logging
 import math
 import nibabel
 import numpy as np
 from skimage.filters import gaussian
 
 from shimmingtoolbox.unwrap.unwrap_phase import unwrap_phase
+from shimmingtoolbox.masking.threshold import threshold as mask_threshold
+
+logger = logging.getLogger(__name__)
+# Threshold to apply when creating a mask to remove 2*pi offsets
+VALIDITY_THRESHOLD = 0.2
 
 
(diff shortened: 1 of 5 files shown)


shimmingtoolbox/prepare_fieldmap.py
```diff
@@ -101,3 +116,46 @@ def prepare_fieldmap(list_nii_phase, echo_times, unwrapper='prelude', mag=None,
 
     # return fieldmap_hz_gaussian
     return fieldmap_hz
+
+
+def correct_2pi_offset(unwrapped, mag, mask, validity_threshold):
+    """ Removes 2*pi offsets from unwrapped for a time series. If there is not offset, it returns the same array.
```
gaspardcereza: It should be "unwrapper" or "unwrapping" right ?

gaspardcereza: Also "If there is no offset"

po09i: Corrected here: ce1d691

gaspardcereza (approved):
LGTM now !

shimmingtoolbox/prepare_fieldmap.py
```diff
@@ -101,3 +116,46 @@ def prepare_fieldmap(list_nii_phase, echo_times, unwrapper='prelude', mag=None,
...
+
+        # Calculate the means in the same validity mask
+        ma_0 = np.ma.array(unwrapped[..., i_time - 1], mask=validity_mask == False)
+        mean_0 = np.ma.mean(ma_0)
+        ma_1 = np.ma.array(unwrapped[..., i_time], mask=validity_mask == False)
+        mean_1 = np.ma.mean(ma_1)
+
+        # Calculate the number of offset by rounding to the nearest integer.
+        n_offsets_float = (mean_1 - mean_0) / (2 * np.pi)
+        n_offsets = round(n_offsets_float)
```
jcohenadad, line 149: i would suggest not rounding, and testing explicitly in the code if value is between -0.5 and 0.5 (or whatever threshold you choose)

PR #333, opened 2022-01-28
Minor improvements of B1+ shimming functionalities

gaspardcereza:
## Description
This PR aims to add minor improvements to the B1+ shimming workflow.

List of improvements:
- Raise an error when the mask and the B1 maps don't overlap (shimming over nothing)
- Add constraint to prevent CV reduction algorithm to  converge towards null values (perfectly uniform)
- Simplified cost function for algorithm 2 (MLS shimming). Yields the same results, but slightly faster,
- Replaced cost function for algorithm 3 (SAR efficient shimming) by one that yields better results (Q matrices are now a requirement to algo 3, which is normal when willing to evaluate SAR efficiency),
- Replacing 'CoV' by more commonly used 'CV' to refer to the coefficient of variation,
- Add phase only shimming algorithm (`--algo 4`),
- Move B1+ output figure creation from API to CLI level (makes the `b1_shim` API more flexible),
- Remove phase-only shimming result from output figure as it would appear twice when running algo 4,
- Scale colorbar with image size on output figure,
- Add unit test related to new functionalities.


jcohenadad (commented):
i reviewed quickly but it seems fine to me. A more thorough review is required, ideally by Daniel Papp so he can get used to the code ecosystem.

gaspardcereza:
> i reviewed quickly but it seems fine to me. A more thorough review is required, 

I'll wait for @po09i to review as he already has a good overview of what it's there

> ideally by Daniel Papp so he can get used to the code ecosystem.

I'll see with @dpapp86 if he needs help getting familiar with GitHub

shimmingtoolbox/cli/b1shim.py
```diff
@@ -67,15 +72,46 @@ def b1shim_cli(fname_b1_map, fname_mask, algorithm, target, fname_vop, sed, path
...
 
+    # Plot B1+ shimming results
+    b1_shimmed = montage(combine_maps(b1_map, shim_weights))  # RF-shimming result
+    if mask_resampled is not None:
+        b1_shimmed_masked = b1_shimmed*montage(mask_resampled)
+    else:
+        b1_shimmed_masked = b1_shimmed*montage(threshold(b1_map.sum(axis=-1), thr=0))
+
+    b1_shimmed_masked[b1_shimmed_masked == 0] = np.nan  # Replace 0 values by nans for image transparency
+    vmax = np.percentile(b1_shimmed, 99)  # Reduce high values influence on display
```
dpapp86: I am not a fan of fully empirical display window selection. What happens if this window is not correct for the B1 maps?
My alternative suggestion would be to see what values you have within your mask, and set the window to that, with a 20% margin on either side.

shimmingtoolbox/cli/b1shim.py
```diff
@@ -67,15 +72,46 @@ def b1shim_cli(fname_b1_map, fname_mask, algorithm, target, fname_vop, sed, path
...
+    vmax = 5*np.ceil(vmax/5)  # Ceil max range value to next multiple of 5 for good colorbar display
+
+    plt.figure()
+    plt.imshow(b1_shimmed, vmax=vmax, cmap='gray')  # Display background in gray
+    plt.imshow(b1_shimmed_masked, vmin=0, vmax=vmax, cmap="jet")  # Overlay colored shimming ROI
```
dpapp86: Jet is a bad colourmap. The three links below explain it in greater detail, but in essence, the colours on jet have varying luminance. Eg, things can _appear_ brighter on jet than they should be, and visually create differences where there are none. 
https://jakevdp.github.io/blog/2014/10/16/how-bad-is-your-colormap/
https://gorelik.net/2020/08/17/what-is-the-biggest-problem-of-the-jet-and-rainbow-color-maps-and-why-is-it-not-as-evil-as-i-thought/
https://towardsdatascience.com/your-colour-map-is-bad-heres-how-to-fix-it-lessons-learnt-from-the-event-horizon-telescope-b82523f09469
The following documentation gives some good insights. The keyword is "perceptually uniform". Personally, I tend to just go with greyscale. I don't know if python has the "turbo" map from my second link, but that is also a good jet analogue.
https://matplotlib.org/2.0.2/users/colormaps.html

shimmingtoolbox/cli/b1shim.py
```diff
@@ -67,15 +72,46 @@ def b1shim_cli(fname_b1_map, fname_mask, algorithm, target, fname_vop, sed, path
         vop = None
 
     shim_weights = b1shim(b1_map, mask=mask_resampled, algorithm=algorithm, target=target,
```
dpapp86, line 74: I'm not seeing (because I am unfamiliar with the code) why the path_output input has been removed here (line 74), and where the output is generated in the new code. Can you please point me at what I am missing?

dpapp86 (commented):
Looks good to me. Added some minor comments for clarification.

shimmingtoolbox/cli/b1shim.py
gaspardcereza: `path_output` was used here because the b1shim API needed it to output the figure into the specified folder. Now that the figure generation has been moved to from the API level (`shimmingtoolbox/shim/b1shim`) to  the CLI level (`shimmingtoolbox/cli/b1shim`), the API does not require it anymore.

shimmingtoolbox/cli/b1shim.py
gaspardcereza: I first went with jet because it makes slight variations easily distinguishable but your point. As I cannot use 'gray' because it is used for background display, I can go with 'Reds' which is also perceptually uniform (see below).
 
## 'jet'
image

## 'Reds'
image

shimmingtoolbox/cli/b1shim.py
dpapp86: Got it

shimmingtoolbox/cli/b1shim.py
gaspardcereza: 20% is kind of an arbitrary choice that could not suit all B1 maps too...
I'd also like to keep the 0 value as the lower display bound. I'll think about it more

shimmingtoolbox/cli/b1shim.py
gaspardcereza: @po09i that might be of interest to you too as we discussed using jet for your plots

gaspardcereza: Other possibility: use a diverging cmap centered on the mean B1 value: 
image

Pros: slight inhomogeneities are twice easier to visualize because of the 2 colormaps 
Cons: Not the most intuitive way to visualize data (at least in my opinion)

jcohenadad: how about Viridis?

image

jcohenadad: or hot?

image

gaspardcereza: ## 'viridis'

image

gaspardcereza: "hot" is not perfectly sequential: "Others (afmhot, copper, gist_heat, and hot) have kinks in the  functions" 

So I think the problem that we have with 'jet' would also apply here

gaspardcereza: Regarding the "Turbo" colormap, it is not natively included in Matplotlib...

gaspardcereza: I decided to stick to the default Viridis in the end after watching [this](https://www.youtube.com/watch?v=xAoljeRJ3lU).

po09i (approved):
Looks good!

PR #338, opened 2022-02-10
Update st_dicom_to_nifti help

gaspardcereza:
## Description
I find the current help returned by `st_dicom_to_nifti` a bit confusing so I wanted to think of ways to improve it together. 

### Current displayed help

image

### Proposed help

image

## Additional remarks
- I think the `--remove-tmp / dont-remove-tmp` flag is a bit long, we could maybe consider replacing 'remove' by 'rm' as it is a common shortcut among the CLI community.
- Should `dont-remove-tmp` be the default ? I personally think the BIDS structure is already dense enough and I don't really see the point in keeping it but maybe there is a reason to this. 
- Maybe all the [required] flags should be listed first as they are essential 

@po09i let me know what you think.

### DIFF
--- a/shimmingtoolbox/dicom_to_nifti.py
+++ b/shimmingtoolbox/dicom_to_nifti.py
@@ -12,10 +12,14 @@ from shimmingtoolbox import __dir_config_dcm2bids__
 from shimmingtoolbox.utils import create_output_dir
 
 
-def dicom_to_nifti(path_dicom, path_nifti, subject_id='sub-01', path_config_dcm2bids=__dir_config_dcm2bids__, remove_tmp=False):
+def dicom_to_nifti(path_dicom, path_nifti, subject_id='sub-01', path_config_dcm2bids=__dir_config_dcm2bids__,
+                   remove_tmp=False):
     """ Converts dicom files into nifti files by calling dcm2bids
 
     Args:
+        path_config_dcm2bids:
+        subject_id:
+        remove_tmp:
         path_dicom (str): path to the input dicom folder
         path_nifti (str): path to the output nifti folder
 
--- a/shimmingtoolbox/dicom_to_nifti.py
+++ b/shimmingtoolbox/dicom_to_nifti.py
@@ -17,12 +17,11 @@ def dicom_to_nifti(path_dicom, path_nifti, subject_id='sub-01', path_config_dcm2
     """ Converts dicom files into nifti files by calling dcm2bids
 
     Args:
-        path_config_dcm2bids:
-        subject_id:
-        remove_tmp:
-        path_dicom (str): path to the input dicom folder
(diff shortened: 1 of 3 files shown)


po09i (commented):
I like your changes!

> I think the --remove-tmp / dont-remove-tmp flag is a bit long, we could maybe consider replacing 'remove' by 'rm' as it is a common shortcut among the CLI community.

I agree

> Should dont-remove-tmp be the default ? I personally think the BIDS structure is already dense enough and I don't really see the point in keeping it but maybe there is a reason to this.

I think the default is fine, it allows to keep the NIfTI files of things not included in the config file.

shimmingtoolbox/dicom_to_nifti.py
```diff
@@ -12,10 +12,14 @@
...
 
-def dicom_to_nifti(path_dicom, path_nifti, subject_id='sub-01', path_config_dcm2bids=__dir_config_dcm2bids__, remove_tmp=False):
+def dicom_to_nifti(path_dicom, path_nifti, subject_id='sub-01', path_config_dcm2bids=__dir_config_dcm2bids__,
+                   remove_tmp=False):
     """ Converts dicom files into nifti files by calling dcm2bids
 
     Args:
+        path_config_dcm2bids:
+        subject_id:
+        remove_tmp:
```
po09i: Those are empty

test/cli/test_cli_dicom_to_nifti.py
```diff
@@ -19,8 +19,7 @@ def test_cli_dicom_to_nifti(test_dcm2niix_installation):
         path_dicoms = os.path.join(__dir_testing__, 'dicom_unsorted')
         path_nifti = os.path.join(tmp, 'nifti')
         subject_id = 'sub-test'
-        result = runner.invoke(dicom_to_nifti_cli, ['--input', path_dicoms, '--output', path_nifti,
-                                                    '--subject', subject_id])
+        result = runner.invoke(dicom_to_nifti_cli, ['-i', path_dicoms, '-o', path_nifti, '--subject', subject_id])
```
po09i: We should add the `catch_exception=False` option

gaspardcereza: From what I understand from the [Click documentation](https://click.palletsprojects.com/en/8.0.x/api/), not catching the exceptions would make our test less robust so I would leave it this way.

shimmingtoolbox/dicom_to_nifti.py
gaspardcereza: Filled in [19307ad](https://github.com/shimming-toolbox/shimming-toolbox/pull/338/commits/19307add8b95c6c8e469e84b90fad94aafc11958)

po09i (approved):
Looks good to me!

PR #340, opened 2022-02-15
Implement order 2 in st_b0shim static and realtime CLI

po09i:
## Description
Implement the 2nd order so that it outputs in the appropriate coordinate system. 

More specifically, when selecting to output in the shim coordinate system, the output now converts to the shimCS of the manufacturer (only Siemens is implemented right now) for the 2nd order as well. Siemens ShimCS is LAI therefore a change from RAS to LAI is required. This means inverting X and Z for the first order and inverting  ZY and XY for the 2nd order (Z2 is double inverted and X2-Y2 is double inverted).

The bounds are also converted. This means that we input the bounds of the manufacturer in their original CS (ShimCS for Siemens) in the config file.


gaspardcereza (approved):
LGTM

PR #344, opened 2022-02-22
Fix B1+ maps right after their conversion into NIfTI

gaspardcereza:
## Description
The B1+ maps directly output by `dcm2nnix` are bogus due to an error in the DICOM tag used to build the affine matrix. They are also shuffled due to the absence of DICOM tags indicating which B1+ volume corresponds to which Tx element. This made it impossible to visualize them in FSLeyes.

These maps used to be corrected at the moment they were loaded by the B1+ shimming CLI.
This PR moves the correction step so that it is now performed right after the `dcm2bids` conversion (when calling `st_dicom_to_nifti`).  This way, the user can open the B1+ maps in FSLeyes and load them from the overlay when using the GUI.


gaspardcereza:
@po09i there's a lot of changes in this PR so maybe we should review it together during tomorrow's coding session. But it has to be merged for the GUI to behave as intended.

shimmingtoolbox/cli/b1shim.py
```diff
@@ -19,9 +19,9 @@
 
 
 @click.command(context_settings=CONTEXT_SETTINGS)
-@click.option('--b1map', 'fname_b1_map', required=True, type=click.Path(exists=True),
+@click.option('--b1map', 'path_b1_map', required=True, type=click.Path(exists=True),
```
po09i: If it's a full path it should be `fname_xxxxx`

shimmingtoolbox/cli/b1shim.py
```diff
@@ -40,54 +40,55 @@
...
-    json_b1["ImageComments"] = 'Complex uncombined B1 map (nT/V)'
-    fname_nii_b1 = os.path.join(path_output, 'TB1maps_uncombined.nii.gz')
-    nib.save(nii_b1, fname_nii_b1)
-    file_json_b1 = open(os.path.join(path_output, 'TB1maps_uncombined.json'), mode='w')
-    json.dump(json_b1, file_json_b1)
-    b1_map_combined = b1_map.sum(axis=-1)
-    nii_b1_map_combined = nib.Nifti1Image(b1_map_combined, nii_b1.affine, header=nii_b1.header)
+    # Load B1 map
+    nii_b1 = nib.load(path_b1_map)
+    json_b1 = json.load(open(path_b1_map.split('.nii')[0] + '.json'))
```
po09i: closing the file is good practice, you can use `with` as well

shimmingtoolbox/cli/b1shim.py
```diff
@@ -40,54 +40,55 @@
...
+    b1_map = np.asarray(nii_b1.dataobj)
 
     # Load static anatomical mask
-    if fname_mask is not None:
-        nii_mask = nib.load(fname_mask)
-        mask_resampled = resample_from_to(nii_mask, nii_b1_map_combined, order=1, mode='grid-constant').get_fdata()
+    if path_mask is not None:
+        nii_mask = nib.load(path_mask)
+        # Recombine the Tx B1+ maps to get same dimensions as the mask for mask
```
po09i: Comment

test/test_dicom_to_nifti.py
```diff
@@ -1,14 +1,81 @@
...
+#     "ParallelReductionFactorInPlane": 2,
+#     "PixelBandwidth": 440,
+#     "DwellTime": 1.78e-05,
+#     "PhaseEncodingDirection": "j-",
+#     "ImageOrientationPatientDICOM": [-1, 0, 0, 0, 1, 0],
+#     "ImageOrientationText": "Tra",
+#     "InPlanePhaseEncodingDirectionDICOM": "COL",
+#     "ConversionSoftware": "dcm2niix",
+#     "ConversionSoftwareVersion": "v1.0.20211006"
+# }
```
po09i: Remove

po09i (changes requested):

po09i (approved):
Looks good!

PR #346, opened 2022-02-24
Solve launchers needing to be launched from the ST venv

### LINKED ISSUE #343: dcm2bids not recognized when using st_dicom_to_nifti outside the venv
### Bug Description
dcm2bids is not recognized when using `st_dicom_to_nifti` outside the virtual environment. This is an unexpected behaviour since we want the user to use all the CLIs that don't require external dependencies. dcm2bids is not an external dependency.

### Reproduction Steps
Launch `st_dicom_to_nifti` outside the venv

We updated dcm2bids and dcm2niix to be an internal dependency. It is therefore downloaded in the venv. We previously used subprocess to call the users dcm2bids and dcm2niix but it does not exists outside the venv.

### Screenshots:
Screen Shot 2022-02-17 at 12 50 17 PM

po09i:
## Description
We simplified the install process of Shimming Toolbox and have included launchers so that they can be used outside of the shimming toolbox environment. Those launchers work correctly until they run a subprocess. The issue is described in #343 for dcm2bids. The problem seems to be similar to https://github.com/shimming-toolbox/fsleyes-plugin-shimming-toolbox/issues/21. The problem seems to be that the subprocess seems to have the PATH from when it was launched on the command line rather than being updated to the ST venv in the shebang. To fix this, the PATH is updated in Python and fed to the subprocess call.

## Linked issues
Fixed #343

### DIFF
--- a/shimmingtoolbox/utils.py
+++ b/shimmingtoolbox/utils.py
@@ -7,6 +7,10 @@ import os
 import tqdm
 import subprocess
 import logging
+from pathlib import Path
+
+HOME_DIR = str(Path.home())
+PATH_ST_VENV = f"{HOME_DIR}/shimming-toolbox/python/envs/st_venv/bin"
 
 
 def run_subprocess(cmd):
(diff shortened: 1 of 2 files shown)


gaspardcereza:
@po09i the CI failed because of the following error in `test_dicom_to_nifti`: 
```        
from distutils.file_util import copy_file
    
        if not dry_run and not os.path.isdir(src):
            raise DistutilsFileError(
>                 "cannot copy tree '%s': not a directory" % src)
E           distutils.errors.DistutilsFileError: cannot copy tree 'dummy_path': not a directory

../../../shimming-toolbox/python/envs/st_venv/lib/python3.7/distutils/dir_util.py:124: DistutilsFileError
```

gaspardcereza (approved):
LGTM

shimmingtoolbox/utils.py
```diff
@@ -7,6 +7,10 @@
 import tqdm
 import subprocess
 import logging
+from pathlib import Path
+
+HOME_DIR = str(Path.home())
+PATH_ST_VENV = f"{HOME_DIR}/shimming-toolbox/python/envs/st_venv/bin"
```
jcohenadad, line 13: Consider opening an issue to address users moving ST installation folder (we've seen that many times with our SCT users).

PR #374, opened 2022-04-05
Deal with saturation pulse when B0 shimming

### LINKED ISSUE #373: Set currents to 0 during Fat Saturation pulses
### Feature Description:
During an EPI, there can be a Fat Saturation pulse. The gradients/shim coils need to be switched off during that time to avoid spin history effects.

When B0 shimming with trigger related output files (`chronological`), add a line of 0s before each slice. This requires a trigger before each slice and before the fat saturation pulse.

We should check the tags to know if there is a fat sat pulse in the target image.
- The dicom tag (0018,0022) can contain the information: `FS` for Fat Sat
- BIDS has a JSON sidecar tag called `ScanOptions` that can contain `FS` for Fat Sat

po09i:
## Description
This PR adresses the fat saturation pulse issues that can arise when shim coils are active during a fat sat.

This PR specifically:
- Parses the BIDS json sidecar to know if there were fat saturation pulses
- Outputs shim coefficients of 0s before each slice when using `chronological` (triggers) outputs if fat sat is used
- Adds commas after every shim coefficients for easy transfer to other files
- When using `absolute` for `--output-value-format` option, output the initial coefficients instead of 0s

Note:
- This requires a trigger before every fat sat pulse and every slice and assumes fat sat pulses are before every slice

## Linked issues
Fixes #373

### DIFF
--- a/shimmingtoolbox/cli/b0shim.py
+++ b/shimmingtoolbox/cli/b0shim.py
@@ -82,7 +82,9 @@ def b0shim_cli():
                                       "Use 'slicewise' to output in row 1, 2, 3, etc. the shim coefficients for slice "
                                       "1, 2, 3, etc. Use 'chronological' to output in row 1, 2, 3, etc. the shim value "
                                       "for trigger 1, 2, 3, etc. The trigger is an event sent by the scanner and "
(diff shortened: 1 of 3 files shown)


gaspardcereza (approved):
LGTM!

PR #376, opened 2022-04-05
Resample input masks on the target anat when B0 shimming

### LINKED ISSUE #375: Add ability to input mask with different shape and affine than the anat when b0 shimming
When B0 shimming, masks need to have the same shape and affine as the target anatomical image, it would be great to allow masks to be different for the anat. This would allow to mask on different sequences where the masking could be better than on the anat.

Allow the input of any mask and resample on the target anatomical image.

po09i:
## Description
This PR resamples the input masks on the target anatomical image when using B0 shimming. This allows to select a ROI from any geometry.

In more details, this PR:
- Resamples the input mask in the `shim_sequencer` if the affine and/or shape is different
- Resamples the input static and/or riro mask in the `shim_realtime_pmu_sequencer` if the affine and/or shape is different
- Updates the description of the CLI to reflect the changes
- Updates the relevant tests

## Linked issues
Fixes #375


gaspardcereza (approved):
LGTM!

PR #379, opened 2022-04-06
Share a single conda environment.

kousu:
pst_venv and st_venv were separated due to version conflicts -- we believe
over wxPython. We have lost the git log explaining exactly what they were.

They seem to be compatible now, especially after https://github.com/shimming-toolbox/fsleyes-plugin-shimming-toolbox/pull/35
so we can recombine everything into a single env.

Notice that this switches from using Miniconda-latest to Miniconda-py3.7, so that the base environment has
the version of python we expect. This is safe because this conda install is *just* for shimming-toolbox.

Corresponds to https://github.com/shimming-toolbox/fsleyes-plugin-shimming-toolbox/pull/37/

## Checklist

- [ ] I've given this PR a concise, self-descriptive, and meaningful title
- [ ] I've assigned a reviewer

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -98,10 +98,7 @@ jobs:
 
       - name: Run unit tests
         run: |
-          source $HOME/shimming-toolbox/python/etc/profile.d/conda.sh  # to be able to call conda
-          set +u
-          conda activate st_venv
-          set -u
+          source ~/shimming-toolbox/python/bin/activate  # to be able to call conda
           py.test . -v --cov shimmingtoolbox/ --cov-report term-missing
 
       - name: macOS Shellcheck
--- a/installer/create_venv.sh
+++ b/installer/create_venv.sh
@@ -48,4 +48,4 @@ sed_i 's/^ENABLE_USER_SITE.*$/ENABLE_USER_SITE = False/' "$ST_DIR/$PYTHON_DIR/li
 
 # activate miniconda
 # shellcheck disable=SC1091
-source python/etc/profile.d/conda.sh
+source python/etc/profile.d/conda.sh # useless?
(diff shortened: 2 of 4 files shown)


gaspardcereza:
The CI does not pass @kousu... 

However I tried it locally (WSL2 ubuntu 20.04) and everything seems to work just fine (I manage to perform a whole B1 shimming scenario).
It might be working because I already have a local conda install.

Regarding the installation times:

With `time make install CLEAN=true`:
```bash
real    9m45.297s
user    3m58.676s
sys     0m33.077s
```

With `time make install` (I deleted the conda env in the meantime):
```bash
real    7m48.091s
user    3m52.772s
sys     0m25.148s
```

So it's no drastically faster than when we have multiple venvs, but it's definitely very clean !

kousu:
> The CI does not pass @kousu... 

I think I see the problem. I was going quickly yesterday and skipped touching the workflows. I'll probably be able to fix it quickly.

kousu:
While I'm here, do we know why this is here?

https://github.com/shimming-toolbox/shimming-toolbox/blob/f53d61ab8ded893090809a927e260c1390c3d640/.github/workflows/tests.yml#L26-L28

https://github.com/shimming-toolbox/shimming-toolbox/blob/f53d61ab8ded893090809a927e260c1390c3d640/.github/workflows/tests.yml#L41-L44

It causes all the extra runs:

Screenshot 2022-04-08 at 21-16-52 shimming-toolbox_shimming-toolbox Code for performing real-time shimming using external MRI shim coils

which seems pointless because we specifically use conda to override the user's python with our own choice.

I think in the long run, using `matrix:` like this, doing it like ivadomed does, is a great idea, but it is pointless if we're always using conda.

gaspardcereza:
> which seems pointless because we specifically use conda to override the user's python with our own choice.

What you mean is that we should only test for Python 3.7 ? If so that would make sense to me yes

po09i (approved):
Looks good!

PR #380, opened 2022-04-09
Drop `actions/setup-python`

kousu:
## Checklist

## Description

We always install in `conda`, so CI only actually tests python 3.7, and moreover it's conda's python 3.7, not `actions/setup-python`'s.

Dropping this cuts the testing time by over half, since the "3.8" runs were actually the same as the 3.7 runs, and in both `actions/setup-python` was redundant because `make install` immediately overrode `python`.

If we head in a direction where the code is stable and cross-platform enough that we can ease out of demanding `conda`, then `actions/setup-python` could become useful again.

## Linked issues

https://github.com/shimming-toolbox/shimming-toolbox/pull/379#issuecomment-1093555601

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -98,10 +98,7 @@ jobs:
 
       - name: Run unit tests
         run: |
-          source $HOME/shimming-toolbox/python/etc/profile.d/conda.sh  # to be able to call conda
-          set +u
-          conda activate st_venv
-          set -u
+          source ~/shimming-toolbox/python/bin/activate  # to be able to call conda
           py.test . -v --cov shimmingtoolbox/ --cov-report term-missing
 
       - name: macOS Shellcheck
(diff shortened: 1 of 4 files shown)


kousu:
Weird, the CI tests didn't run? I'm marking this a draft until I can look into it. This depends on #379 being merged first, anyway.

po09i:
PR #403 switched to python 3.9 and removed `actions/setup-python` 3.7 and 3.8 in favour of python 3.9. This can therefore be closed.

kousu:
Thanks for keeping me in the loop @po09i !

You might want to reconsider this. The point wasn't just dropping 3.7/3.8, it was mostly to drop this section:

```
      - name: Set up Python
        uses: actions/setup-python@v2
        with:
          python-version: ${{ matrix.python-version }}
```

which installs python in the testing VM. This is redundant because shimming-toolbox installs python inside of conda (inside the testing VM).

Issue #369, opened 2022-03-23
Packaging, Environments


kousu:
### Description of Problem / Inefficiency:

There are several points that could be improved with the installation process.

#### Short term action tasks:

- [ ] avoid long conda solver times: (PR: https://github.com/shimming-toolbox/fsleyes-plugin-shimming-toolbox/pull/36)
  - maybe figure out if there's a dependency that's causing the problem
  - try to find a conda config file we could tweak to make it faster
- [x] don't downgrade wxPython (PR: https://github.com/shimming-toolbox/fsleyes-plugin-shimming-toolbox/issues/34)
  - solve the bugs with wxPython==4.1.1 
  - The reason for the downgrade has been lost when moving the plugin from [ST repository](https://github.com/shimming-toolbox/shimming-toolbox) to the [plugin repository](https://github.com/shimming-toolbox/fsleyes-plugin-shimming-toolbox)

- [x] drop st_venv and just use the base conda, to save download time and make it the deployed environment cleaner (PRs: https://github.com/shimming-toolbox/fsleyes-plugin-shimming-toolbox/pull/37/, https://github.com/shimming-toolbox/shimming-toolbox/pull/379)
  - problem: need to download an installer with the version of python you want to use
    - https://docs.conda.io/en/latest/miniconda.html#linux-installers
    - https://docs.conda.io/en/latest/miniconda.html#macos-installers
- [x] ditto for pst_venv -> base conda (PRs: ditto, we just did both in one day)
  - needs solving any dependency conflicts
  - there may be very very many

#### Medium term action tasks:

- look into psutil compilation errors at pip install
- drop the Makefile + installer scripts; replace with a pip package(s) -> publish to pypi
  - can't explicitly depend on dcm2niix
  
- don't assume you've installed conda
   - we can keep a helpful user 101 installer script that does `curl conda-installer.sh && sh conda-installer.sh && activate conda && pip install shimming-toolbox && create_launchers()`, but for people with pre-existing conda installs they can choose to avoid the redundancy/conflict with our conda and just use theirs (or just use `pip install --user`)

#### Long term action tasks:

- Write a conda package -> publish to conda-forge
  - can explicitly depend on dcm2niix, and maybe prelude
  -

kousu:
Here's a diagram we drew during the meeting today, mapping out the dependencies we have:

[...]

Issue #381, opened 2022-04-10
Launchers and Envs


kousu:
https://github.com/shimming-toolbox/shimming-toolbox/blob/4a99cec/docs/source/user_section/installation.rst#installation-1 is misleading

It says

> The installer will install FSLeyes and dcm2niix into isolated environments. They will be used by Shimming Toolbox and will not interfere if you already have FSLeyes or dcm2niix installed. If you have previously installed FSLeyes, it will not support the GUI.

It's true that the isolated `dcm2niix` won't interfere with daily use, but that leaves open the reverse problem. Because [`dcm2bids` calls `dcm2niix`](https://github.com/UNFmontreal/Dcm2Bids/blob/7f79ecb7239ae047b72c9e21b1e785859df2298d/dcm2bids/dcm2niix.py#L99-L100) just like that, with no environment prefix (and how could it do otherwise?), it uses just the first dcm2niix in your $PATH. Same with our own code:

https://github.com/shimming-toolbox/shimming-toolbox/blob/b3e0e53c93124e105e735370c40f8c3304358dc4/shimmingtoolbox/cli/check_env.py#L171

This means if you already have `dcm2niix` installed on your system, then _depending on whether you have the conda env activated_, you will use one or the other:

1. If you have the conda env activated, `$ST_DIR/$PYTHON/bin` is early on your PATH, so conda's `dcm2niix` takes precendence
2. If you have it deactived, but you have the launchers installed, then you can run, say, `st_maths` and if that ends up calling `dcm2niix` at some point then the OS's copy takes precedence

which means you might inexplicably get inconsistent behaviour.

And it's the same story for `fsleyes`, which is installed by https://github.com/shimming-toolbox/fsleyes-plugin-shimming-toolbox/'s install script. But that's probably even more finicky because if someone has fsleyes installed and set up the way they like it, they probably don't want to use the containerized `fsleyes`, but there's no way to expose the plugin to the non-containerized version, right? Because fsleyes has to look into the PYTHONPATH it knows about and the user's pre-existing fsleyes doesn't know about our conda's PYTHONPATH.

Related: #369

gaspardcereza:
FSleyes indeed needs to be installed in the same venv as the plugin to be able to open it so I don't really see any better way to do it...

po09i:
[...]

PR #382, opened 2022-04-10
Consolidate install script.

kousu:
## Checklist

- [ ] I've given this PR a concise, self-descriptive, and meaningful title
- [ ] I've assigned a reviewer

## Description

The install process has sprawled out. This re-contains it.

It drops Makefile, because we aren't using it: all the Makefile
really does is call some bash scripts (in fact, the Makefile even
embedded a bash 'if' for idempotency, ignoring Make's primary feature).

Then it combines installer/*.sh into a single script, and strips
away all the redundancy and unused code.

'make install' is now './install-101'
'make install CLEAN=true' is now './install-101 -c'
'make help' is gone; just read install-101 instead.

Then it adds these features:

- use bash's fully strict mode
- download Miniconda3 to a checksummed ~/.cache, like pip does, to speed up reinstalls
- use the same checksumming to handle corrupted/maliciously edited installers
- use curl --fail --retry 3 to handle network errors
- install to ~/.local/shimming-toolbox, to avoid cluttering the user's $HOME
- import https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3606;
  (this code must have been forked from SCT before that PR)
- update conda d'abord to silence its nagging
- make shell rc more idempotent: instead of trusting a comment, check
  the actual value of $PATH.
- write "$HOME" into the shell rc, instead of using the long path
  that $HOME was at install time, meaning the install is portable
  if someone moves their $HOME or changes their username, and the
  shell rc is a bit more legible.

## Linked issues

Towards https://github.com/shimming-toolbox/shimming-toolbox/issues/369#issuecomment-1090809073


kousu:
I want to split this apart into a series of PRs:

- one to package dcm2bids.json properly, and delete it from the install script

   (currently commented out, so tests are probably failing at the moment)

   (we should use [`setuptools_scm`](https://github.com/pypa/setuptools_scm) + [`importlib.metadata.files`](https://docs.python.org/3/library/importlib.metadata.html#distribution-files) or [`importlib.resources.path`](https://docs.python.org/3/library/importlib.html#importlib.resources.path) ((I'm still pretty unclear on which I'm supposed to use. I think [pypa](github.com/pypa/) are themselves unclear on this))  
- one to turn on miniforge (it's faster)
- one to turn on mamba (it's even faster again)
- one to expose launchers by adding `$ENV/bin` directly to `$PATH`
 
   I'm not at all convinced this is a good option, but I want to have a PR for it so we can at least say we tried. It would address explore #381.
- one to put launchers in `~/.local/bin` directly, and drop editing the shell RC entirely

By the end I think we can get the install script down under 100 lines. And once we're there, we can start _discouraging_ it's use because there will be very little it will be doing that people can't do just by typing `pip` directly.

Maybe we can eventually strip it down far enough that we can replace handling the non-python parts with conda by handling the non-python parts with distro-specific package managers:

```
if [[ "$(uname)" == "Darwin" ]]; then
  brew install dcm2niix
elif [[ "$(uname)" == "Linux" ]]; then
  . /etc/os-release
  if [[ "$ID" == "debian" || "$ID_LIKE" == "debian" ]]; then
    sudo apt-get install -y dcm2niix
  elif [[ "$ID" == "centos" ]]; then
    sudo yum install -y dcm2niix
  elif [[ "$ID" == "arch" ]]; then
    pikaur -Syu --noconfirm dcm2niix
  fi
fi

pip install .
```

PR #383, opened 2022-04-19
Reduce memory usage and speed things up when shimming

po09i:
## Description
When shimming with large coil profiles and/or many target slices, we can end up using a large amount of memory (>10Gb). I have isolated the problem to be when we calculate the shimmed profiles. This causes Shimming toolbox to be really slow since it saturates the RAM.

This PR:
- Removes variables that consumed lots of memory but were not used afterwards.
- Adds a decorator to time a function and output it in debug logger mode
- Improves the stability of the optimizer by scaling the output of the `_residual` function


po09i:
Current improvements using coil profiles (128x128x75), fieldmap (127x127x55), target (110x110x55). Here is the peak RAM usage.

Before: 14.5GB

After: 4.13GB

Total time to process 156s:
- Sequencer: 134s
  - processing: 112s
  - Calculating theoretical shim: 22s
- Plot coefs: 20s
- Other: 2s

gaspardcereza (approved):
LGTM. The timeit decorator idea is great !

PR #384, opened 2022-04-21
Add fatsat option in dynamic and realtime dynamic b0shimming

po09i:
## Description
#374 added the ability to parse the JSON sidecar for a fat saturation pulse. It automatically added 0s if it detected a fatsat pulse and if the `--output-file-format-coil` is 'chronological'. In practice, we sometimes want to turn off the feature. This PR adds a `--fatsat` option with the following input values:
- 'auto': Automatically detect a fatsat pulse and add the 0s
- 'yes' : Add a the 0s even if there is no fatsat pulse parsed
- 'no': Do not add 0s even if it does parse a fatsat pulse


gaspardcereza (approved):
LGTM

PR #386, opened 2022-04-25
Improve dcm2bids config file and fieldmap conversion

po09i:
## Description
This changes the dcm2bids config file to use `SequenceName` instead of `SeriesDescription` to find the modality of the image. This allows to target more specifically while also being more useful to different groups that might not have the same 
`SeriesDescription` but would be using the same sequence.

This PR also addresses a bug where phase2.nii.gz images that were phasediff images would not get renamed correctly.

### DIFF
--- a/config/dcm2bids.json
+++ b/config/dcm2bids.json
@@ -299,6 +299,50 @@
             "criteria": {
                 "SequenceName": "*tfl2d1_16"
             }
+        },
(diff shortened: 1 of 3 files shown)


gaspardcereza (approved):
LGTM. I assume that didn't break anything since the coverage is improved 👌

PR #387, opened 2022-05-03
Allow spaces in the CLIs

po09i:
## Description
This PR adds the ability to have spaces in the input paths of the CLIs. The limitation we previously had was when using any subprocess calls, the algorithm to split the call would split at every space therefore the command would not be correct (`cd a b` != `cd "a b"`). This is now fixed by using `shlex` and quotes around arguments/options that can have spaces.

### Note:
When validating the changes, I noticed that dcm2bids does not handle spaces in paths so `st_dicom_to_nifti` can't either. However, they do have a [PR](https://github.com/UNFmontreal/Dcm2Bids/pull/158) about this, so the changes should propagate soon.

### DIFF
--- a/shimmingtoolbox/utils.py
+++ b/shimmingtoolbox/utils.py
@@ -10,6 +10,7 @@ import logging
 from pathlib import Path
 import time
 import functools
+import shlex
 
 logger = logging.getLogger(__name__)
 
@@ -23,14 +24,15 @@ def run_subprocess(cmd):
     Args:
         cmd (string): full command to be run on the command line
     """
-    logging.debug(f'{cmd}')
+    logger.debug(f'{cmd}')
     try:
         env = os.environ.copy()
         # Add ST PATH before the rest of the path so that it takes precedence
         env["PATH"] = PATH_ST_VENV + ":" + env["PATH"]
 
+        command = shlex.split(cmd)
         subprocess.run(
-            cmd.split(' '),
+            command,
             text=True,
             check=True,
             env=env
--- a/shimmingtoolbox/utils.py
+++ b/shimmingtoolbox/utils.py
@@ -10,7 +10,6 @@ import logging
 from pathlib import Path
 import time
 import functools
-import shlex
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


jcohenadad (approved):
not tested, but nothing shocking and @po09i tested successfully

shimmingtoolbox/utils.py
```diff
@@ -23,14 +24,15 @@ def run_subprocess(cmd):
...
         cmd (string): full command to be run on the command line
     """
-    logging.debug(f'{cmd}')
+    logger.debug(f'{cmd}')
     try:
         env = os.environ.copy()
         # Add ST PATH before the rest of the path so that it takes precedence
         env["PATH"] = PATH_ST_VENV + ":" + env["PATH"]
 
+        command = shlex.split(cmd)
```
kousu: dcm2bids actually [removed shlex](https://github.com/UNFmontreal/Dcm2Bids/pull/158/files#diff-269f9a2dff43a28dbadeda57cde6a9be58a1586ff0c4c69d36614c82b9a1dd71L120) to solve the same issue:

```diff
-    return check_output(shlex.split(commandLine))
+    return check_output(commandLine)
```

and replaced passing the command as a string with passing it as a list ([eg](https://github.com/UNFmontreal/Dcm2Bids/pull/158/files#diff-d02dd4a0ecc637d23f2a143ccc5c95aa749a5ec362bfb85133650c1de7e5c158L98))

```diff
-            commandTpl = "dcm2niix {} -o {} {}"
-            cmd = commandTpl.format(self.options, self.outputDir, dicomDir)
+            cmd = ['dcm2niix', *shlex.split(self.options),
+                   '-o', self.outputDir, dicomDir]
```

[or](https://github.com/UNFmontreal/Dcm2Bids/pull/158/files#diff-056f9fc95825035e71b1b7f0b37031326139ac5d3578135b82374cd7c4377c99L10)

```diff
-    "defaceTpl": "pydeface --outfile {dstFile} {srcFile}",
+    "defaceTpl": ["pydeface", "--outfile", "dstFile", "srcFile"],
```

(ultimately this is because the underlying C [execve(2)](https://manpages.ubuntu.com/manpages/jammy/en/man2/execve.2.html) call itself takes a list: `char *const argv[]`).

Passing a list is the only really reliable way to spawn subprocesses. This PR as it stands handles files with spaces in their name, but it will break in really confusing ways if a user manages to make a file with a `'` in its name, or maybe if they have a `'` in their username (and thus their `$HOME`).

kousu (changes requested):

shimmingtoolbox/utils.py
po09i: Good point! I'll do the necessary changes

kousu: Excellent work, thank you :)

kousu (commented):
If I click "Comment" does that remove my "Request Changes"?

kousu (approved):
Evidently not :roll_eyes: 

Here's your :heavy_check_mark:

shimmingtoolbox/utils.py
po09i: Fixed here: [539a5b7](https://github.com/shimming-toolbox/shimming-toolbox/pull/387/commits/539a5b7f4942adaf28afe553984f303301abc9fb)

PR #389, opened 2022-05-11
Allow 4d masks and target anatomical when using dynamic shimming

po09i:
## Description
When using dynamic shimming, we previously were shimming using 3d GREs. When shimming EPIs, the target anatomical is usually 4d. This was not supported since we expected 3d target anatomical images. This PR allows to accept 4d target images by averaging over the 4th dimension (note that we could simply used the 1st of the volumes since the target image is simply used to get the slice orientations, the voxel data is not used).

Similarly, the mask created from a 4d target anatomical can also be 4d (i.e. using threshold on an EPI, while not being recommended because of the distortions, is possible). This PR allows to accept 4d masks by computing an average over the 4th dimension of the mask and thresholding to remove noise voxels and create a 3d binary mask.

In summary, this PR 
- Allows to input 3d and 4d masks
- Allows to input 3d and 4d target anatomical images

### DIFF
--- a/shimmingtoolbox/shim/sequencer.py
+++ b/shimmingtoolbox/shim/sequencer.py
@@ -96,7 +96,7 @@ def shim_sequencer(nii_fieldmap, nii_anat, nii_mask_anat, slices, coils: ListCoi
     if anat.ndim == 3:
         pass
     elif anat.ndim == 4:
-        logger.debug("Target anatomical is 4d, taking the average and converting to 3d")
+        logger.info("Target anatomical is 4d, taking the average and converting to 3d")
         anat = np.mean(anat, axis=3)
         nii_anat = nib.Nifti1Image(anat, nii_anat.affine, header=nii_anat.header)
     else:
--- a/shimmingtoolbox/shim/sequencer.py
+++ b/shimmingtoolbox/shim/sequencer.py
@@ -93,12 +93,35 @@ def shim_sequencer(nii_fieldmap, nii_anat, nii_mask_anat, slices, coils: ListCoi
 
     # Make sure anat has the appropriate dimensions
     anat = nii_anat.get_fdata()
-    if anat.ndim != 3:
-        raise ValueError("Anatomical image must be in 3d")
+    if anat.ndim == 3:
+        pass
+    elif anat.ndim == 4:
+        logger.debug("Target anatomical is 4d, taking the average and converting to 3d")
+        anat = np.mean(anat, axis=3)
+        nii_anat = nib.Nifti1Image(anat, nii_anat.affine, header=nii_anat.header)
+    else:
+        raise ValueError("Target anatomical image must be in 3d or 4d")
 
     # Make sure the mask has the appropriate dimensions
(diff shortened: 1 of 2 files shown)


shimmingtoolbox/shim/sequencer.py
```diff
@@ -93,12 +93,35 @@ def shim_sequencer(nii_fieldmap, nii_anat, nii_mask_anat, slices, coils: ListCoi
 
     # Make sure anat has the appropriate dimensions
     anat = nii_anat.get_fdata()
-    if anat.ndim != 3:
-        raise ValueError("Anatomical image must be in 3d")
+    if anat.ndim == 3:
+        pass
+    elif anat.ndim == 4:
+        logger.debug("Target anatomical is 4d, taking the average and converting to 3d")
```
jcohenadad:
```suggestion
        logger.info("Target anatomical is 4d, taking the average and converting to 3d")
```

looks like useful info for the user

shimmingtoolbox/shim/sequencer.py
```diff
@@ -93,12 +93,35 @@ def shim_sequencer(nii_fieldmap, nii_anat, nii_mask_anat, slices, coils: ListCoi
...
+    else:
+        raise ValueError("Target anatomical image must be in 3d or 4d")
 
     # Make sure the mask has the appropriate dimensions
-    if nii_mask_anat.get_fdata().ndim != 3:
-        raise ValueError("Mask image must be in 3d")
+    mask = nii_mask_anat.get_fdata()
+    if mask.ndim == 3:
+        pass
+    elif mask.ndim == 4:
```
jcohenadad, line 109: that seems strange to input a 4d mask (why would ppl do that?), and if they do, i'm not sure if averaging across volumes is the solution (eg: why not taking the first mask of the series?).

what were the scenario you had in mind for inputting 4d masks?

po09i: > that seems strange to input a 4d mask (why would ppl do that?)

If the user thresholds an EPI to create a mask, the mask will be 4d. We should probably not recommend it but I was think the user should probably have the flexibility to do so. 

> i'm not sure if averaging across volumes is the solution (eg: why not taking the first mask of the series?)

I'm averaging then thresholding to go back to a binary mask (for now). The idea was to do a logical_and() of all volumes but I found it included dead voxels, adding everything then thresholding allowed to remove those dead voxels.

jcohenadad: > If the user thresholds an EPI to create a mask, the mask will be 4d. We should probably not recommend it but I was think the user should probably have the flexibility to do so.

meh... that's not very common practice. I don't mind keeping the option (as you already implemented it), but i was just curious

shimmingtoolbox/shim/sequencer.py
po09i: Fixed here: [6a1a4f0](https://github.com/shimming-toolbox/shimming-toolbox/pull/389/commits/6a1a4f01047ff06dc303dc281897321e7407c0f4)

jcohenadad (approved):

Issue #329, opened 2022-01-26
CI hanging on different OSs and tests


po09i:
### Summary
Since the merge of #296 with master, the CI hangs on different OSs, and different tests. The tests were passing in that branch when the PR was active. Using the latest master version c146f89 the tests are passing locally on:
- python 3.8 Monterey intel Mac
- python 3.7 Monterey Apple silicon Mac
- python 3.7 Catalina intel Mac
- python 3.7 Windows running WSL

Link to different CI builds:
- #296 merge with master: https://github.com/shimming-toolbox/shimming-toolbox/actions/runs/1748981084/attempts/1
- #296 merge with master manual rerun: https://github.com/shimming-toolbox/shimming-toolbox/actions/runs/1748981084
- #328: https://github.com/shimming-toolbox/shimming-toolbox/actions/runs/1752219059

po09i:
@joshuacwnewton
@jcohenadad was mentioning how SCT have had similar issues recently. Does this ring a bell?

po09i:
The issue seems to be with multiprocessing. When I removed the call to it in `shimmingtoolbox.coils.coordinates:resample_from_to()`, the tests did not hang.

po09i:
This is similar to #278. In that issue, we fixed the problem by using spawn and it has been working fine, this seems to be a different issue. 

Potential fix: ahill187 also mentioned in issue #278 that using `script_runner` would be a good idea to limit the number of spawned processes ("destroys" them after each tests). This might resolve the issue since there are a lot of "closed" but not "destroyed" processes when running py.test. See the issue for more details.

The problem is that script_runner does not allow debugging, CliRunner from Click does but does not support the "destroying" of processes. For now multiprocessing is not necessary and I can't easily debug it so I will remove it as it does not offer much .

PR #390, opened 2022-05-12
Use multiprocessing's fork method to speed up optimizing

kousu:
- avoid python interpreter's startup time
- avoid passing large data via pickle by instead using a global that can be inherited by the subprocesses

Basically this is apply [the advice](https://docs.python.org/3/library/multiprocessing.html#programming-guidelines)

> Better to inherit than pickle/unpickle
>
>    When using the spawn or forkserver start methods many types from multiprocessing need to be picklable so that child processes can use them. However, one should generally avoid sending shared objects to other processes using pipes or queues. Instead you should arrange the program so that a process which needs access to a shared resource created elsewhere can inherit it from an ancestor process.

### DIFF
--- a/shimmingtoolbox/shim/sequencer.py
+++ b/shimmingtoolbox/shim/sequencer.py
@@ -179,7 +179,8 @@ def _eval_static_shim(opt: Optimizer, nii_fieldmap_orig, nii_mask, coef, slices,
             # TODO: Output json sidecar
             # TODO: Update the shim settings if Scanner coil?
             # Output the resulting fieldmap since it can be calculated over the entire fieldmap
-            nii_shimmed_fmap = nib.Nifti1Image(shimmed[..., 0], nii_fieldmap_orig.affine, header=nii_fieldmap_orig.header)
+            nii_shimmed_fmap = nib.Nifti1Image(shimmed[..., 0], nii_fieldmap_orig.affine,
+                                               header=nii_fieldmap_orig.header)
             fname_shimmed_fmap = os.path.join(path_output, 'fieldmap_calculated_shim.nii.gz')
             nib.save(nii_shimmed_fmap, fname_shimmed_fmap)
         else:
@@ -833,28 +834,32 @@ def select_optimizer(method, unshimmed, affine, coils: ListCoil, pmu: PmuResp = 
     return optimizer
 
 
-_optimize_scope = {}  # i?: (_optimizer: Optimizer, nii_mask_anat, slices_anat, shimwise_bounds=None,
-                      #     dilation_kernel='sphere', dilation_size=3, path_output=None)
-def _optimize(_optimizer: Optimizer, nii_mask_anat, slices_anat, shimwise_bounds=None,
-              dilation_kernel='sphere', dilation_size=3, path_output=None):
+# Used as a global dictionary to pass argument to the _opt function. This avoids to pickle the data resulting in a slow
+# start of the multiprocessing pool
+# uuid: (_optimizer: Optimizer, nii_mask_anat, slices_anat, shimwise_bounds=None, dilation_kernel='sphere',
+# dilation_size=3, path_output=None)
+_optimize_scope = {}
 
+
+def _optimize(optimizer: Optimizer, nii_mask_anat, slices_anat, shimwise_bounds=None,
+              dilation_kernel='sphere', dilation_size=3, path_output=None):
     # Count shims to perform
     n_shims = len(slices_anat)
-
-    global _optimize_scope
-    global optimizer
-    optimizer = _optimizer # not a super safe thing to do
 
     # multiprocessing optimization
     mp.set_start_method('fork', force=True)
     run_id = uuid.uuid4().hex
-    _optimize_scope[run_id] = (optimizer, nii_mask_anat, slices_anat, dilation_kernel, dilation_size, path_output, shimwise_bounds)
-    with mp.Pool(mp.cpu_count()) as pool:
-        del _optimize_scope[run_id] # should be safe to del here? because at this point all the child processes have forked and inherited their copy
-        try:
+    global _optimize_scope
+    _optimize_scope[run_id] = (
+        optimizer, nii_mask_anat, slices_anat, dilation_kernel, dilation_size, path_output, shimwise_bounds)
+    try:
+        with mp.Pool(mp.cpu_count()) as pool:
+            # should be safe to del here. Because at this point all the child processes have forked and inherited their
+            # copy
+            del _optimize_scope[run_id]
             results = pool.starmap_async(_opt, [(run_id, i) for i in range(n_shims)]).get()
-        except mp.context.TimeoutError:
-            logger.info("Multiprocessing might have hung, retry the same command")
+    except mp.context.TimeoutError:
+        logger.info("Multiprocessing might have hung, retry the same command")
 
     # TODO: Add a callback to have a progress bar, otherwise the logger will probably output in a messed up order
     results.sort(key=lambda x: x[0])
@@ -864,8 +869,8 @@ def _optimize(_optimizer: Optimizer, nii_mask_anat, slices_anat, shimwise_bounds
 
 
 def _opt(run_id, i):
-
-    (optimizer, nii_mask_anat, slices_anat, dilation_kernel, dilation_size, path_output, shimwise_bounds) = _optimize_scope[run_id]
+    (optimizer, nii_mask_anat, slices_anat, dilation_kernel, dilation_size, path_output, shimwise_bounds) = \
+        _optimize_scope[run_id]
 
     logger.info(f"Shimming shim group: {i + 1} of {len(slices_anat)}")
     # Create nibabel object of the unshimmed map
--- a/shimmingtoolbox/shim/sequencer.py
+++ b/shimmingtoolbox/shim/sequencer.py
@@ -29,6 +29,7 @@ from shimmingtoolbox.shim.shim_utils import calculate_metric_within_mask
 ListCoil = List[Coil]
 
 logger = logging.getLogger(__name__)
+mp.set_start_method('fork', force=True)
 
 supported_optimizers = {
     'least_squares_rt': PmuLsqOptimizer,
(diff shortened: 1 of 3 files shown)


kousu:
Beware: this precludes Windows compatibility with this code. But maybe that's okay.

shimmingtoolbox/shim/sequencer.py
```diff
@@ -830,25 +831,37 @@ def select_optimizer(method, unshimmed, affine, coils: ListCoil, pmu: PmuResp =
...
+    
+    global _optimize_scope
+    global optimizer
+    optimizer = _optimizer # not a super safe thing to do
 
     # multiprocessing optimization
-    mp.set_start_method('spawn', force=True)
+    mp.set_start_method('fork', force=True)
+    run_id = uuid.uuid4().hex
+    _optimize_scope[run_id] = (optimizer, nii_mask_anat, slices_anat, dilation_kernel, dilation_size, path_output, shimwise_bounds)
```
kousu: I wonder if you can use a closure instead of a global:

```suggestion
    def __opt(i):
        return  _opt(i, optimizer, nii_mask_anat, slices_anat, dilation_kernel, dilation_size, path_output, shimwise_bounds)
```

And then below call 

```
         results = pool.starmap_async(__opt, [i for i in range(n_shims)]).get()
```

((and we would want to rename the functions for clarity))

((and undo the other patches in this branch))

By creating a function inside of a function, all the local variables at the moment of its creation get stored into its [closure](https://docs.python.org/3/library/inspect.html#inspect.getclosurevars).

I think that means that at the point of `os.fork()`, `optimizer` and `nii_mask_anat` and all the rest have and will get "zero-copied" into each subprocess, and you don't need `global _optimizer_scope` or `uuid` to disambiguate anything.

But I'm not sure and I haven't tested. `Pool.apply_async` might try to pickle the `func` you pass it. I tried to read through multiprocessing's code and got lost. I think we should just try it on another-other branch and see how it affects the runtime.

kousu:
Ah looks like this hit https://github.com/shimming-toolbox/shimming-toolbox/issues/329

Screenshot 2022-05-12 at 17-41-25 Use multiprocessing's fork method to speed up optimizing by kousu · Pull Request #390 · shimming-toolbox_shimming-toolbox

(360 minutes is 6 hours; it hit the upper limit on Actions runs)

...I wonder if it's related to https://github.com/actions/runner/issues/884 ? 2/3 of the failures in the older thread were macOS only, just like now, though in https://github.com/shimming-toolbox/shimming-toolbox/actions/runs/1748981084/attempts/1 macOS passed but Ubuntu failed so maybe not.

kousu:
Anyway I'd be willing to try to sort this out. I already struggled with pytest in SCT a bunch of times. My impression from that was that it makes a lot of assumptions about the code its testing that aren't always the safest, in the name of making tests nicer to write.

po09i:
> Anyway I'd be willing to try to sort this out.

That would be great! I think having multiprocessing would be a great way to speed things up, I tried to debug this once and got to this [point](https://github.com/shimming-toolbox/shimming-toolbox/issues/329#issuecomment-1022887049). Going further at the time was not worth it, but right now it definitely is. I can try converting everything using multiprocessing to script runner for a start.

shimmingtoolbox/shim/sequencer.py
po09i: I tried it and the function being called by `pool.starmap_async()` needs to be a "top level function". See [here](https://stackoverflow.com/questions/52265120/python-multiprocessing-pool-attributeerror). With that in mind, I don't think this solution works.

po09i: I tried it and the function being called by `pool.starmap_async()` needs to be a "top level function". See [here](https://stackoverflow.com/questions/52265120/python-multiprocessing-pool-attributeerror). With that in mind, I don't think this solution works.

kousu: Ah cool, thanks for actually trying it out! I was going to get around to trying it at some point.

I forgot about the top-level restriction; okay, in that case, I think this weird pseudo-global thing is the best option.

We could maybe attach the values to that top-level function itself:

```
_opt.nii_mask_anat = nii_mask_anat
```

and then so long as the subprocesses never recurse it should be okay, and we can skip `run_id`. But maybe that's more fragile than it's worth.

shimmingtoolbox/shim/sequencer.py
```diff
@@ -833,26 +834,33 @@ def select_optimizer(method, unshimmed, affine, coils: ListCoil, pmu: PmuResp =
...
+              dilation_kernel='sphere', dilation_size=3, path_output=None):
     # Count shims to perform
     n_shims = len(slices_anat)
-    
-    global _optimize_scope
-    global optimizer
-    optimizer = _optimizer # not a super safe thing to do
 
     # multiprocessing optimization
     mp.set_start_method('fork', force=True)
```
kousu: Since this sets a global value (inside mp, but still global) maybe it should be done right after the `import`s, or maybe in the script's `main`.

shimmingtoolbox/shim/sequencer.py
```diff
@@ -833,26 +834,33 @@ def select_optimizer(method, unshimmed, affine, coils: ListCoil, pmu: PmuResp =
...
-    _optimize_scope[run_id] = (optimizer, nii_mask_anat, slices_anat, dilation_kernel, dilation_size, path_output, shimwise_bounds)
-    with mp.Pool(mp.cpu_count()) as pool:
-        del _optimize_scope[run_id] # should be safe to del here? because at this point all the child processes have forked and inherited their copy
-        results = pool.starmap_async(_opt, [(run_id, i) for i in range(n_shims)]).get()
-    
+    global _optimize_scope
+    _optimize_scope[run_id] = (
+        optimizer, nii_mask_anat, slices_anat, dilation_kernel, dilation_size, path_output, shimwise_bounds)
+    try:
+        with mp.Pool(mp.cpu_count()) as pool:
```
kousu: `cpu_count()` is the default:

```suggestion
        with mp.Pool() as pool:
```

shimmingtoolbox/shim/sequencer.py
po09i: I think we should put it at the top of this file, the reason I say this is that the script is not the only thing calling this API, other tests are as well. Otherwise we need to also set it in every test that has mp (sequencer). But maybe that's not too bad.. What do you think?

shimmingtoolbox/shim/sequencer.py
po09i: Resolved here: [2a09aa8](https://github.com/shimming-toolbox/shimming-toolbox/pull/390/commits/2a09aa8f1a44ce36378bbfef174bc3ecca073140)

po09i:
MacOS has failed consistently since we used 'fork', the default has been changed to 'spawn' for MacOS since Python 3.8 for known issues when using 'fork'. This is probably what is happening here. Maybe we can have something like:
```
if macOS:
  mp.set_start_method('spawn', force=True)
else:
  mp.set_start_method('fork', force=True)
```

Problem is that the curent implementation does not work with spawn
Screen Shot 2022-05-18 at 4 42 59 PM

kousu:
> Problem is that the curent implementation does not work with spawn

Oh right, of course. Sad. We can probably get around this by using `if mp.get_start_method() == 'fork'` around the `global` bits? It'll make the code janky but hopefully it'll be worth it for the time savings?


We could also try out 'forkserver'. I don't think the `global` scope hack will work with that either; actually I don't really understand exactly what it does

> When the program starts and selects the forkserver start method, a server process is started. From then on, whenever a new process is needed, the parent process connects to the server and requests that it fork a new process. The fork server process is single threaded so it is safe for it to use [os.fork()](https://docs.python.org/3/library/os.html#os.fork). No unnecessary resources are inherited.
>
>    Available on Unix platforms which support passing file descriptors over Unix pipes.

maybe this means that if you call `mp.Process(target=_opt, (optimizer, ....))` then it wraps `optimizer` up in a way that's efficient to pass to the new subprocess. But I'm not sure how to replicate that with a `Pool()`, because we need to pass a different slice of data to each worker don't we?


---

Actually, as I was writing that I went and read, and I think I see what we missed: [`Pool(..., initializer, initargs, ...)`](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool). 

> If initializer is not None then each worker process will call `initializer(*initargs)` when it starts.

This seems like the API they've provided for the hack I made up on the fly. So instead of using `global _optimize_scope`, we make a small function like

```
def _worker_init(optimizer_, ....):
  global optimizer
  optimizer = optimizer_
   ...
```

and call

```
mp.Pool(initializer=_worker_init, initargs=(optimizer, ....))
```

and use

```
def _opt(run_id):
   < something; we can sketch this in tomorrow >
```

, and then that should be safe no matter whether we use fork/spawn/forkserver. I *think* what it should do is, that if we're using forkserver or fork, the initargs are in scope at the point the forkserver forks, avoiding the expensive pickling step, and if using spawn then they'll still have to be pickled but at least they can be passed with a consistent API.

shimmingtoolbox/shim/sequencer.py
kousu: For what it's worth, python [suggests](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods)

> To select a start method you use the set_start_method() in the if __name__ == '__main__' clause of the main module.

But since the default on Linux is 'fork', the default on macOS is 'spawn' like you [suggested](https://github.com/shimming-toolbox/shimming-toolbox/pull/390#issuecomment-1130515864) so maybe we should just not set it at all, but then we're back to square one, and it'll just be fast on Linux and slow on macOS.

po09i:
> Actually, as I was writing that I went and read, and I think I see what we missed: [Pool(..., initializer, initargs, ...)](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool).

Implemented here: [75a047d](https://github.com/shimming-toolbox/shimming-toolbox/pull/390/commits/75a047d1ceba9dc2d2cff59179cecd515c22bea9)

kousu:
It's green! Dope. But I wonder how much of this time-difference is attributable to using `spawn` on macOS?

Screenshot 2022-05-19 at 15-56-06 Use multiprocessing's fork method to speed up optimizing by kousu · Pull Request #390 · shimming-toolbox_shimming-toolbox


Maybe the macOS machines GitHub offers are just generally slower than Linux ones?

po09i:
The Mac tests have always been slower than linux. If we only look at the **unit tests** comparing the [last CI ](https://github.com/shimming-toolbox/shimming-toolbox/actions/runs/2348204962)that was done on master and the [CI](https://github.com/shimming-toolbox/shimming-toolbox/actions/runs/2353882196) on this last commit: 

- MacOS is ~3-4 minutes slower (7min->10min) to run the CI (Due to lots of small tests and spawn being slow to start, mp helps a lot when using big datasets).
- Linux, tests seem comparable timewise (~4min) to run the CI due to the almost instantaneous forking of processes.

PR #391, opened 2022-05-13
Add option to save the calculated mask when using st_prepare_fieldmap

po09i:
## Description
This PR adds an option in `st_prepare_fieldmap` to output the calculated mask. This is useful to calculate masks that take the fieldmap's validity region into consideration.

In more detail, this PR adds:
- `--savemask` option in `st_prepare_fieldmap`

### DIFF
--- a/shimmingtoolbox/cli/prepare_fieldmap.py
+++ b/shimmingtoolbox/cli/prepare_fieldmap.py
@@ -14,6 +14,7 @@ from shimmingtoolbox.utils import create_fname_from_path, set_all_loggers, creat
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
 FILE_OUTPUT_DEFAULT = 'fieldmap.nii.gz'
+MASK_OUTPUT_DEFAULT = 'mask.nii.gz'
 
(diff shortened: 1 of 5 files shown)


shimmingtoolbox/cli/prepare_fieldmap.py
```diff
@@ -14,6 +14,7 @@
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
 FILE_OUTPUT_DEFAULT = 'fieldmap.nii.gz'
+MASK_OUTPUT_DEFAULT = 'mask.nii.gz'
```
jcohenadad: maybe you want to call it mask_fieldmap to distinguish it from other masks in the pipeline (eg: mask_shim)

jcohenadad (approved):
i left a minor comment, but i’m approving bc it looks good overall

shimmingtoolbox/cli/prepare_fieldmap.py
po09i: Good point, fixed here: [5c8318e](https://github.com/shimming-toolbox/shimming-toolbox/pull/391/commits/5c8318e6ef77cf7c6167aa48dc94f04c57b68de7)

PR #392, opened 2022-05-13
Use script runner instead of CliRunner when using multiprocessing

po09i:
## Description
This PR builds on #278. It uses script_runner instead of CliRunner to run scripts that use multiprocessing.

I verified that the processes were resetting between each tests locally. 

Note:
`script_runner` can launch tests with subprocess or inprocess. Subprocess is what allows the processes to reset between tests. However, we can't use the PyCharm's debug features with subprocess. This is where inprocess comes in, inprocess allows to debug easily. The option can be changed using `@pytest.mark.script_launch_mode('inprocess')` as a decorator of the test being run.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -3,3 +3,4 @@ markers =
     prelude: marks tests dependent on prelude
     dcm2niix: marks tests dependent on dcm2niix
     sct: marks tests dependent on sct
+script_launch_mode = subprocess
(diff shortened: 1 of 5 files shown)


po09i:
I manually triggered the CI and all the tests passed. See [here](https://github.com/shimming-toolbox/shimming-toolbox/actions/runs/2324757758)

4rnaudB:
@po09i Close this PR?

PR #393, opened 2022-05-17
Use --savemask option from prelude to calculate the output mask in st_prepare_fieldmap

po09i:
## Description
Use `--savemask` option from prelude to calculate the output mask in `st_prepare_fieldmap`. This is a follow up to #391 that would assume the mask provided was not changed by prelude. This PR remedies that assumption by using the mask directly calculated by prelude. This PR also outputs a `.json` file for the output mask.


jcohenadad (approved):
i haven't tested but the code LGTM!

PR #394, opened 2022-05-31
Fix bug where there were dead voxels in the calculated shim figures

po09i:
## Description
Fix bug where there were dead voxels in the calculated shim figures when using `st_b0shim dynamic`


gaspardcereza (approved):
LGTM!

PR #395, opened 2022-06-02
Improve mask resampling

po09i:
## Description
When we use st_b0shim, there is a resampling of the sliced mask onto the target space. We also have the option to dilate that resampled sliced mask. When we dilate, it is possible to go outside the ROI, when this happens, we can get unexpected results if the fieldmap is undefined outside of that region. Therefore, when an output sliced mask is dilated, it should still be within the ROI of the full mask.

This PR reflects those changes.


gaspardcereza (approved):
LGTM!

PR #400, opened 2022-07-11
Resolve deprecated st_venv instructions and calls

mathieuboudreau:
## Checklist

## Description

The virtual environment named `st_venv` that used to be created during the installation was removed in PR #379, however installation instructions still mention this. Trying to activate this venv results in an error to new users. The solution on my end was to either do `condo activate base` after sourcing the `conda.sh`, which I'm not really a fan of (it's confusing to me because base usually refers to the OS's base environment), or `condo activate ~/shimming-toolbox/python/`, which is what I implemented since it uses a similar relative path convention as other parts of the instructions.

I also changed one part of the code in the subprocesses where the bin of the `st_venv` was being added to the path; I've changed it to the bin of the base environment now.

## Linked issues

Related to the closed PR #379

Not an issue, but this was first raised in this discussion: https://github.com/shimming-toolbox/shimming-toolbox/discussions/398

### DIFF
--- a/docs/source/user_section/installation.rst
+++ b/docs/source/user_section/installation.rst
@@ -104,7 +104,7 @@ To run the entire testing suite, run ``pytest`` from the
 
   cd ~/shimming-toolbox/shimming-toolbox
   source $HOME/shimming-toolbox/python/etc/profile.d/conda.sh
-  conda activate ~/shimming-toolbox/python/     
+  conda activate ~/shimming-toolbox/python/
   pytest
 
 See https://docs.pytest.org/ for more options.
@@ -165,4 +165,5 @@ You can then activate the ``shimming-toolbox`` environment and start coding!
 
 .. code:: bash
 
+    source $HOME/shimming-toolbox/python/etc/profile.d/conda.sh
     conda activate ~/shimming-toolbox/python/
--- a/docs/source/user_section/installation.rst
+++ b/docs/source/user_section/installation.rst
@@ -104,7 +104,7 @@ To run the entire testing suite, run ``pytest`` from the
 
   cd ~/shimming-toolbox/shimming-toolbox
   source $HOME/shimming-toolbox/python/etc/profile.d/conda.sh
-  conda activate st_venv
+  conda activate ~/shimming-toolbox/python/     
   pytest
 
 See https://docs.pytest.org/ for more options.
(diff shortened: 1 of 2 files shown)


docs/source/user_section/installation.rst
```diff
@@ -104,7 +104,7 @@ To run the entire testing suite, run ``pytest`` from the
 
   cd ~/shimming-toolbox/shimming-toolbox
   source $HOME/shimming-toolbox/python/etc/profile.d/conda.sh
-  conda activate st_venv
+  conda activate ~/shimming-toolbox/python/
```
kousu: `source` should take care of activating the base environment. That's how I was doing it, anyway. So we should just not need this at all.

```suggestion
```

docs/source/user_section/installation.rst
```diff
@@ -124,7 +124,7 @@ To test shimming-toolbox without ``prelude``:
...
```
kousu, line 127:
```suggestion
```

docs/source/user_section/installation.rst
```diff
@@ -165,4 +165,4 @@ You can then activate the ``shimming-toolbox`` environment and start coding!
 
 .. code:: bash
 
-    conda activate st_venv
+    conda activate ~/shimming-toolbox/python/
```
kousu, line 169: I'm just looking over this quickly but I think this should probably be

```suggestion
    source $HOME/shimming-toolbox/python/etc/profile.d/conda.sh
```

shimmingtoolbox/utils.py
```diff
@@ -14,7 +14,7 @@
 logger = logging.getLogger(__name__)
 
 HOME_DIR = str(Path.home())
-PATH_ST_VENV = f"{HOME_DIR}/shimming-toolbox/python/envs/st_venv/bin"
+PATH_ST_VENV = f"{HOME_DIR}/shimming-toolbox/python/bin"
```
kousu, line 17: Huh... this shouldn't need to be here. We should be using `__file__` or [`importlib.resources.files().joinpath()`](https://importlib-resources.readthedocs.io/en/latest/using.html).

But that's something to fix in another PR.

mathieuboudreau:
Thanks @kousu !

From what I recall in my discussion (https://github.com/shimming-toolbox/shimming-toolbox/discussions/398), the venv wasn't being activated after sourcing the conda.sh script. This might be because I'm in a `zsh` environment; I'll try again both in `zsh` and `bash` environments in a new user profile on my Mac and will report back on my experiences.

po09i:
On zsh, I also need to activate the base environment after sourcing, but that can be done with a simple `conda activate`

shimmingtoolbox/utils.py
po09i: Opened an issue here: https://github.com/shimming-toolbox/shimming-toolbox/issues/411

po09i (approved):

PR #402, opened 2022-07-12
Change SCT CLI call to check version

### LINKED ISSUE #401: Failing SCT dependency check test
### Checklist

- [x] I've given this issue a self-descriptive, meaningful title (e.g. *Installation failure: problem creating launchers*)
- [x] I've described my environment
- [x] I've described the bug
- [x] I've described the steps to reproduce the behaviour
- [ ] I've included screenshots

### Environment details

- [x] I followed the [installation instructions](https://shimming-toolbox.org/en/latest/2_getting_started/1_installation.html)

Project version:

Operating System and version:

PRETTY_NAME="Ubuntu 22.04 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04 LTS (Jammy Jellyfish)"

Python version: 3.7.1

Path to Python binary: /path/to/bin/python 

/home/GRAMES.POLYMTL.CA/u097381/shimming-toolbox/python/bin/python

### Bug Description

Currently, all GitHub acion CI runs fail because of the SCT depencency check. This is because SCT just changed the output of their `sct_check_dependencies` output 4 days ago ([see here](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blame/df74db89bcf64ed8cb3a8d0ae8a1f3ef24e8714f/spinalcordtoolbox/scripts/sct_check_dependencies.py#L201-L203)), from this PR: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3820. This results in a different (and unexpected) output of `st_depencency_check` for systems using an older version of SCT:

```

(base) mathieuboudreau@Mathieus-MacBook-Pro shimming-toolbox % st_check_dependencies
Check if prelude is installed.......................[OK]
    Part of FSL (ID: 6.0.4:ddd0a010)
    prelude (Version 2.0 in c# minor)
    Phase Region Expanding Labeller for Unwrapping Discrete Estimates
    Copyright(c) 2001, University of Oxford (Mark Jenkinson)
Check if Spinal Cord Toolbox is installed...........[OK]
    - version: git-mb/3810-mtsat-7d495ca33892ccb20148ec5e6f1b9650c1f3f2fe

```

Vs the current one:

```

(base) u097381@bireli:~/shimming-toolbox/test/cli$ st_check_dependencies
Check if prelude is installed.......................[OK]
    Part of FSL (ID: 5.0.11)
    prelude (Version 2.0 in c# minor)
    Phase Region Expanding Labeller for Unwrapping Discrete Estimates
    Copyright(c) 2001, University of Oxford (Mark Jenkinson)
Check if Spinal Cord Toolbox is installed...........[OK]
    SYSTEM INFORMATION

```

### Reproduction Steps

Open a new PR, re-run tests, or reinstall the latest version of SCT locally.

SCT just changed the output of their `sct_check_dependencies` output 4 days ago ([see here](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blame/df74db89bcf64ed8cb3a8d0ae8a1f3ef24e8714f/spinalcordtoolbox/scripts/sct_check_dependencies.py#L201-L203)), from this PR: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3820.

So, this ST function will need to be updated to consider the latest output of SCT check depenceny, '

https://github.com/shimming-toolbox/shimming-toolbox/blob/4013cae6031e544e5a3ae77e2b623e17a14b90af/shimmingtoolbox/cli/check_env.py#L180-L196

however, care should be taken because that might mean that setups using older SCT versions might fail tests then if not done correctly.

mathieuboudreau:
## Checklist

## Description

SCT recently ([see here](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blame/df74db89bcf64ed8cb3a8d0ae8a1f3ef24e8714f/spinalcordtoolbox/scripts/sct_check_dependencies.py#L201-L203), from this PR: https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/3820) added a line in their `sct_depencency_check` output, resulting in `st_depencency_check` in outputting

```
Check if Spinal Cord Toolbox is installed...........[OK]
    SYSTEM INFORMATION
```

instead of

```
Check if Spinal Cord Toolbox is installed...........[OK]
    - version: git-mb/3810-mtsat-7d495ca33892ccb20148ec5e6f1b9650c1f3f2fe
```

The root cause on our end was that this line:

https://github.com/shimming-toolbox/shimming-toolbox/blob/4013cae6031e544e5a3ae77e2b623e17a14b90af/shimmingtoolbox/cli/check_env.py#L194

where `.split("\n")[1] ` was counting from the top line of the SCT output and trying to fetch `- version ...`, but instead was fetching the new line in their output. My proposed solution that will be retrocompatible (I tested it locally on the old vs new SCT versions) is to fetch the SCT `- version` line from counting from the bottom with `.split("\n")[-5] ` instead.

## Linked issues

Resolves #401

### DIFF
--- a/shimmingtoolbox/cli/check_env.py
+++ b/shimmingtoolbox/cli/check_env.py
@@ -191,7 +191,7 @@ def get_sct_version() -> str:
     if sct_version.returncode != 0:
         raise subprocess.CalledProcessError("Error while getting SCT's version")
     version_output: str = sct_version.stdout.rstrip()
-    version_output = version_output.split("\n\n")[2].split("\n")[1]
+    version_output = version_output.split("\n\n")[2].split("\n")[-5]
 
     return version_output
 
--- a/shimmingtoolbox/cli/check_env.py
+++ b/shimmingtoolbox/cli/check_env.py
@@ -187,11 +187,10 @@ def get_sct_version() -> str:
         str: Version of the ``SCT`` installation.
     """
     # `sct_check_dependencies -short` returns
-    sct_version: str = subprocess.run(["sct_check_dependencies", "-short"], capture_output=True, encoding="utf-8")
+    sct_version: str = subprocess.run(["sct_version", "-short"], capture_output=True, encoding="utf-8")
     if sct_version.returncode != 0:
         raise subprocess.CalledProcessError("Error while getting SCT's version")
     version_output: str = sct_version.stdout.rstrip()
-    version_output = version_output.split("\n\n")[2].split("\n")[-5]
 
     return version_output
 
--- a/test/cli/test_cli_check_env.py
+++ b/test/cli/test_cli_check_env.py
@@ -84,7 +84,7 @@ def test_get_sct_version(test_sct_installation):
     """Checks sct version output for expected structure.
     """
     sct_version_info = st_ce.get_sct_version()
-    version_regex = r"- version *"
+    version_regex = r"git*"
     assert re.search(version_regex, sct_version_info)
 
 


shimmingtoolbox/cli/check_env.py
```diff
@@ -191,7 +191,7 @@ def get_sct_version() -> str:
     if sct_version.returncode != 0:
         raise subprocess.CalledProcessError("Error while getting SCT's version")
     version_output: str = sct_version.stdout.rstrip()
-    version_output = version_output.split("\n\n")[2].split("\n")[1]
+    version_output = version_output.split("\n\n")[2].split("\n")[-5]
```
joshuacwnewton: Oh my goodness, I'm so sorry for the disruption. I had no idea that shimming-toolbox was fetching SCT's version string by parsing the plaintext output of `sct_check_dependencies`.

I would recommend using the `sct_version` command instead, which outputs the information you're looking for on its own:

```console
joshua@tadpole:~$ sct_version
git-mgp/remove-coverage-b3d00edffa43d49a7bac0f9b3ddcc8d570334edd
```

mathieuboudreau: I haven't been a dev here, I'm just a new user, so I didn't know either! I think this won't work out of the box though, because their test seems to be looking for the `-version` flag, 

https://github.com/shimming-toolbox/shimming-toolbox/blob/4013cae6031e544e5a3ae77e2b623e17a14b90af/test/cli/test_cli_check_env.py#L82-L88

but that could be fixed too if the actual devs feel it's a better solution. I don't know if their regex for `- version` is used anywhere else in the code.

gaspardcereza: I agree that using `sct_version` would be way cleaner, especially as we might expect the parsed string to change again in the future (and therefore the [-5] index might not be suited anymore). Is there a chance that `sct_version` could not be available with early versions of sct ? Even if I doubt that people will have very old versions installed this could still be a potential concern...

mathieuboudreau: @gaspardcereza seems like `sct_version` was only introduced in December 2020,

https://github.com/spinalcordtoolbox/spinalcordtoolbox/commits/master/spinalcordtoolbox/scripts/sct_version.py

**However**, it was introduced the same day that `sct_check_dependencies` was introduced,

https://github.com/spinalcordtoolbox/spinalcordtoolbox/commits/master/spinalcordtoolbox/scripts/sct_check_dependencies.py

So any potential SCT early version problem of not having one would be the same as the other.

joshuacwnewton: Ah, the git history there is a bit misleading. In december 2020, the entire `scripts` folder was moved. GitHub does a poor job at displaying history when the file location changes, so it ends the history there.

`sct_version` was introduced May 2020 (https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/2723) while `sct_check_dependencies` has been present since the very first SCT commit (https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/10f9a5d5628669c7a8cdbe565a5b57b50c72b488).

joshuacwnewton: Still, "past SCT versions" will be a problem either way, even with the current proposed changes.

Say we were to make the changes in this PR (changing `[1]` to `[-5`]). In that case, any output for `sct_check_dependencies` commands from past versions of SCT will also no longer work, since they expect `[1]`, I think?

joshuacwnewton: Two suggestions:

* Perhaps you could `try:` using `sct_version`, and if the command isn't present, fall back to checking `sct_check_dependencies`?
* For `sct_check_dependencies`, perhaps you could filter for lines that start with `- version:`, that way it might would work regardless of the position in the output?

gaspardcereza: From what I understand, [-5] should work with present and past versions because the version string has always been located 5 lines from the end of the output. However if in the future some new line is appended to the end of the output, it won't work anymore

joshuacwnewton: Ah, my apologies. Please disregard my comment! You're absolutely right.

mathieuboudreau: So, as I understand it, the consensus is that we should just try resolving this bug using `sct_version`? As I mentioned, both the `sct_version` and `sct_check_depencencies` command were created the same day (even though the behaviours of `sct_check_depencencies` existed elsewhere in SCT in prior versions), so at the very least the changes made in this PR will be retrocompatible up to the earliest version of `sct_check_depencencies`, plus all versions since the recent bug (on our end) was introduced.

I'll make the changes ASAP and push it so that the PR can go ahead and the ST tests stop failing.

mathieuboudreau: I've made the change and changed the title. The only issue now is that the old was that the project was testing this was to look for a line that included the string "version", but our new output doesn't contain that string anymore (this was included in the output of sct_check_depencencies but not sct_version). 

See lines:

https://github.com/shimming-toolbox/shimming-toolbox/blob/4013cae6031e544e5a3ae77e2b623e17a14b90af/test/cli/test_cli_check_env.py#L82-L88

I *think* the proper version output should always start with `git` now, so maybe we could test for that instead? Here is the output for me locally:

```
(Pdb) print(version_output)
git-mb/3810-mtsat-7d495ca33892ccb20148ec5e6f1b9650c1f3f2fe
```

I'll try that and see if the CI tests work on GitHub with it.

mathieuboudreau: Done in [e8d330c](https://github.com/shimming-toolbox/shimming-toolbox/pull/402/commits/e8d330cb562aba48ddef28ea450651b061d8b73c)

joshuacwnewton: > I _think_ the proper version output should always start with `git` now, so maybe we could test for that instead? Here is the output for me locally:

One thing to maybe keep in mind is that `sct_version` is written so that if a stable release version of SCT is installed, it will output just the release string (e.g. `5.6`), rather than a version prefixed with `git-`.

However, since the version string check is only relevant in the context of ST's test suite, I think this shouldn't ever be an issue, given that ST's test suite only ever clones SCT:

https://github.com/shimming-toolbox/shimming-toolbox/blob/4013cae6031e544e5a3ae77e2b623e17a14b90af/.github/workflows/tests.yml#L55-L58

And I think we can safely assume that if the test suite can fetch `git-` version strings, then it can fetch any kind of version string (`git-`, `5.6`, etc.) in production. So, this LGTM!

mathieuboudreau: >So, this LGTM!

Great, thanks! Can you approve so this can be merged?

joshuacwnewton (approved):

PR #405, opened 2022-08-11
Add standard deviation criteria and options to select it in st_b0shim CLI

po09i:
## Description
This PR adds the possibility to optimize the standard deviation specifically and allows to select it with an option in the CLI.

Specifically, this PR:
- Adds a _residual function in the least squares optimizer for the standard deviation
- Adds `--optimizer-criteria` to the options for both dynamic and realtime shimming
- Adds relevant tests


Aurelienpujolm (approved):
It works fine for me !  Good job !

PR #413, opened 2022-10-06
Implement maximization of shim settings based on signal intensity

po09i:
## Description
This PR implements a new shimming algorithm that experimentally maximizes signal intensity based on different shim values that were acquired. This would typically be a 4d EPI where the 4th dimension is the different shim settings. This was proposed in this [article](https://onlinelibrary.wiley.com/doi/10.1002/hbm.26018). 

- input: 4d volume containing the different volumes of different shim settings in the 4th dimension
- mask: 3d mask containing the region to calculate the mean signal intensity
- output: The output contains the index of the volume that maximizes signal intensity in the ROI for each slice


4rnaudB (approved):
Works well

PR #415, opened 2022-11-15
Version/python version updates for pre-commit, sklearn and read the docs

po09i:
## Description
This PR resolves three problems with versions and python versions.

- pip install sklearn is now deprecated and we should use pip install scikit-learn
- We are still using python 3.7 for building the documentation leading to errors (3.7 --> 3.9)
- We are still using python 3.7 for the pre-commit check leading to errors (3.7 --> 3.9)

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v2
         with:
-          python-version: '3.7'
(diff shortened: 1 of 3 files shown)


4rnaudB (commented):
Seems good

4rnaudB:
image

4rnaudB (approved):

po09i:
Did you reinstall ST and are you showing the version from the right environment?

PR #419, opened 2022-11-19
Multi echo field mapping

behrouzvia:
## Checklist

## Description
This PR will be using multi-echo phase inputs to calculate the field map.

### DIFF
--- a/shimmingtoolbox/prepare_fieldmap.py
+++ b/shimmingtoolbox/prepare_fieldmap.py
@@ -121,40 +121,41 @@ def prepare_fieldmap(list_nii_phase, echo_times, mag, unwrapper='prelude', mask=
         fieldmap_hz = fieldmap_rad / (2 * math.pi)  # [Hz]
 
     else:
-        # Calculates three echo field mapp
-        # Run the unwrapper for each phase individually
-        phasediff_unwrapped_1 = unwrap_phase(list_nii_phase[0], unwrapper=unwrapper, mag=mag, mask=mask,
-                                             fname_save_mask=fname_save_mask)
-        phasediff_unwrapped_2 = unwrap_phase(list_nii_phase[1], unwrapper=unwrapper, mag=mag, mask=mask,
-                                             fname_save_mask=fname_save_mask)
-        phasediff_unwrapped_3 = unwrap_phase(list_nii_phase[2], unwrapper=unwrapper, mag=mag, mask=mask,
-                                             fname_save_mask=fname_save_mask)
-
-        unwrapped_phase_1 = nibabel.Nifti1Image(phasediff_unwrapped_1, list_nii_phase[0].affine)
-        unwrapped_phase_2 = nibabel.Nifti1Image(phasediff_unwrapped_2, list_nii_phase[1].affine)
-        unwrapped_phase_3 = nibabel.Nifti1Image(phasediff_unwrapped_3, list_nii_phase[2].affine)
-
-        nibabel.save(unwrapped_phase_1, 'unwrapped_phase_1.nii')
-        nibabel.save(unwrapped_phase_2, 'unwrapped_phase_2.nii')
-        nibabel.save(unwrapped_phase_3, 'unwrapped_phase_3.nii')
-
-        rad_over_time = np.zeros(phase[0].shape)
-        for i in range(phase[2].shape[0]):
-            for j in range(phase[2].shape[1]):
-                for k in range(phase[2].shape[2]):
-                    y1 = phasediff_unwrapped_1[i, j, k]
-                    y2 = phasediff_unwrapped_2[i, j, k]
-                    y3 = phasediff_unwrapped_3[i, j, k]
-                    Y = np.asarray([y1, y2, y3])
+        # Calculates field map based on multi echo phases by running the prelude unwrapper for each phase individually.
+        unwrapped = [unwrap_phase(list_nii_phase[echo_number], unwrapper=unwrapper, mag=mag, mask=mask,
+                                  fname_save_mask=fname_save_mask) for echo_number in range(len(list_nii_phase))]
+        n_echoes = len(list_nii_phase)  # Number of Echoes
+        n_y = unwrapped[0].shape[0]  # Number of voxels along Y direction (rows) in each volume.
+        n_x = unwrapped[0].shape[1]  # Number of voxels along X direction (columns) in each volume.
+        n_z = unwrapped[0].shape[2]  # Number of slices along Z direction in each volume.
+        list_unwrapped_nii = []
+        for echo in range(n_echoes):
+            list_unwrapped_nii.append(unwrapped[echo])
+        # merges all the phases in the 4th dimension
+        merged_unwrapped_phases = np.stack(list_unwrapped_nii, axis=3)
+        # creates a mask based on the given magnitude image
+        new_mask = mask_threshold(mag - mag.min(), VALIDITY_THRESHOLD * (mag.max() - mag.min()))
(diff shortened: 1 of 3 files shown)


shimmingtoolbox/prepare_fieldmap.py
```diff
@@ -97,28 +108,54 @@ def prepare_fieldmap(list_nii_phase, echo_times, mag, unwrapper='prelude', mask=
...
-    fieldmap_hz = fieldmap_rad / (2 * math.pi)  # [Hz]
+        # Divide by echo time
+        fieldmap_rad = phasediff_unwrapped / echo_time_diff  # [rad / s]
+        fieldmap_hz = fieldmap_rad / (2 * math.pi)  # [Hz]
 
+    else:
+        # Calculates field map based on multi echo phases by running the prelude unwrapper for each phase individually.
+        unwrapped = [unwrap_phase(list_nii_phase[echo_number], unwrapper=unwrapper, mag=mag, mask=mask,
+                                  fname_save_mask=fname_save_mask) for echo_number in range(len(list_nii_phase))]
+        n_echoes = len(list_nii_phase)  # Number of Echoes
```
po09i:
```suggestion
        n_echoes = len(list_nii_phase)  # Number of Echoes
        unwrapped = [unwrap_phase(list_nii_phase[echo_number], unwrapper=unwrapper, mag=mag, mask=mask,
                                  fname_save_mask=fname_save_mask) for echo_number in range(n_echoes)]
        
```

shimmingtoolbox/prepare_fieldmap.py
```diff
@@ -97,28 +108,54 @@ def prepare_fieldmap(list_nii_phase, echo_times, mag, unwrapper='prelude', mask=
...
+        n_y = unwrapped[0].shape[0]  # Number of voxels along Y direction (rows) in each volume.
+        n_x = unwrapped[0].shape[1]  # Number of voxels along X direction (columns) in each volume.
+        n_z = unwrapped[0].shape[2]  # Number of slices along Z direction in each volume.
+        list_unwrapped_nii = []
+        for echo in range(n_echoes):
+            list_unwrapped_nii.append(unwrapped[echo])
```
po09i: You can avoid this expansive for loop by using `unwrapped` which already contains all the data you need.
You can write over `unwrapped` (which should probably be renamed to `nii_unwrapped`)

```
nii_unwrappred = np.moveaxis(np.stack(nii_unwrapped, axis=0), 0, 3)
```

shimmingtoolbox/prepare_fieldmap.py
```diff
@@ -97,28 +108,54 @@ def prepare_fieldmap(list_nii_phase, echo_times, mag, unwrapper='prelude', mask=
...
+        # merges all the phases in the 4th dimension
+        merged_unwrapped_phases = np.stack(list_unwrapped_nii, axis=3)
+        # creates a mask based on the given magnitude image
+        new_mask = mask_threshold(mag - mag.min(), VALIDITY_THRESHOLD * (mag.max() - mag.min()))
```
po09i: I believe `mask` already exists so you can use it instead of calculating it again.

shimmingtoolbox/prepare_fieldmap.py
```diff
@@ -97,28 +108,54 @@ def prepare_fieldmap(list_nii_phase, echo_times, mag, unwrapper='prelude', mask=
...
+        mean = 0
+        for i_time in range(n_echoes):  # corrects 2pi offset between phases
+            new_mean = np.mean(merged_unwrapped_phases[..., i_time][new_mask])
+            n_offsets_float = (mean - new_mean) / (2 * np.pi)
+            n_offsets = round(n_offsets_float)
+            if n_offsets >= 1:
+                i_unwrapped = merged_unwrapped_phases[..., i_time] + n_offsets * np.pi * 2
+                merged_unwrapped_phases[..., i_time] = i_unwrapped
+            mean = new_mean
```
po09i: You should use the `correct_2pi_offset()` function which does the same thing that you implemented. The function also deals with other things such as:
- Calculates a mask that's common to both "time points/echoes" so that the calculated mean is reliable for both "time points/echoes".
- Only adds the offset in the mask (This avoids having the background set to n*2pi)
- Provides user feedback if some correction was necessary, or if the correction offset is close to ambiguous.

It also reuses code :)

po09i:
- You should add a test so that the code you are implementing gets tested. It also provides an easy way for reviewers of your PR to test your code. 
- One of the CI tests is failing. This is a test to check that 3 echoes properly fails (You could refactor this test to test your PR). However, this test uses 4d data, if what you implemented does not support 4d inputs, it should fail graciously (write a "if 4d: raise error...")
Screen Shot 2022-11-21 at 3 03 46 PM

shimmingtoolbox/prepare_fieldmap.py
```diff
@@ -97,28 +108,54 @@ def prepare_fieldmap(list_nii_phase, echo_times, mag, unwrapper='prelude', mask=
...
+        rad_over_time = np.zeros((n_y, n_x, n_z))
+        for slice in range(0, n_z):
+            for voxel_x in range(n_x):
+                for voxel_y in range(n_y):
+                    Y = merged_unwrapped_phases[voxel_y, voxel_x, slice, :]
+                    X = np.asarray(echo_times)
+                    reg = linregress(X, Y)
+                    rad_over_time[voxel_y, voxel_x, slice] = reg.slope
```
po09i: Triple `for` loops are usually not optimal for speed, you could time it and consider using `sklearn.linear_model.LinearRegression` which can do multiple regressions at once. Here is an example: https://github.com/shimming-toolbox/shimming-toolbox/blob/4275c51a227e06665f39e3a792cb57b94ee40b6e/shimmingtoolbox/shim/realtime_shim.py#L139

I'm not familiar with `linregress`, maybe it can do multiple regressions at once as well.

po09i (changes requested):

shimmingtoolbox/prepare_fieldmap.py
```diff
@@ -6,6 +6,8 @@
 import nibabel
 import numpy as np
 from skimage.filters import gaussian
+from scipy.stats import linregress
```
po09i:
```suggestion

```

shimmingtoolbox/prepare_fieldmap.py
```diff
@@ -6,6 +6,8 @@
...
+from sklearn import linear_model
```
po09i: You can directly import LinearRegression instead.

shimmingtoolbox/prepare_fieldmap.py
```diff
@@ -137,6 +162,8 @@ def correct_2pi_offset(unwrapped, mag, mask, validity_threshold):
 
     """
     # Create a mask that excludes the noise
+    # if len(np.shape(mag)) == 3:
+    #     mag = np.repeat(mag[..., np.newaxis], len(unwrapped), axis=3)
```
po09i: Is this necessary?

po09i (changes requested):
Nice fixes from the last suggestions. You should "resolve" the ones that you have resolved.
 
There is a test that is not passing, you should look into why and how to fix it :)

shimmingtoolbox/prepare_fieldmap.py
behrouzvia: No, it's not. First I wanted to change the function itself but then decided not to do that.

po09i (approved):
Looks good! Great work!

PR #425, opened 2022-12-12
Add a cli to create a spherical mask

po09i:
## Description
We currently can mask using (box, square, threshold and SCT). I needed to create a spherical mask so I implemented one.


Aurelienpujolm (approved):
It works perfectly fine for me, great work !

PR #423, opened 2022-12-02
Faster B0 shimming 

Aurelienpujolm:
## Checklist

- [x] I've given this PR a concise, self-descriptive, and meaningful title
- [x] I've linked relevant issues in the PR body
- [x] I've applied the relevant labels to this PR
- [x] I've added relevant tests for my contribution
- [x] I've updated the documentation and/or added correct docstrings
- [x] I've assigned a reviewer

## Description
This PR will be trying to make the B0 shimming as fast as possible 
Specially, this PR will 

- Change the function to minimize to a newest one which is faster to minimize (residuals mse)
- Find a way to plot the newest currents coefficients as fast as possible 
- Reduced the processing time of the sequencer 

## Linked issues
This pull request fixes #412 about the time of the B0 shimming

Issue #412, opened 2022-10-04
Speed up B0 shimming


po09i:
### Description of Problem / Inefficiency:
B0 shimming is important but also needs to be fast. 

We currently have 2 main solvers to process and find the currents that best fit B0 maps. Both solvers can be found in the shimmingtoolbox/optimizer.
- [scipy.optimize.minimize](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.minimize.html) (SLSQP)
  - Allows to input constraints, slower
- pseudo inverse
  - No constraints, faster

We would like to make things faster while also respecting coil constraints.

### Implementation Suggestion:
Along with the current implementation, some previous work had been done in Matlab which could be a useful reference: [ST-matlab](https://github.com/shimming-toolbox/shimming-toolbox-matlab/blob/7f0ba2d480a697a16a0d4b20154c61171f7e13ca/Coils/Shim_Template/ShimOpt.m#L940)

### Additional Context / Screenshots:
Initial issue: #20

po09i:
@Aurelienpujolm You can summarize your findings here. I have created a branch for you. You can access it by:
```
git fetch
git checkout ap/faster-solver
```

Aurelienpujolm:
### **Description of problem inefficiency :** 
B0 shimming is important but also needs to be fast. 
For reference, using the django mac computer, shimming with 50 slices, and using 32 channel coil takes around 90 seconds which is too slow. 
Around 20 seconds is taken to get the shimming coefficients, by using the _scipy.optimize.minimize_, 36 seconds to plot figures and around 26 seconds to evaluate the static shim.

### **First suggestion :** 
I first suggest minimizing the time to get the shimming coefficient.

To do this,  I suggest changing the function to minimize from _residual_mae_ (which is using absolute values)  to  an equivalent _residual_mse_. Indeed the residual mse function should always be faster to minimize than the residual mae. Then adding a jacobian in the solver should also reduce the calculation time of _scipy.optimize.minimize._ (For one slice the calculation time can be divided by 10 using mse function and jacobian compare to the mse naive function) 

However, it can be possible to let the user choose which function he wants to minimize even if the optimization gives the same results with both functions.  

In addition, it should be interesting to let the user choose if he wants to use a multiprocessing solver or not. 
[...]

PR #432, opened 2023-01-24
Faster B0 shimming by changing some mathematical expressions

Aurelienpujolm:
## Checklist

## Description
This PR is make to be able to achieve faster B0 shimming, indeed, B0 shimming, is an important process in MRI, and need to be fast. However, before this PR, for a dataset like this : 
- coil ( 32 channels) ;
- anatomic image ([174,174,50] );
-  fieldmap([112,112,70]);
-  mask ([112,112,70]). 
The B0 shimming process took around **130 seconds on the Django Computer**, and with multiprocessing. We found that a big part of this time was caused by the optimizations of the different currents which are calculated in the lsq_optimizer class. After some research I found out that a simple modification can speed up a lot the optimization. 
For this, we just have to change some functions with a switch from np.sum(A*B,...), to A@B (or B@A), indeed in python, it's much more faster to do matrix multiplications than np.sum (And it gave exactly the same result ).
Here is the proof : Let's do this line of code : np.sum(coil_mat * coef, axis = 1 ) (This is used in the residuals_mse function)
with dim(coilmat)=(n,m) and dim(coef)=(m)(Here let's call coil_mat A, and coef B)
Also, it's important to note that in python the multiplication of two matrices is different that the mathematical multiplication of matrices.
Here A multiply by B is equal : 
```math
C= \begin{bmatrix}
 A_{1,1}*B_1,  A_{1,2}*B_2,..., A_{1,m}*B_m \\
...,...,...\\
A_{n,1}*B_1, A_{n,2}*B_2,...,A_{n,m}*B_m
\end{bmatrix}
```
And the np.sum(C, axis = 1 ), will do this :
```math
\begin{bmatrix}
 \\
\sum_{i=1}^{m} A_{1,i}*B_i \\
...\\
\sum_{i=1}^{m}A_{n,i}*B_i\\
\end{bmatrix} 
```
(I don't know why the term in the sum are going below, when I'm writing this on github...)
Which is also just the mathematical multiplication of two matrices !
So I've decided to change some parts of the code with that modification.( I didn't change all of the parts where we have np.sum, because it was already fast enough, and it's more readable for a user to keep np.sum(A*B))
This allows having much more faster optimization. Indeed, on Django mac computer, without multiprocessing it is now taking much less time. Indeed it went from **90 seconds to only 20 seconds** (With around 15 of them caused by resampling). So it does helps ! 
I also did a modification in the function eval_static_shim in the sequencer. to reduce the time by around 10 seconds (While keeping the same results). 

With all of these changes, **B0 shimming should now be faster (80 seconds VS 130)**.  Even with the problem of the high time processing of different resampling, and the fact that multiprocessing on mac computer is still really slow to start.(For this reason, I believe that with these change, it's for now better to not use multiprocessing on mac. However, if we succeed in doing faster multiprocessing, the optimization can be way faster !) 

## Linked issues
Adresses https://github.com/shimming-toolbox/shimming-toolbox/pull/423 , https://github.com/shimming-toolbox/shimming-toolbox/issues/412

### DIFF
--- a/shimmingtoolbox/optimizer/lsq_optimizer.py
+++ b/shimmingtoolbox/optimizer/lsq_optimizer.py
@@ -118,7 +118,7 @@ class LsqOptimizer(Optimizer):
         # Finally np.abs(coef).dot(self.reg_vector) is equivalent and faster to self.reg_factor*np.mean(np.abs(coef) / self.reg_factor_channel)
         # For the mathematical demonstration see : https://github.com/shimming-toolbox/shimming-toolbox/pull/432 
         # MSE regularized to minimize currents
-        shimmed_vec = unshimmed_vec + coil_mat@coef
+        shimmed_vec = unshimmed_vec + coil_mat @ coef
         return (shimmed_vec).dot(shimmed_vec) / len(unshimmed_vec) / factor + np.abs(coef).dot(self.reg_vector)
 
     def _residuals_std(self, coef, unshimmed_vec, coil_mat, factor):
--- a/shimmingtoolbox/optimizer/lsq_optimizer.py
+++ b/shimmingtoolbox/optimizer/lsq_optimizer.py
@@ -90,7 +90,7 @@ class LsqOptimizer(Optimizer):
                             avoid positive directional linesearch
 
         Returns:
-            float: Residuals for least squares optimization -- equivalent to flattened shimmed vector
+            float: Residuals for least squares optimization 
         """
 
         # MAE regularized to minimize currents
@@ -109,7 +109,7 @@ class LsqOptimizer(Optimizer):
                             avoid positive directional linesearch
 
         Returns:
-            float: Residuals for least squares optimization -- equivalent to flattened shimmed vector
+            float: Residuals for least squares optimization 
         """
         # Old one was : np.mean((unshimmed_vec + np.sum(coil_mat * coef, axis=1, keepdims=False))**2) / factor + \
         #                (self.reg_factor * np.mean(np.abs(coef) / self.reg_factor_channel))
@@ -132,7 +132,7 @@ class LsqOptimizer(Optimizer):
                             avoid positive directional linesearch
 
         Returns:
-            float: Residuals for least squares optimization -- equivalent to flattened shimmed vector
+            float: Residuals for least squares optimization 
         """
 
         # STD regularized to minimize currents
--- a/shimmingtoolbox/optimizer/lsq_optimizer.py
+++ b/shimmingtoolbox/optimizer/lsq_optimizer.py
@@ -94,7 +94,7 @@ class LsqOptimizer(Optimizer):
         """
 
         # MAE regularized to minimize currents
-        return np.mean(np.abs(unshimmed_vec + np.sum(coil_mat * coef, axis=1, keepdims=False))) / factor + \
+        return np.mean(np.abs(unshimmed_vec + coil_mat @ coef)) / factor + \
                (self.reg_factor * np.mean(np.abs(coef) / self.reg_factor_channel))
 
     def _residuals_mse(self, coef, unshimmed_vec, coil_mat, factor):
@@ -114,7 +114,7 @@ class LsqOptimizer(Optimizer):
         # Old one was : np.mean((unshimmed_vec + np.sum(coil_mat * coef, axis=1, keepdims=False))**2) / factor + \
         #                (self.reg_factor * np.mean(np.abs(coef) / self.reg_factor_channel))
         # MSE regularized to minimize currents
-        return np.mean((unshimmed_vec + coil_mat@coef) ** 2) / factor + \
+        return np.mean((unshimmed_vec + coil_mat @ coef) ** 2) / factor + \
                (self.reg_factor * np.mean(np.abs(coef) / self.reg_factor_channel))
 
     def _residuals_std(self, coef, unshimmed_vec, coil_mat, factor):
@@ -133,7 +133,7 @@ class LsqOptimizer(Optimizer):
         """
 
         # STD regularized to minimize currents
-        return np.std(unshimmed_vec + np.sum(coil_mat * coef, axis=1, keepdims=False)) / factor + \
+        return np.std(unshimmed_vec + coil_mat @ coef) / factor + \
                (self.reg_factor * np.mean(np.abs(coef) / self.reg_factor_channel))
 
     def _define_scipy_constraints(self):
--- a/shimmingtoolbox/optimizer/lsq_optimizer.py
+++ b/shimmingtoolbox/optimizer/lsq_optimizer.py
@@ -90,7 +90,7 @@ class LsqOptimizer(Optimizer):
                             avoid positive directional linesearch
 
         Returns:
-            float: Residuals for least squares optimization -- equivalent to flattened shimmed vector
+            float: Residuals for least squares optimization 
         """
 
         # MAE regularized to minimize currents
@@ -109,13 +109,16 @@ class LsqOptimizer(Optimizer):
                             avoid positive directional linesearch
 
         Returns:
-            float: Residuals for least squares optimization -- equivalent to flattened shimmed vector
+            float: Residuals for least squares optimization 
         """
         # Old one was : np.mean((unshimmed_vec + np.sum(coil_mat * coef, axis=1, keepdims=False))**2) / factor + \
         #                (self.reg_factor * np.mean(np.abs(coef) / self.reg_factor_channel))
+        # First it's faster to switch np.sum(coil_mat*coef,axis=1,keepdims=False) by coil_mat@coef which is way faster
+        #Then for a vector , mean(x**2) is equivalent to x.dot(x)/n It's faster to do this operation instead of a np.mean
+        #Finally np.abs(coef).dot(self.reg_vector) is equivalent and faster to self.reg_factor*np.mean(np.abs(coef) / self.reg_factor_channel)
+        #For the mathematical demonstration see : https://github.com/shimming-toolbox/shimming-toolbox/pull/432 
         # MSE regularized to minimize currents
-        inner = unshimmed_vec + np.matmul(coil_mat, coef)
-        return inner.dot(inner) / len(inner) / factor + np.abs(coef).dot(self.reg_vector)
+        return (unshimmed_vec + coil_mat@coef).dot(unshimmed_vec + coil_mat @ coef) / len(unshimmed_vec) / factor + np.abs(coef).dot(self.reg_vector)
 
     def _residuals_std(self, coef, unshimmed_vec, coil_mat, factor):
         """ Objective function to minimize the standard deviation (STD)
@@ -129,7 +132,7 @@ class LsqOptimizer(Optimizer):
                             avoid positive directional linesearch
 
         Returns:
-            float: Residuals for least squares optimization -- equivalent to flattened shimmed vector
+            float: Residuals for least squares optimization 
         """
 
         # STD regularized to minimize currents
@@ -236,7 +239,7 @@ class LsqOptimizer(Optimizer):
         Returns:
             jacobian (numpy.ndarray) : 1D array of the gradient of the mse function to minimize
         """
-        return self.b * (unshimmed_vec + np.matmul(coil_mat, coef)) @ coil_mat + np.sign(coef) * self.reg_vector
+        return self.b * (unshimmed_vec + coil_mat @ coef) @ coil_mat + np.sign(coef) * self.reg_vector
 
     def optimize(self, mask):
         """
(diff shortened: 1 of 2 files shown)


shimmingtoolbox/optimizer/lsq_optimizer.py
```diff
@@ -90,11 +90,11 @@ def _residuals_mae(self, coef, unshimmed_vec, coil_mat, factor):
                             avoid positive directional linesearch
 
         Returns:
-            numpy.ndarray: Residuals for least squares optimization -- equivalent to flattened shimmed vector
+            float: Residuals for least squares optimization -- equivalent to flattened shimmed vector
```
po09i:
```suggestion
            float: Residuals for least squares optimization
```

shimmingtoolbox/optimizer/lsq_optimizer.py
```diff
@@ -109,12 +109,13 @@ def _residuals_mse(self, coef, unshimmed_vec, coil_mat, factor):
...
```
po09i:
```suggestion
            float: Residuals for least squares optimization
```

shimmingtoolbox/optimizer/lsq_optimizer.py
```diff
@@ -109,12 +109,13 @@ def _residuals_mse(self, coef, unshimmed_vec, coil_mat, factor):
...
         """
-
+        # Old one was : np.mean((unshimmed_vec + np.sum(coil_mat * coef, axis=1, keepdims=False))**2) / factor + \
+        #                (self.reg_factor * np.mean(np.abs(coef) / self.reg_factor_channel))
         # MSE regularized to minimize currents
-        return np.mean((unshimmed_vec + np.sum(coil_mat * coef, axis=1, keepdims=False)) ** 2) / factor + \
-               (self.reg_factor * np.mean(np.abs(coef) / self.reg_factor_channel))
+        inner = unshimmed_vec + np.matmul(coil_mat, coef)
+        return inner.dot(inner) / len(inner) / factor + np.abs(coef).dot(self.reg_vector)
```
po09i: - Use `@` operator instead of `np.matmul` for consistency within the code.
- Rename inner to something more appropriate
- Does making it a one-liner improve the speed (`len(inner) == len(unshimmed_vec)`)
- Are you saying that` inner.dot(inner) / len(inner)` is equal and faster to `np.mean(inner ** 2)`?
- Can you show how the regularization penalty term that you now have is the same as what was there before

shimmingtoolbox/optimizer/lsq_optimizer.py
```diff
@@ -128,11 +129,11 @@ def _residuals_std(self, coef, unshimmed_vec, coil_mat, factor):
                             avoid positive directional linesearch
 
         Returns:
-            numpy.ndarray: Residuals for least squares optimization -- equivalent to flattened shimmed vector
+            float: Residuals for least squares optimization -- equivalent to flattened shimmed vector
```
po09i:
```suggestion
            float: Residuals for least squares optimization
```

shimmingtoolbox/optimizer/lsq_optimizer.py
```diff
@@ -235,13 +236,7 @@ def _residuals_mse_jacobian(self, coef, unshimmed_vec, coil_mat, factor):
...
             jacobian (numpy.ndarray) : 1D array of the gradient of the mse function to minimize
         """
-        jacobian = np.array([
-            self.b * np.sum((unshimmed_vec + np.sum(coil_mat * coef, axis=1, keepdims=False)) * coil_mat[:, j]) +
-            np.sign(coef[j]) * (self.reg_factor / (9 * self.reg_factor_channel[j]))
-            for j in range(coef.size)
-        ])
-
-        return jacobian
+        return self.b * (unshimmed_vec + np.matmul(coil_mat, coef)) @ coil_mat + np.sign(coef) * self.reg_vector
```
po09i: Same comment for the matmul

shimmingtoolbox/shim/sequencer.py
```diff
@@ -288,8 +286,7 @@ def _cal_shimmed_anat_orient(coefs, coils, nii_mask_anat, nii_fieldmap, slices,
     shimmed_anat_orient = fieldmap_anat
 
     for i_shim in list_shim_slice:
-        # We want to do the np.sum with a 3D matrix if possible
-        corr = np.sum(coefs[i_shim] * coils_anat[:, :, slices[i_shim], :], axis=3, keepdims=False)
+        corr = np.sum(coefs[i_shim] * coils_anat[:,:, slices[i_shim], :], axis=3, keepdims=False)
```
```suggestion
        corr = np.sum(coefs[i_shim] * coils_anat[:, :, slices[i_shim], :], axis=3, keepdims=False)
```

shimmingtoolbox/shim/sequencer.py
```diff
@@ -939,27 +936,24 @@ def select_optimizer(method, unshimmed, affine, coils: ListCoil, opt_criteria, p
 
     return optimizer
 
-
 @timeit
 def _optimize(optimizer: Optimizer, nii_mask_anat, slices_anat, opt_criteria, shimwise_bounds=None,
               dilation_kernel='sphere', dilation_size=3, path_output=None):
-    # Number of optimizations to perform
+    # Count shims to perform
```
po09i: Rollback change

po09i (changes requested):

shimmingtoolbox/optimizer/lsq_optimizer.py
Aurelienpujolm: Done

shimmingtoolbox/optimizer/lsq_optimizer.py
Aurelienpujolm: For the three first points, I changed it. 
Then, for the second point : 
The demonstration that inner.dot(inner) / len(inner) is equal to np.mean(inner**2) can be done with the following mathematical law :  
The scalar product (dot product) of two vectors is equal to the sum of the product of their respectives elements.
Let inner be a vector of dimension n, we have :
inner.dot(inner) = inner[0]*inner[0] + inner[1]*inner[1] + ... + inner[n-1]*inner[n-1]
Then the mean of inner**2 is equal :
np.mean(inner**2) = (inner[0]**2 + inner[1]**2 + ... + inner[n-1]**2) / n
By dividing  inner.dot(inner) by n, we have :
inner.dot(inner) / n = (inner[0]**2 + inner[1]**2 + ... + inner[n-1]**2) / n
This is the proof that inner.dot(inner) / n is equal at np.mean(inner**2).
Then for your last point 
This can be proved by applying the definition of the multiplication of two vectors of size n $n$, and that  $\mathbf{a}$ is the vector $\left| coef \right|$ and $\mathbf{b}$ is the vector $\frac{reg_{factor}}{n reg_{factor_channel}}$ . 

Where $a_i$ and $b_i$ are the corresponding elements of each vector.

By applying the definition of the multiplication of vectors, we have:

$\mathbf{a}\cdot\mathbf{b}=a_{1}b_{1}+a_{2}b_{2}+\cdots+a_{n}b_{n}$

By using the notation of the summation, this can be written like:

$\sum_{i=1}^{n} a_{i}b_{i}$

By switching the expression of $\mathbf{a}$ and $\mathbf{b}$, we have :

$\sum_{i=1}^{n} \left| coef_i \right| \frac{reg_{factor}}{n reg_{factor_channel}}$

And by reorganising, we have :

$\frac{reg_{factor}}{n reg_{factor_channel}} \sum_{i=1}^{n} \left| coef_i \right|$

With the definition of the mean, we can written the sum as:

$n \frac{1}{n} \sum_{i=1}^{n} \left| coef_i \right|$

If we compare this expression with $reg_factor * np.mean(np.abs(coef) / reg_factor_channel)$, we can see that they are equivalent.

Then if you compare the time of the first version and the one that I suggest, you can see that the speed of the function is decreased by a lot.

shimmingtoolbox/optimizer/lsq_optimizer.py
```diff
@@ -109,12 +109,16 @@ def _residuals_mse(self, coef, unshimmed_vec, coil_mat, factor):
                             avoid positive directional linesearch
 
         Returns:
-            numpy.ndarray: Residuals for least squares optimization -- equivalent to flattened shimmed vector
+            float: Residuals for least squares optimization 
         """
-
+        # Old one was : np.mean((unshimmed_vec + np.sum(coil_mat * coef, axis=1, keepdims=False))**2) / factor + \
+        #                (self.reg_factor * np.mean(np.abs(coef) / self.reg_factor_channel))
+        # First it's faster to switch np.sum(coil_mat*coef,axis=1,keepdims=False) by coil_mat@coef which is way faster
```
po09i: Fix comments

shimmingtoolbox/optimizer/lsq_optimizer.py
```diff
@@ -109,12 +109,16 @@ def _residuals_mse(self, coef, unshimmed_vec, coil_mat, factor):
...
+        #Then for a vector , mean(x**2) is equivalent to x.dot(x)/n It's faster to do this operation instead of a np.mean
+        #Finally np.abs(coef).dot(self.reg_vector) is equivalent and faster to self.reg_factor*np.mean(np.abs(coef) / self.reg_factor_channel)
+        #For the mathematical demonstration see : https://github.com/shimming-toolbox/shimming-toolbox/pull/432 
         # MSE regularized to minimize currents
-        return np.mean((unshimmed_vec + np.sum(coil_mat * coef, axis=1, keepdims=False)) ** 2) / factor + \
-               (self.reg_factor * np.mean(np.abs(coef) / self.reg_factor_channel))
+        return (unshimmed_vec + coil_mat@coef).dot(unshimmed_vec + coil_mat @ coef) / len(unshimmed_vec) / factor + np.abs(coef).dot(self.reg_vector)
```
po09i: Rename old inner

shimmingtoolbox/shim/sequencer.py
```diff
@@ -181,7 +183,12 @@ def _eval_static_shim(opt: Optimizer, nii_fieldmap_orig, nii_mask, coef, slices,
         nib.save(nii_merged_coils, os.path.join(path_output, "merged_coils.nii.gz"))
     unshimmed = nii_fieldmap_orig.get_fdata()
     # If the fieldmap was changed (i.e. only 1 slice) we want to evaluate the output on the original fieldmap
-    merged_coils, _ = opt.merge_coils(unshimmed, nii_fieldmap_orig.affine)
+    if equal_fieldmap:
+        # We extend to make sure that merged_coils @ coef[i_shim] has the same dimensions of corrections[...,i_shim]
+        merged_coils = np.expand_dims(opt.merged_coils[:, :, 1, :], 2)
```
po09i: What does this line do? Why are we not just using `opt.merged_coils`

shimmingtoolbox/shim/sequencer.py
```diff
@@ -939,27 +943,24 @@ def select_optimizer(method, unshimmed, affine, coils: ListCoil, opt_criteria, p
...
+    # However, if there is a way to make multiprocessing faster, we should do it.
+    if opt_criteria == 'mse' and sys.platform != 'linux':
+        _worker_init(optimizer, nii_mask_anat, slices_anat, dilation_kernel, dilation_size, path_output,
+                     shimwise_bounds)
+        results = []
+        for i in range(n_shims):
+            result = _opt(i)
+            results.append(result)
+    else:
+    # multiprocessing optimization
```
```suggestion
        # multiprocessing optimization
```

shimmingtoolbox/shim/sequencer.py
Aurelienpujolm: I fixed this in the last commit !

shimmingtoolbox/optimizer/lsq_optimizer.py
```diff
@@ -109,12 +109,17 @@ def _residuals_mse(self, coef, unshimmed_vec, coil_mat, factor):
...
+        # Old one was : np.mean((unshimmed_vec + np.sum(coil_mat * coef, axis=1, keepdims=False))**2) / factor + \
+        #                (self.reg_factor * np.mean(np.abs(coef) / self.reg_factor_channel))
+        # switch np.sum(coil_mat*coef,axis=1,keepdims=False) by coil_mat@coef which is way faster
+        # Then for a vector , mean(x**2) is equivalent to x.dot(x)/n , it's faster to do this operation instead of a np.mean
+        # Finally np.abs(coef).dot(self.reg_vector) is equivalent and faster to self.reg_factor*np.mean(np.abs(coef) / self.reg_factor_channel)
+        # For the mathematical demonstration see : https://github.com/shimming-toolbox/shimming-toolbox/pull/432 
         # MSE regularized to minimize currents
-        return np.mean((unshimmed_vec + np.sum(coil_mat * coef, axis=1, keepdims=False)) ** 2) / factor + \
-               (self.reg_factor * np.mean(np.abs(coef) / self.reg_factor_channel))
+        shimmed_vec = unshimmed_vec + coil_mat@coef
```
po09i:
```suggestion
        shimmed_vec = unshimmed_vec + coil_mat @ coef
```

po09i (approved):

PR #439, opened 2023-03-02
Refactoring of the B0 shimming

Aurelienpujolm:
## Checklist

## Description
This PR is made to refactor the B0 shimming to make it more readable and efficient. 
For this, I've changed the sequencer files in three different classes to make it easier to understand, and to let futur programmers create new sequencer classes by inheriting the Sequencer class. 
This PR, will be followed by two PR to fix the issue of the multiprocessing and the one about resampling.

### DIFF
--- a/shimmingtoolbox/cli/b0shim.py
+++ b/shimmingtoolbox/cli/b0shim.py
@@ -20,8 +20,9 @@ from shimmingtoolbox import __dir_config_scanner_constraints__
 from shimmingtoolbox.cli.realtime_shim import gradient_realtime
 from shimmingtoolbox.coils.coil import Coil, ScannerCoil, convert_to_mp
 from shimmingtoolbox.pmu import PmuResp
-from shimmingtoolbox.shim.sequencer import *
-from shimmingtoolbox.shim.sequencer import shim_max_intensity
+from shimmingtoolbox.shim.sequencer import ShimSequencer, RealTimeSequencer
+from shimmingtoolbox.shim.sequencer import shim_max_intensity,  define_slices
+from shimmingtoolbox.shim.sequencer import extend_fmap_to_kernel_size, parse_slices, new_bounds_from_currents
 from shimmingtoolbox.utils import create_output_dir, set_all_loggers, timeit
 from shimmingtoolbox.shim.shim_utils import phys_to_gradient_cs, phys_to_shim_cs, shim_to_phys_cs
 CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
@@ -258,12 +259,12 @@ def dynamic(fname_fmap, fname_anat, fname_mask_anat, method, opt_criteria, slice
     # Get shimming coefficients
     # 1 ) Create the Shimming sequencer object
     sequencer = ShimSequencer(nii_fmap_orig, nii_anat, nii_mask_anat, list_slices, list_coils,
-                           method=method,
-                           opt_criteria=opt_criteria,
-                           mask_dilation_kernel='sphere',
-                           mask_dilation_kernel_size=dilation_kernel_size,
-                           reg_factor=reg_factor,
-                           path_output=path_output)
+                              method=method,
+                              opt_criteria=opt_criteria,
+                              mask_dilation_kernel='sphere',
+                              mask_dilation_kernel_size=dilation_kernel_size,
+                              reg_factor=reg_factor,
+                              path_output=path_output)
     #2) Launch shim sequencer
     coefs = sequencer.shim_sequencer()
     # Output
--- a/shimmingtoolbox/cli/b0shim.py
+++ b/shimmingtoolbox/cli/b0shim.py
@@ -7,6 +7,7 @@ the gradient method in a st_shim CLI with the argument being:
 - fieldmap_realtime
 - gradient_realtime
 """
+
 import click
 import copy
 import json
@@ -25,6 +26,7 @@ from shimmingtoolbox.shim.sequencer import shim_max_intensity,  define_slices
 from shimmingtoolbox.shim.sequencer import extend_fmap_to_kernel_size, parse_slices, new_bounds_from_currents
 from shimmingtoolbox.utils import create_output_dir, set_all_loggers, timeit
 from shimmingtoolbox.shim.shim_utils import phys_to_gradient_cs, phys_to_shim_cs, shim_to_phys_cs
+
 CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
@@ -1225,6 +1227,7 @@ def max_intensity(fname_input, fname_mask, fname_output, verbose):
         f.write(f"{index_per_slice[n_slices - 1]}")
 
     logger.info(f"Txt file is located here:\n{fname_output}")
+    
 
 b0shim_cli.add_command(gradient_realtime)
 b0shim_cli.add_command(dynamic)
--- a/shimmingtoolbox/cli/b0shim.py
+++ b/shimmingtoolbox/cli/b0shim.py
@@ -265,8 +265,8 @@ def dynamic(fname_fmap, fname_anat, fname_mask_anat, method, opt_criteria, slice
                               mask_dilation_kernel_size=dilation_kernel_size,
                               reg_factor=reg_factor,
                               path_output=path_output)
-    #2) Launch shim sequencer
-    coefs = sequencer.shim_sequencer()
+    # 2) Launch shim sequencer
+    coefs = sequencer.shim()
     # Output
     # Load output options
     options = _load_output_options(json_anat_data, fatsat)
@@ -341,6 +341,8 @@ def dynamic(fname_fmap, fname_anat, fname_mask_anat, method, opt_criteria, slice
 
     logger.info(f"Coil txt file(s) are here:\n{os.linesep.join(list_fname_output)}")
     logger.info(f"Plotting figure(s)")
+    sequencer.eval(coefs)
+    logger.info(f" Plotting currents")
 
     # Plot the coefs after outputting the currents to the text file
     end_channel = 0
@@ -672,16 +674,17 @@ def realtime_dynamic(fname_fmap, fname_anat, fname_mask_anat_static, fname_mask_
     # Load PMU
     pmu = PmuResp(fname_resp)
     # 1 ) Create the real time pmu sequencer object
-    real_time_pmu_sequencer = RealTimeSequencer(nii_fmap_orig, json_fm_data, nii_anat, nii_mask_anat_static, nii_mask_anat_riro,
-                                      list_slices, pmu, list_coils,
-                                      method=method,
-                                      opt_criteria=opt_criteria,
-                                      mask_dilation_kernel='sphere',
-                                      mask_dilation_kernel_size=dilation_kernel_size,
-                                      reg_factor=reg_factor,
-                                      path_output=path_output)
+    sequencer = RealTimeSequencer(nii_fmap_orig, json_fm_data, nii_anat, nii_mask_anat_static,
+                                                nii_mask_anat_riro,
+                                                list_slices, pmu, list_coils,
+                                                method=method,
+                                                opt_criteria=opt_criteria,
+                                                mask_dilation_kernel='sphere',
+                                                mask_dilation_kernel_size=dilation_kernel_size,
+                                                reg_factor=reg_factor,
+                                                path_output=path_output)
     # 2) Launch the sequencer
-    out = real_time_pmu_sequencer.shim_realtime_pmu_sequencer()
+    out = sequencer.shim()
     coefs_static, coefs_riro, mean_p, p_rms = out
 
     # Output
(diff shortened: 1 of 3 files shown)


po09i (commented):
These are really big changes, to facilitate the review, can you go through the changes yourself first. For example, things that I noticed at first glance are:
- Don't import with a "*" to avoid unnecessary imports in the future
- The documentation of arguments has not been updated
- Fix all warnings created by PyCharm

shimmingtoolbox/cli/b0shim.py
```diff
@@ -7,7 +7,6 @@
 - fieldmap_realtime
 - gradient_realtime
 """
-
```
po09i: Add the space back

shimmingtoolbox/cli/b0shim.py
```diff
@@ -21,12 +20,11 @@
...
 from shimmingtoolbox.pmu import PmuResp
-from shimmingtoolbox.shim.sequencer import shim_sequencer, shim_realtime_pmu_sequencer, new_bounds_from_currents
-from shimmingtoolbox.shim.sequencer import define_slices, extend_fmap_to_kernel_size, parse_slices
-from shimmingtoolbox.shim.sequencer import shim_max_intensity
+from shimmingtoolbox.shim.sequencer import ShimSequencer, RealTimeSequencer
+from shimmingtoolbox.shim.sequencer import shim_max_intensity,  define_slices
+from shimmingtoolbox.shim.sequencer import extend_fmap_to_kernel_size, parse_slices, new_bounds_from_currents
 from shimmingtoolbox.utils import create_output_dir, set_all_loggers, timeit
 from shimmingtoolbox.shim.shim_utils import phys_to_gradient_cs, phys_to_shim_cs, shim_to_phys_cs
-
```
po09i, line 29: Add space back

shimmingtoolbox/cli/b0shim.py
```diff
@@ -259,14 +257,16 @@ def dynamic(fname_fmap, fname_anat, fname_mask_anat, method, opt_criteria, slice
...
+    # 1 ) Create the Shimming sequencer object
+    sequencer = ShimSequencer(nii_fmap_orig, nii_anat, nii_mask_anat, list_slices, list_coils,
+                              method=method,
+                              opt_criteria=opt_criteria,
+                              mask_dilation_kernel='sphere',
+                              mask_dilation_kernel_size=dilation_kernel_size,
+                              reg_factor=reg_factor,
+                              path_output=path_output)
+    #2) Launch shim sequencer
+    coefs = sequencer.shim_sequencer()
```
po09i: Rename the method to just "shim" (shim_sequencer --> shim)

shimmingtoolbox/cli/b0shim.py
```diff
@@ -671,16 +671,17 @@ def realtime_dynamic(fname_fmap, fname_anat, fname_mask_anat_static, fname_mask_
...
-                                      opt_method=method,
+                                      method=method,
                                       opt_criteria=opt_criteria,
                                       mask_dilation_kernel='sphere',
                                       mask_dilation_kernel_size=dilation_kernel_size,
                                       reg_factor=reg_factor,
                                       path_output=path_output)
-
+    # 2) Launch the sequencer
+    out = real_time_pmu_sequencer.shim_realtime_pmu_sequencer()
```
po09i: Rename the method to just "shim"

shimmingtoolbox/cli/b0shim.py
```diff
@@ -1221,7 +1222,6 @@ def max_intensity(fname_input, fname_mask, fname_output, verbose):
 
     logger.info(f"Txt file is located here:\n{fname_output}")
 
-
```
po09i, line 1224: Add space back

shimmingtoolbox/shim/sequencer.py
```diff
@@ -42,838 +38,1063 @@
...
-                   path_output=None):
-    """
-    Performs shimming according to slices using one of the supported optimizers and coil profiles.
-
-    Args:
-        nii_fieldmap (nibabel.Nifti1Image): Nibabel object containing fieldmap data in 3d and an affine transformation.
-        nii_anat (nibabel.Nifti1Image): Nibabel object containing anatomical data in 3d.
-        nii_mask_anat (nibabel.Nifti1Image): 3D anat mask used for the optimizer to shim in the region of interest.
+class Sequencer(object):
+    def _select_optimizer(self, unshimmed, affine, pmu: PmuResp = None):
```
po09i: "\_" designate private methods. Since this is now a class, we should make it available in the object. This would allow to change the optimizer instead of recreating a new object. So remove the "_" and make sure that calling it indeed does correctly change the optimizer.

shimmingtoolbox/shim/sequencer.py
```diff
@@ -42,838 +38,1063 @@
...
+        self.opt_criteria = opt_criteria
+        self.mask_dilation_kernel = mask_dilation_kernel
+        self.mask_dilation_kernel_size = mask_dilation_kernel_size
+        self.reg_factor = reg_factor
+        self.path_output = path_output
+        self.optimizer = None
+        self.bounds = None
+        self.extending = True
+
+    def shim_sequencer(self):
```
po09i: I think we should move most of the error checking in the init and have the "shim" part be mostly just that, shimming. You could create a local method that includes most of this error checking. I would think that there is a lot of duplication between dynamic sequencer and the realtime one so most of this error checking could probably move to the base sequencer class.

po09i: Also as mentioned above, rename to shim()

shimmingtoolbox/shim/sequencer.py
```diff
@@ -42,838 +38,1063 @@
...
+        ax.set_xlabel('Shim group')
+        ax.set_ylabel('Coefficients (Physical CS [RAS])')
+        ax.legend()
+        ax.set_title("Currents through shims")
+        fname_figure = os.path.join(self.path_output, 'fig_currents.png')
+        fig.savefig(fname_figure)
+        logger.debug(f"Saved figure: {fname_figure}")
+
+
+class ShimSequencer(Sequencer):
```
po09i, line 148: Document the attributes of the class (every time you declare self.xxxxx and that it is not a method)

shimmingtoolbox/shim/sequencer.py
po09i: Consider using @property so you can error check every time an attribute is changed. This would allow to be cleaner rather than have a "one function that checks everything" and if one attribute changes then we have to call the function that checks everything again.

po09i (commented):
I started my review, I think this is already a lot of things to change so I will wait for your next changes before continuing
Some other minor comments:
- Remove unnecessary imports
- Add the type information in the Return. (See google documentation style)

shimmingtoolbox/shim/sequencer.py
Aurelienpujolm: I documented the attributes, it should be good now !

shimmingtoolbox/shim/sequencer.py
Aurelienpujolm: I separated the two parts of the sequencer. However, I do think that is good to keep in the real time sequencer, the calculation of the parameters for the real time shimming. (Because it's not just checking the dimensions of everything, but also a lot of mathematical operation etc.)
If needed, I can also separate it in a different function

kousu: @Aurelienpujolm asked me for some advice. I have not read all the code in detail, but on this point:

1. `check_inputs` should be deleted.

    My instinct is that checking preconditions  be done _in each method_, near where they are used. Perhaps some precondition checking should happen in `__init__`, but it is a lot easier to avoid bugs if each function that assumes it has a 3d array checks that it has 3d array. It means more code but it's boilerplate code that's easy to write and easy to fix if it's wrong, and it avoid the ambiguity about "when/where is it safe to/must I call this `check_inputs()` function?".

    Also `check_inputs()` isn't just checking inputs, it is editing them too, which is misleading:

   https://github.com/shimming-toolbox/shimming-toolbox/blob/fe37b5d1a725a2deaab7697a275489e16928d267/shimmingtoolbox/shim/sequencer.py#L220-L230

    There should only be one line that says `self.nii_fieldmap = ` and it should be in `__init__`. All the transformations/normalizations needed should be done to that data before it is saved as a member variable.
2. But `property()` isn't a good idea either. [Immutability](https://realpython.com/python-mutable-vs-immutable-types/) is.

    Simply disallow people editing the shim parameters. If they want to change the shim parameters, force them to make a new shim object, which then runs through all the precondition checking in `__init__()` (or maybe in `_check_inputs()`, but again I'd recommend that only if it is marked as a private method so that users aren't confused).

kousu: > I think we should move most of the error checking in the init and have the "shim" part be mostly just that, shimming.

I also think you can go further and stop using a class at all: combine `__init__()` and `shim()` into a single **function** (**not a method**) that does all the work in one step, and returns an object with the results (meaning your `.coefs`, `.slices`, etc). Move all the helper methods like `plot_current` and `eval` to their own top-level functions.

An _object_ is supposed to be a little miniature state machine: some data plus some methods that manipulate that data to keep the different pieces of it consistent. But state machines are notoriously hard to get right; it's so easy to add a little edit to some method that subtly invalidates other pieces of data, but only if that method gets called in a certain order. It's a nightmare to keep track of. Immutability is such a much stronger technique that objects are just unnecessary, in my opinion.

I read the addition of [`dataclass`](https://docs.python.org/3/library/dataclasses.html) as python recognizing that objects are a mistake, and structs + top level functions that know how to interpret them are better.

----

And while avoiding objects, also avoid inheritance. Your life is better without it. It is difficult and unnecessary and has wasted a lot of money in the last 20 years. ["Composition" is better](https://www.digitalocean.com/community/tutorials/composition-vs-inheritance), which simply means: write subroutines. That doesn't mean you can't have polymorphism -- you can design APIs and then write a series of functions that fit them, a set that all have the same input and output types or you can design [protocols](https://peps.python.org/pep-0544/) (what other languages call 'interfaces') -- but you can do it without confusing yourself and future maintainers with trying to use inheritance.

shimmingtoolbox/shim/sequencer.py
```diff
@@ -23,857 +21,1365 @@
...
-                                    'cube'. See :func:`shimmingtoolbox.masking.mask_utils.dilate_binary_mask` for more
-                                    details.
-        mask_dilation_kernel_size (int): Length of a side of the 3d kernel to dilate the mask. Must be odd. For example,
-                                         a kernel of size 3 will dilate the mask by 1 pixel.
-        path_output (str): Path to the directory to output figures. Set logging level to debug to output debug
+            path_output (str): Path to the directory to output figures. Set logging level to debug to output debug
                            artefacts.
+        """
+        super().__init__(slices, mask_dilation_kernel, mask_dilation_kernel_size, reg_factor, path_output)
+        self.nii_fieldmap, self.nii_fieldmap_orig, self.extending = self.get_fieldmap(nii_fieldmap)
```
```suggestion
        self.nii_fieldmap, self.nii_fieldmap_orig, self.fmap_is_extended = self.get_fieldmap(nii_fieldmap)
```

Change `self.extending` to `self.fmap_is_extended` since just extending is not clear

shimmingtoolbox/shim/sequencer.py
```diff
@@ -23,857 +21,1365 @@
...
+        Returns:
+            numpy.ndarray: coefs : Coefficients of the coil profiles to shim (len(slices) x n_channels)
+        """
+        # Select and initialize the optimizer
+        self.select_optimizer()
+        # Optimize and get the coefficients
+        coefs = self.optimize(self.nii_mask_anat)
+        return coefs
+
+    # TODO : Maybe put the select_optimizer in the init
```
```suggestion

```

shimmingtoolbox/shim/sequencer.py
```diff
@@ -23,857 +21,1365 @@
...
+                           'least_squares_rt'.
+        Returns:
+            Optimizer : Initialized Optimizer object
+        """
+
+        # global supported_optimizers
+        if self.method in supported_optimizers:
+            if self.method == 'least_squares':
+                optimizer = supported_optimizers[self.method](self.coils, unshimmed, affine, self.opt_criteria,
+                                                              reg_factor=self.reg_factor)
```
po09i, line 1062: This will never get executed because if it's realtime, it will use the realtime optimizer:
```
        if self.method == 'least_squares':
            self.method = 'least_squares_rt'
```

We can remove it

Aurelienpujolm: I'm not sure about that, because in the case of a RT shimming, we are first doing a static optimization and then an optimization with the Riro ? If it's the case then we have to make sure to do both optimization ? (I am not sure about that)

po09i: Good catch

po09i (approved):
LGTM! Great job! Can you write a comment in the PR about what is the current time/RAM usage you are getting with the same dataset/options/computer used for previous optimization PRs.

Aurelienpujolm:
With this PR, a Mac computer 10.15 with a RAM of 16 Go, and 8 hearts,  and a dataset of : 
coil ( 32 channels) ;
anatomic image ([174,174,50] );
fieldmap([112,112,70]);
mask ([112,112,70]).
The time of the B0 shimming is now around 63 seconds VS 80 before this PR(https://github.com/shimming-toolbox/shimming-toolbox/pull/432) . However, this time will be lowered by some new PR : 
- One that will allow the user to save the fig in a faster way (on the same computer it allows me to save around 8 seconds so far)
- One that will save some time by not resampling two times the same mask in some cases (it will save in my case around 15 seconds) 
- Another one that will reimplement multiprocessing in the optimization to save up some time (8 seconds of save)
- A last one to save up some time in calc_shimmed_anat_orient by not resampling the coils on the mask. (5 seconds of save)
With all of these changes, the B0 shimming could take around 25-30 seconds which will be great, and it could be even lower with better computers)

PR #442, opened 2023-03-09
Fixed a particular case, when there is nothing to shimmed in an optimization

Aurelienpujolm:
## Checklist

## Description
In some case, it's possible that there is nothing to shimmed in an optimization, like for example if the riro is equal zero in real time shimming. Because of this, the program could crash because we were dividing by stability factor that can be equal to zero if unshimmed_vec is only equal zero. 
This PR fix this issue

### DIFF
--- a/shimmingtoolbox/optimizer/lsq_optimizer.py
+++ b/shimmingtoolbox/optimizer/lsq_optimizer.py
@@ -275,7 +275,7 @@ class LsqOptimizer(Optimizer):
         currents_0 = self.get_initial_guess()
         #In case if there is nothing to shimmed
         if np.all(unshimmed_vec == 0):
-            return currents_0
+            return np.zeros(np.shape(currents_0))
         # Optimize
         # When clipping to bounds, scipy raises a warning. Since this can be frequent for our purposes, we ignore that
         # warning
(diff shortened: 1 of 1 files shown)


shimmingtoolbox/optimizer/lsq_optimizer.py
```diff
@@ -271,7 +273,9 @@ def optimize(self, mask):
 
         # Set up output currents
         currents_0 = self.get_initial_guess()
-
+        #In case if there is nothing to shimmed
+        if np.all(unshimmed_vec == 0):
+            return currents_0
```
po09i: I don't think you shout base yourself of `currents_0`. It depends on the initial guessing method (it's 0s for now but it could also be set to return halfway through the bounds or return the values of an adjacent slice). If `unshimmed_vec` is 0s then the output currents should always be 0s.

Aurelienpujolm: It's fixed !

shimmingtoolbox/optimizer/lsq_optimizer.py
```diff
@@ -271,7 +273,9 @@ def optimize(self, mask):
 
         # Set up output currents
         currents_0 = self.get_initial_guess()
-
+        #In case if there is nothing to shimmed
```
po09i:
```suggestion
        # If what to shim is already 0s
```

po09i (approved):

PR #443, opened 2023-03-27
Make Resampling faster by using multiprocessing with joblib

### LINKED ISSUE #434: Makes resampling faster
### Description
I have been using the _resample_from_to_ function from the nibabel library (https://github.com/nipy/nibabel) to resample images for example in the B0 shimming, and I have noticed that it takes a long time to execute, especially when working with large images. I am concerned that this could become a bottleneck for the B0 shimming,  and would like to find a solution to improve its performance.

### Steps to Reproduce

- Go to ap/faster-B0shimming-v2
- Load a large anatomic image, mask, and coils (to have the specific files please contact aurelienpujolm@outlook.fr)  
- Launch the B0 dynamic test with the files
- Observe the execution time, and profile the run 

### Expected Behavior
The execution time of the resample_from_to function should be reasonable and not take an excessive amount of time.
### Actual Behavior
The execution time of the _resample_from_to_ function is much longer than expected and takes an excessive amount of time. For example to resample a slice of an anatomic image of dimension (174,174,50), to a mask of dimensions (112,112,70), this function can takes up to 0.2 seconds (which is a lot because this procedures can be done up to 250 times in the B0 shimming ) 
### Environment

- Operating System: [MacOs 10.15]
- Python version: [3.9]
- Django computer

### Additional Context
During my research, I’ve found out that the most consuming process in the _resample_from_to function_ come from the use of the _scipy.ndimage.affine_transform_ function (https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.affine_transform.html#scipy.ndimage.affine_transform), which is a high time consuming function (takes 0.2 seconds per iteration with the size of the images described above).
Some other libraries should be able to do affine transformation like this function, however I did not succeed to reproduce the same results as with the scipy function. I tried the _OpenCv_ library (https://docs.opencv.org/3.4/d4/d61/tutorial_warp_affine.html ), _scikit-image_(https://scikit-image.org/docs/stable/api/skimage.transform.html#skimage.transform.warp ). (Maybe if someone succeeds in reproducing the results it could be great). 
Also the _Dipy_ library (https://github.com/dipy/dipy ) has a function to resample images, and it could be interesting to see if we can use it.

Fixing this issue should reduce the time of the BO shimming, and would be helpful  !

Aurelienpujolm:
## Checklist

## Description
This PR will fix the issue of slow resampling for huge datasets. Indeed,  doing all of the resamplings for big dataset can take up to one minute. This can be solved in the case of 4D images by multiprocessing.  Because in this case, without multiprocessing, we would make the resampling for each 3D array corresponding to the fourth dimension index. So with a multiprocessing loop on the fourth dimension , we can, in theory, divide the time for the resampling divide by the number of cores in the computer. 
However, this is still not perfect because joblib takes some time to start, but it can reduce the time of 4D resampling by a lot of time. (In my case, with a quadcores , it goes from around 20 seconds to 8 seconds, which is a good start, but not the ideal 2.5 seconds).
If, in the future, someone finds a good way to make fast and safe multiprocessing, it could be a good idea to implement it in the resampling, to save more time 

## Linked issues
Addresses #434

### DIFF
--- a/shimmingtoolbox/coils/coordinates.py
+++ b/shimmingtoolbox/coils/coordinates.py
@@ -5,12 +5,9 @@
 import numpy as np
 from nibabel.affines import apply_affine
 import math
-import multiprocessing as mp
 import nibabel as nib
-from joblib import Parallel, delayed
 from nibabel.processing import resample_from_to as nib_resample_from_to
-
-
+from joblib import Parallel, delayed, effective_n_jobs
 def generate_meshgrid(dim, affine):
     """
     Generate meshgrid of size dim, with coordinate system defined by affine.
@@ -167,7 +164,7 @@ def resample_from_to(nii_from_img, nii_to_vox_map, order=2, mode='nearest', cval
 
     elif from_img.ndim == 4:
         nt = from_img.shape[3]
-        nb_cpu_used = int(mp.cpu_count())
+        nb_cpu_used = int(effective_n_jobs())
         results = Parallel(n_jobs=nb_cpu_used, backend='loky')(
             delayed(_resample_4d)( nii_from_img, nii_to_vox_map, it, order, mode, cval, out_class)
             for it in range(nt))
@@ -188,6 +185,13 @@ def _resample_4d(nii_from_img, nii_to_vox_map, i, order, mode, cval, out_class )
     resampled_image = nib_resample_from_to(nii_from_img_3d, nii_to_vox_map, order=order, mode=mode,
                                             cval=cval, out_class=out_class).get_fdata()
     return i, resampled_image
+
+
+# def _resample_3d(nii_3d, nii_to_vox_map, order, mode, cval, out_class):
+#
+#     nii_resampled_3d = nib_resample_from_to(nii_3d, nii_to_vox_map, order=order, mode=mode, cval=cval,
+#                                             out_class=out_class)
+#     return nii_resampled_3d.get_fdata()
 
 
 def get_main_orientation(cosines: list):
--- a/shimmingtoolbox/coils/coordinates.py
+++ b/shimmingtoolbox/coils/coordinates.py
@@ -165,13 +165,12 @@ def resample_from_to(nii_from_img, nii_to_vox_map, order=2, mode='nearest', cval
 
     elif from_img.ndim == 4:
         nt = from_img.shape[3]
-        resampled_4d = np.zeros(nii_to_vox_map.shape + (nt,))
-
-        for it in range(nt):
-            nii_from_img_3d = nib.Nifti1Image(from_img[..., it], nii_from_img.affine, header=nii_from_img.header)
-            nii_resampled_3d = nib_resample_from_to(nii_from_img_3d, nii_to_vox_map, order=order, mode=mode,
-                                                    cval=cval, out_class=out_class)
-            resampled_4d[..., it] = nii_resampled_3d.get_fdata()
+        nb_cpu_used = int(mp.cpu_count())
+        results = Parallel(n_jobs=nb_cpu_used, backend='loky')(
+            delayed(_resample_4d)( nii_from_img, nii_to_vox_map, it, order, mode, cval, out_class)
+            for it in range(nt))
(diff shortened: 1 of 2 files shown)


shimmingtoolbox/coils/coordinates.py
```diff
@@ -165,13 +167,12 @@ def resample_from_to(nii_from_img, nii_to_vox_map, order=2, mode='nearest', cval
...
     elif from_img.ndim == 4:
         nt = from_img.shape[3]
-        resampled_4d = np.zeros(nii_to_vox_map.shape + (nt,))
-
-        for it in range(nt):
-            nii_from_img_3d = nib.Nifti1Image(from_img[..., it], nii_from_img.affine, header=nii_from_img.header)
-            nii_resampled_3d = nib_resample_from_to(nii_from_img_3d, nii_to_vox_map, order=order, mode=mode,
-                                                    cval=cval, out_class=out_class)
-            resampled_4d[..., it] = nii_resampled_3d.get_fdata()
+        nb_cpu_used = int(mp.cpu_count())
```
po09i: Y a t'il une alternative à `mp.cpu_count()` dans joblib? Si oui ça éviterait d'importer joblib et multiprocessing

shimmingtoolbox/coils/coordinates.py
```diff
@@ -165,13 +167,12 @@ def resample_from_to(nii_from_img, nii_to_vox_map, order=2, mode='nearest', cval
...
+        results = Parallel(n_jobs=nb_cpu_used, backend='loky')(
+            delayed(_resample_4d)( nii_from_img, nii_to_vox_map, it, order, mode, cval, out_class)
+            for it in range(nt))
+        results.sort(key=lambda x: x[0])
```
po09i: Are you sure you need "it" to sort the output data? Maybe joblib already outputs it in the right order.

shimmingtoolbox/coils/coordinates.py
Aurelienpujolm: J'ai trouvé une alternative, mais pour l'instant il faut garder multiprocessing dans le setup.py, puisque dans cette version du code, il est encore utilisé dans _optimize() 
Je l'enlèverais dans une future PR, une fois que multiprocessing sera totalement enlevé !

shimmingtoolbox/coils/coordinates.py
Aurelienpujolm: So, I do need it to make a loop on joblib https://joblib.readthedocs.io/en/latest/parallel.html.
However, it should be possible to removed the following line "results.sort", but I suggest keeping it, to make sure that there is no problem with the indexes of the image

shimmingtoolbox/coils/coordinates.py
```diff
@@ -7,8 +7,7 @@
 import math
 import nibabel as nib
 from nibabel.processing import resample_from_to as nib_resample_from_to
-
-
+from joblib import Parallel, delayed
```
po09i, line 10:
```suggestion
from joblib import Parallel, delayed


```

shimmingtoolbox/coils/coordinates.py
```diff
@@ -165,14 +164,11 @@ def resample_from_to(nii_from_img, nii_to_vox_map, order=2, mode='nearest', cval
...
-        for it in range(nt):
-            nii_from_img_3d = nib.Nifti1Image(from_img[..., it], nii_from_img.affine, header=nii_from_img.header)
-            nii_resampled_3d = nib_resample_from_to(nii_from_img_3d, nii_to_vox_map, order=order, mode=mode,
-                                                    cval=cval, out_class=out_class)
-            resampled_4d[..., it] = nii_resampled_3d.get_fdata()
-
+        results = Parallel(-1, backend='loky')(
+            delayed(_resample_4d)(it, nii_from_img, nii_to_vox_map, order, mode, cval, out_class)
+            for it in range(nt))
+        # results.sort(key=lambda x: x[0])
```
po09i:
```suggestion
```

po09i (approved):

Issue #446, opened 2023-04-12
st_dicom_to_nifti of dual-echo fieldmapping sequences wrongly rename echo 2 to phasediff and delete echo1


po09i:
### Bug Description
Dcm2niix outputs phasediff images with the name "_ph2". ST has a check to rename "ph2" into phasediff. Among other things, we look that the json sidecar has both echo times, which would indicate a phasediff image. This is done on all phase images in the output 'fmap' directory and renames the files accordingly. If a sequence outputs 2 phase images (phase1 and phase2), and that for some reason, the JSON files of both phase images have both echo times instead of only their respective echo time, ST would rename both to the same thing, destroying echo1 in the process and wrongly identifying  them as phasediff.

po09i:
This is a caveat that was originally mentioned when the feature was introduced: https://github.com/rordenlab/dcm2niix/issues/400#issuecomment-632113117

PR #447, opened 2023-04-12
Fix phasediff renaming issue for dual-echo fieldmaps

po09i:
## Description
This PR solves an issue in st_dicom_to_nifti where it renamed files `phase1` and `phase2` to `phasediff` in dual echo field maps. To solve the issue, a check makes sure that the file is `EchoNumber`: 2 (phase difference images have an EchoNumber of 2) and makes sure that there is not a `phase1` file that was converted which would indicate phase2 is not a phasediff.

## Linked issues
Solves #446


dpapp86 (approved):
Tried, tested, and approved. The two phase images created via the ST GUI are now identical, both in terms of image data and JSON sidecars, to the two phase images created via dcm2niix

PR #451, opened 2023-05-11
Changed the optimization to get faster results

Aurelienpujolm:
## Checklist

## Description
This PR is made to make the optimization faster thanks to joblib multiprocessing. On a Mac computer 10.15 with a RAM of 16 Go, and 8 hearts, and a dataset of :
coil ( 32 channels) ;
anatomic image ([174,174,50] );
fieldmap([112,112,70]);
mask ([112,112,70]).
The optimization went from around 23 seconds to 9 seconds. And is now taking 15 % of the total time of the B0 shimming. It would be interesting in the future to have better usages of mask and resampling, and to make the figures for the user in a faster way

### DIFF
--- a/shimmingtoolbox/masking/mask_utils.py
+++ b/shimmingtoolbox/masking/mask_utils.py
@@ -4,16 +4,15 @@
 import logging
 import nibabel as nib
 import numpy as np
-import os
-from scipy.ndimage import binary_dilation, binary_opening, generate_binary_structure, iterate_structure
 
+from scipy.ndimage import binary_dilation, binary_opening, generate_binary_structure, iterate_structure
 from shimmingtoolbox.coils.coordinates import resample_from_to
 
 logger = logging.getLogger(__name__)
 
 
 def resample_mask(nii_mask_from, nii_target, from_slices=None, dilation_kernel='None', dilation_size=3,
-                  path_output=None):
+                  path_output=None, both_masks=False):
     """
     Select the appropriate slices from ``nii_mask_from`` using ``from_slices`` and resample onto ``nii_target``
 
@@ -26,11 +25,11 @@ def resample_mask(nii_mask_from, nii_target, from_slices=None, dilation_kernel='
         dilation_size (int): Length of a side of the 3d kernel to dilate the mask. Must be odd. For example,
                                          a kernel of size 3 will dilate the mask by 1 pixel.
         path_output (str): Path to output debug artefacts.
+        both_masks (bool): See if we want to return the dilated and non dilated resampled mask
 
     Returns:
         nib.Nifti1Image: Mask resampled with nii_target.shape and nii_target.affine.
     """
-
     mask_from = nii_mask_from.get_fdata()
 
     if from_slices is None:
@@ -42,13 +41,10 @@ def resample_mask(nii_mask_from, nii_target, from_slices=None, dilation_kernel='
 
     # Create nibabel object of sliced mask
     nii_mask = nib.Nifti1Image(sliced_mask.astype(float), nii_mask_from.affine, header=nii_mask_from.header)
-
     # Resample the sliced mask onto nii_target
     nii_mask_target = resample_from_to(nii_mask, nii_target, order=1, mode='grid-constant', cval=0)
-
     # Resample the full mask onto nii_target
     nii_full_mask_target = resample_from_to(nii_mask_from, nii_target, order=0, mode='grid-constant', cval=0)
-
     # TODO: Deal with soft mask
     # Find highest value and stretch to 1
     # Look into dilation of soft mask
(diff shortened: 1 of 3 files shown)


po09i (commented):
Awesome, it is now much faster on big datasets! 

The tests are now much slower however, probably because the tests are small and launching joblib is slow and unecessary in that case. Did we settle on a good way to handle this? (condition on # of pixels to optimize/# of slices/etc)

Aurelienpujolm:
> Awesome, it is now much faster on big datasets!
> 
> The tests are now much slower however, probably because the tests are small and launching joblib is slow and unecessary in that case. Did we settle on a good way to handle this? (condition on # of pixels to optimize/# of slices/etc)

So I have some ideas about that, but first it would be interesting to know how much time we are losing for small dataset (1 seconds or 10 ?) 

This is my ideas : 

### User friendly way
The goal is to provide the user with the option to choose whether or not to use multiprocessing. If they have a small dataset, they can opt-out of multiprocessing, while they can opt-in for a large dataset. Additionally, we can include a message at the end of the optimization to inform the user that multiprocessing would be beneficial for the next shimming with the same dataset size if the optimization takes more than, for instance, 10 seconds.

The primary benefit of this approach is that it simplifies the code development and reduces the time spent in most cases. However, the downside is that if the user decides not to use multiprocessing for the first shimming and has a large dataset, it may result in a delay of up to 10-15 seconds for the first shimming. This is assuming the dataset that we are using for the test is the largest possible.

### The "arbitrary method" 

The second approach requires some preliminary testing to determine whether multiprocessing is needed or not. We need to determine the number of slices that require optimization. If we know that some slices don't require optimization, we can exclude them from the process. Then, we can test the optimization time for the first slice. If it exceeds 10/(n_slices_to_shimmed), we can employ multiprocessing. Otherwise, we can skip it.

The benefit of this method is that the user doesn't need to make a choice. However, if the first slice is significantly smaller than the rest, not using joblib may lead to a loss of time. Similarly, if the first slice is more prominent than the rest, using joblib may result in a time loss.

Personnaly I would suggest to go with the first method, but the second one would also be possible.

po09i:
> User friendly way

This is a good idea, I would like to check first if we can try to keep the user input to a minimum. I think some testing about what takes time might be good. Are the long testing times because we are launching joblib for a single or two slices (easy fix)? Or is it a function of the number of slices/number of pixels/number of processing cores/etc where it quickly becomes unmanageable?

Aurelienpujolm:
> > User friendly way
> 
> This is a good idea, I would like to check first if we can try to keep the user input to a minimum. I think some testing about what takes time might be good. Are the long testing times because we are launching joblib for a single or two slices (easy fix)? Or is it a function of the number of slices/number of pixels/number of processing cores/etc where it quickly becomes unmanageable?

It's a mix between everyhting in fact: 
If we have for exemple less than 8 slices (which is the number of cores of my computer), then multiprocessing may not be the best idea. For example in my case for one slice it takes around 0.4 seconds to do the optimization. So in theory it would be interesting to start joblib multiprocessing from 20 slices.  But it may depends of the computer etc. 
But it's also depending of the number of pixel, because resampling may take longer time if there is a lot of pixel, and so is optimization. And so if doing the procedure for one slice takes 1 seconds then multiprocessing would be necessary starting from 8 slices etc. 
So overall I would say, that it may be difficult to anticipate to do or not joblib multiprocessing.
 It would be also interresting to see how much time we may lose by doing multiprocessing when it's not necessary (is it 1, 5 or 10 seconds)

Aurelienpujolm:
I have now updated the code to address the issue of longer optimization times for smaller datasets while achieving shorter times for larger datasets.
### Performance
First, I have included a graph summarizing the optimization times using 8 different methods with the dataset described in this PR. The graph represents the optimization times for four scenarios: without multiprocessing, with multiprocessing for all steps, with multiprocessing only for mask resampling, and with separate multiprocessing for mask resampling and the "optimization" part.

The subsequent four graphs represent the same scenarios but with different types of optimization. You can refer to the following image for a visual representation:

image
The four trendlines, from top to bottom, represent:

- The first version of optimization without multiprocessing
- The second version of optimization without multiprocessing
- The first version of optimization with multiprocessing for mask resampling
- The second version of optimization with multiprocessing for mask resampling

What is interesting to note is that the time taken follows a linear trend (with some trendlines having an R2 value of 1). Furthermore, the option of multiprocessing with mask resampling consistently outperforms the version without multiprocessing, allowing for time savings of up to 10 seconds.

Additionally, I have provided a similar graph using a smaller dataset, which is the dataset used in the testing phase. Here, the modifications I made did not significantly change the time, with differences of only 0.1 seconds at best. This indicates that the method can still be effectively used for smaller datasets.

image

### Conclusion for the performance :

**The latest method appears to be the best option since it does not slow down the program for small datasets and significantly improves performance for larger datasets.**

### Modification of the optimizer and consequences
The objective of this PR was to improve the optimization speed, which required changing the mathematical expression of the function to be minimized (including the Jacobian). The aim was to minimize the number of calculations within the function (as it would be called thousands of times). After simplification, I simplified the expression of the residuals mean squared error (MSE) from:
image
To: 
image
Where a,b,c are : 
            inv_factor = 1 / (len(unshimmed_vec) * factor)
            a = (coil_mat.T @ coil_mat) * inv_factor + np.diag(self.reg_vector)
            b = 2 * inv_factor * (unshimmed_vec @ coil_mat)
            c = inv_factor * (unshimmed_vec @ unshimmed_vec)
This modification resulted in fewer calculations within the function to be minimized. However, I made a slight change to the formula of the residuals MSE. Instead of using np.abs(coef) for the second term, I replaced it with the equivalent of np.square(coef). This allowed for further simplification and saved computation time.

In terms of results, I observed improved MSE results for shimming but slightly worse results for mean absolute error (MAE). You can refer to the following image for the new shimming results compared to the old results:
image
_New results of the shimming_
image
_old results of the shimming_
**So the newest method seems to be better also for shimming results.** 
### Drawbacks
Due to the inclusion of a squared regulation term, the solution may now find higher current values compared to before. However, since the currents are bound, this should not cause any issues. In my case, the currents are quite close, with the mean of the currents increasing from 0.31A to 0.34A

po09i:
A side effect is that it changes the regularization coefficient to a L2 penalty

po09i (approved):
Awesome work! We are getting pretty fast!

PR #454, opened 2023-05-25
Changed version of matplotlib

Aurelienpujolm:
## Checklist

This PR changed the version of matplotlib to matched the one needed for fsleyes

### DIFF
--- a/shimmingtoolbox/cli/b0shim.py
+++ b/shimmingtoolbox/cli/b0shim.py
@@ -1119,9 +1119,7 @@ def _add_sub_figure(fig, i_plot, n_plots, static_coefs, bounds, min_y, max_y, un
 
     # Add realtime component as an errorbar
     if rt_coefs is not None:
-        rt_coef_ishim = rt_coefs
-        riro = [rt_coef_ishim * -pres_probe_min, rt_coef_ishim * pres_probe_max]
-        riro = np.abs(riro)
+        riro = np.abs([rt_coef * -pres_probe_min, rt_coef * pres_probe_max])
         ax.errorbar(range(n_channels), static_coefs, yerr=riro, fmt='o', elinewidth=4, capsize=6,
                     label='static-riro')
     # Add static component
--- a/shimmingtoolbox/cli/b0shim.py
+++ b/shimmingtoolbox/cli/b0shim.py
@@ -1119,8 +1119,8 @@ def _add_sub_figure(fig, i_plot, n_plots, static_coefs, bounds, min_y, max_y, un
 
     # Add realtime component as an errorbar
     if rt_coefs is not None:
-        riro = np.abs([np.min(rt_coefs * -pres_probe_min, rt_coefs * pres_probe_max),
-                       np.max(rt_coefs * -pres_probe_min, rt_coefs * pres_probe_max)])
+        riro = np.abs([min(rt_coefs * -pres_probe_min, rt_coefs * pres_probe_max),
+                       max(rt_coefs * -pres_probe_min, rt_coefs * pres_probe_max)])
         ax.errorbar(range(n_channels), static_coefs, yerr=riro, fmt='o', elinewidth=4, capsize=6,
                     label='static-riro')
     # Add static component
--- a/shimmingtoolbox/cli/b0shim.py
+++ b/shimmingtoolbox/cli/b0shim.py
@@ -1119,7 +1119,8 @@ def _add_sub_figure(fig, i_plot, n_plots, static_coefs, bounds, min_y, max_y, un
 
     # Add realtime component as an errorbar
     if rt_coefs is not None:
-        riro = np.abs([rt_coefs * -pres_probe_min, rt_coefs * pres_probe_max])
+        riro = np.abs([np.min(rt_coef_ishim * -pres_probe_min, rt_coef_ishim* pres_probe_max),
+                       np.max(rt_coef_ishim * -pres_probe_min, rt_coef_ishim* pres_probe_max))])
         ax.errorbar(range(n_channels), static_coefs, yerr=riro, fmt='o', elinewidth=4, capsize=6,
                     label='static-riro')
     # Add static component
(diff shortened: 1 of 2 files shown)


shimmingtoolbox/cli/b0shim.py
```diff
@@ -1121,6 +1121,7 @@ def _add_sub_figure(fig, i_plot, n_plots, static_coefs, bounds, min_y, max_y, un
     if rt_coefs is not None:
         rt_coef_ishim = rt_coefs
         riro = [rt_coef_ishim * -pres_probe_min, rt_coef_ishim * pres_probe_max]
+        riro = np.abs(riro)
```
po09i: I get that taking the absolute value fixes the error, but does it give the same result if we have a negative value?

Aurelienpujolm: It would do an errorbar on the plot , and if you are looking to the plot that it's plotting you should have what we want (you can just run the realtime test with custom coils to see the case where the error happened before) 
The fact is that the first value which can be negative, needs to be positive. And it should be by how much (with a positive value) we can have values below rt_coef. So for me, this should work.

Aurelienpujolm: Making it with np.abs, is indeed fixing the issue since the yerr is bounds are a distance and not a position : see https://matplotlib.org/stable/gallery/statistics/errorbar_features.html

po09i: `pres_probe_min` and  `pres_probe_max` are always positive (e.g., 1000 and  2000 respectively)
Let's make sure it works using some dummy values (rt_coef_ishim = -1, static_coefs=0)
- If we are at pressure `pres_probe_min` then the output value will be 1000 (static_coefs + rt_coef_ishim * -pres_probe_min)
- If we are at pressure `pres_probe_max ` then the output value will be -2000 (static_coefs + rt_coef_ishim * pres_probe_max)

The plot should show (-2000, 1000)

If we use the absolute value, we would get (-1000, 2000)

po09i (changes requested):
I do not believe these 2 commits ([0d28910](https://github.com/shimming-toolbox/shimming-toolbox/pull/454/commits/0d289103f974198dbbf232df262876405a67e38c), [1953d8f](https://github.com/shimming-toolbox/shimming-toolbox/pull/454/commits/1953d8f3e1086a213899f951189131b42a16990a)) addressed my [comment](https://github.com/shimming-toolbox/shimming-toolbox/pull/454#discussion_r1213658551)

Aurelienpujolm:
> I do not believe these 2 commits ([0d28910](https://github.com/shimming-toolbox/shimming-toolbox/pull/454/commits/0d289103f974198dbbf232df262876405a67e38c), [1953d8f](https://github.com/shimming-toolbox/shimming-toolbox/pull/454/commits/1953d8f3e1086a213899f951189131b42a16990a)) addressed my [comment](https://github.com/shimming-toolbox/shimming-toolbox/pull/454#discussion_r1213658551)

Does this fix the issue for you : 
        riro = [static_coefs - rt_coef_ishim * -pres_probe_min, rt_coef_ishim * pres_probe_max - static_coefs]
        riro = np.abs(riro)
Here, both term needs to be positive, and represent by how much we can go below or above the coefs that we got, with the limit being the same as before [rtcoef_ishim * - pres_prob_min, rt_coef_ishim* pres_prob_max]
Is it what you expected or not ?

shimmingtoolbox/cli/b0shim.py
po09i: > > I do not believe these 2 commits ([0d28910](https://github.com/shimming-toolbox/shimming-toolbox/pull/454/commits/0d289103f974198dbbf232df262876405a67e38c), [1953d8f](https://github.com/shimming-toolbox/shimming-toolbox/pull/454/commits/1953d8f3e1086a213899f951189131b42a16990a)) addressed my [comment](https://github.com/shimming-toolbox/shimming-toolbox/pull/454#discussion_r1213658551)
> 
> Does this fix the issue for you : riro = [static_coefs - rt_coef_ishim * -pres_probe_min, rt_coef_ishim * pres_probe_max - static_coefs]
riro = np.abs(riro)
Here, both term needs to be positive, and represent by how much we can go below or above the coefs that we got, with the limit being the same as before [rtcoef_ishim * - pres_prob_min, rt_coef_ishim* pres_prob_max] Is it what you expected or not ?

No, this gives the same problem as before with the example I gave ((-1000, 2000) and we expect (-2000, 1000))

Aurelienpujolm: In that case 
riro = [static_coef - np.min(rt_coef_ishim * -pres_probe_min,  rt_coef_ishim* pres_probe_max), np.max(rt_coef_ishim * -pres_probe_min,  rt_coef_ishim* pres_probe_max) - static_coef)]
Should do the work no ? (Is the distance that matters, and with your example it's working)

po09i: Shouldn't it be?
```
riro = [np.min(rt_coef_ishim * -pres_probe_min, rt_coef_ishim* pres_probe_max), np.max(rt_coef_ishim * -pres_probe_min, rt_coef_ishim* pres_probe_max))]
```

Aurelienpujolm: No because if for example the minimum is negative it will crash ! 
A safe way to do it without static coef should be to do 
riro = np.abs([np.min(rt_coef_ishim * -pres_probe_min, rt_coef_ishim* pres_probe_max), np.max(rt_coef_ishim * -pres_probe_min, rt_coef_ishim* pres_probe_max))])
Like this the term will always be positive, but it would do the same thing as before also (because is it still a distance)
Is it good for you ?

po09i: What needs to be fed to `yerr` is the distance from static_coef according to your previous [comment](https://github.com/shimming-toolbox/shimming-toolbox/pull/454#discussion_r1213366277). Therefore I think your last proposition including `min`/`max` and `abs` makes sense.

Aurelienpujolm: I needed it to change it a little bit because, riro is an array of dimension (2,n) , so just do np.min and max couldn't work.
So this is the last version : 
        riro = np.zeros((2, rt_coefs.shape[0]))
        for i_slice in range(rt_coefs.shape[0]):
            riro[:, i_slice] = np.abs([min(rt_coefs[i_slice] * -pres_probe_min, rt_coefs[i_slice] * pres_probe_max),
                                    max(rt_coefs[i_slice] * -pres_probe_min, rt_coefs[i_slice] * pres_probe_max)])

po09i: Looks good to me

po09i (approved):
Looks good!

PR #459, opened 2023-06-16
Allow to sort any DICOM files

po09i:
## Description
This PR allows to sort any DICOM files (even those without extension) using `st_sort_dicoms`


dpapp86 (commented):
Works at intended. Approved

dpapp86 (approved):
And approved.

PR #460, opened 2023-06-22
Solved a bug where mask_fmap and the dilated mask where switched

Aurelienpujolm:
## Checklist

## Description
I've just changed the index to get the mask_fmap and the dilated_mask in the sequencer, because I inverted them in the last PR

### DIFF
--- a/shimmingtoolbox/shim/sequencer.py
+++ b/shimmingtoolbox/shim/sequencer.py
@@ -87,8 +87,8 @@ class Sequencer(object):
         Optimization of the currents for each shim group. Wraps :meth:`shimmingtoolbox.shim.sequencer.Sequencer.opt`.
 
         Args:
-            masks_fmap_dilated (np.ndarray): 3D anat mask used for the optimizer to shim in the region of interest.
-                                             (only consider voxels with non-zero values)
+            masks_fmap_dilated (np.ndarray): 3D fieldmap mask used for the optimizer to shim in the region
+                                             of interest (only consider voxels with non-zero values)
         Returns:
                 np.ndarray: Coefficients of the coil profiles to shim (len(slices) x n_channels)
         """
--- a/shimmingtoolbox/shim/sequencer.py
+++ b/shimmingtoolbox/shim/sequencer.py
@@ -88,7 +88,7 @@ class Sequencer(object):
 
         Args:
             masks_fmap (np.ndarray): 3D fieldmap mask used for the optimizer to shim in the region
-                                             of interest (only consider voxels with non-zero values)
+                                     of interest (only consider voxels with non-zero values)
         Returns:
                 np.ndarray: Coefficients of the coil profiles to shim (len(slices) x n_channels)
         """
--- a/shimmingtoolbox/shim/sequencer.py
+++ b/shimmingtoolbox/shim/sequencer.py
@@ -82,12 +82,12 @@ class Sequencer(object):
         self.path_output = path_output
         self.optimizer = None
 
-    def optimize(self, masks_fmap_dilated):
+    def optimize(self, masks_fmap):
         """
         Optimization of the currents for each shim group. Wraps :meth:`shimmingtoolbox.shim.sequencer.Sequencer.opt`.
 
         Args:
-            masks_fmap_dilated (np.ndarray): 3D fieldmap mask used for the optimizer to shim in the region
+            masks_fmap (np.ndarray): 3D fieldmap mask used for the optimizer to shim in the region
                                              of interest (only consider voxels with non-zero values)
         Returns:
                 np.ndarray: Coefficients of the coil profiles to shim (len(slices) x n_channels)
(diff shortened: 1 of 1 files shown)


shimmingtoolbox/shim/sequencer.py
```diff
@@ -327,8 +327,8 @@ def get_resampled_masks(self):
                 for i in range(n_shims))
 
             # We need to transpose the mask to have the good dimensions
-            resampled_mask = np.array([results_mask[it][0].get_fdata() for it in range(n_shims)]).transpose(1, 2, 3, 0)
-            masks_fmap = np.array([results_mask[it][1].get_fdata() for it in range(n_shims)]).transpose(1, 2, 3, 0)
+            resampled_mask = np.array([results_mask[it][1].get_fdata() for it in range(n_shims)]).transpose(1, 2, 3, 0)
```
po09i: Can you give more descriptive names to the variables? What is resampled_mask? How is it different from masks_fmap?

Aurelienpujolm: We can probably go with resampled_mask_dilated or something like that ? 
I think that we should probably go with another PR after this one to fixed all of the confused names in the code !

po09i: I suggest improving the ones you actively interact with in the PR you are working. I would put a dedicated PR for that as low priority but we can keep it in the back of our minds when we do some changes to improve the codebase.

Aurelienpujolm: So for this PR, is masks_fmap_dilated, and masks_fmap good for you ?

shimmingtoolbox/shim/sequencer.py
```diff
@@ -82,12 +82,12 @@ def __init__(self, slices, mask_dilation_kernel, mask_dilation_kernel_size, reg_
...
         self.optimizer = None
 
-    def optimize(self, resampled_mask):
+    def optimize(self, masks_fmap_dilated):
         """
         Optimization of the currents for each shim group. Wraps :meth:`shimmingtoolbox.shim.sequencer.Sequencer.opt`.
 
         Args:
-            resampled_mask (np.ndarray): 3D anat mask used for the optimizer to shim in the region of interest.
+            masks_fmap_dilated (np.ndarray): 3D anat mask used for the optimizer to shim in the region of interest.
```
po09i: Is this right, the name says that it is the fmap mask and the description says that it is the anat

po09i (changes requested):

shimmingtoolbox/shim/sequencer.py
Aurelienpujolm: good catch, I will fixed this

shimmingtoolbox/shim/sequencer.py
```diff
@@ -82,13 +82,13 @@ def __init__(self, slices, mask_dilation_kernel, mask_dilation_kernel_size, reg_
...
 
-    def optimize(self, resampled_mask):
+    def optimize(self, masks_fmap_dilated):
         """
         Optimization of the currents for each shim group. Wraps :meth:`shimmingtoolbox.shim.sequencer.Sequencer.opt`.
 
         Args:
-            resampled_mask (np.ndarray): 3D anat mask used for the optimizer to shim in the region of interest.
-                                             (only consider voxels with non-zero values)
+            masks_fmap_dilated (np.ndarray): 3D fieldmap mask used for the optimizer to shim in the region
```
```suggestion
            masks_fmap (np.ndarray): 3D fieldmap mask used for the optimizer to shim in the region
```

shimmingtoolbox/shim/sequencer.py
```diff
@@ -82,13 +82,13 @@ def __init__(self, slices, mask_dilation_kernel, mask_dilation_kernel_size, reg_
         self.path_output = path_output
         self.optimizer = None
 
-    def optimize(self, resampled_mask):
+    def optimize(self, masks_fmap_dilated):
```
```suggestion
    def optimize(self, masks_fmap):
```

shimmingtoolbox/shim/sequencer.py
```diff
@@ -97,12 +97,12 @@ def optimize(self, resampled_mask):
         coefs = []
         for i in range(n_shims):
             # If there is nothing to shim in this shim group
-            if np.all(resampled_mask[..., i] == 0):
+            if np.all(masks_fmap_dilated[..., i] == 0):
```
```suggestion
            if np.all(masks_fmap[..., i] == 0):
```

shimmingtoolbox/shim/sequencer.py
```diff
@@ -97,12 +97,12 @@ def optimize(self, resampled_mask):
...
                 coefs.append(np.zeros(self.optimizer.merged_coils.shape[-1]))
 
             # Otherwise optimize
             else:
-                coefs.append(self.optimizer.optimize(resampled_mask[..., i]))
+                coefs.append(self.optimizer.optimize(masks_fmap_dilated[..., i]))
```
```suggestion
                coefs.append(self.optimizer.optimize(masks_fmap[..., i]))
```

po09i (changes requested):

po09i (approved):

PR #462, opened 2023-06-27
Improve real-time shimming

po09i:
## Description
This PR aims at improving real-time shimming functionalities. More precisely,
- It improves a plot showing the variation of the pressure probe along with the variation of the B0 field
- Used RMSE instead of the MAE (sum of the absolute values) for some other plots
- Output plots for different slices in a separate directory rather than a single long figure


shimmingtoolbox/pmu.py
```diff
@@ -123,6 +123,18 @@ def read_resp(self, fname_pmu):
...
         return attributes
 
+    def get_times(self):
+        """
+        Get the times in ms at which the respiration took place.
+
+        Returns:
+            np.ndarray: Array containing the timepoints in ms of each data
+        """
+        raster = float(self.stop_time_mdh - self.start_time_mdh) / (len(self.data)-1)
```
4rnaudB, line 133: Why -1?

shimmingtoolbox/shim/sequencer.py
```diff
@@ -1371,44 +1387,102 @@ def plot_static_riro(self, masked_unshimmed, masked_shim_static, masked_shim_sta
...
+
+        # Get the b0 field in the same units as the pressure reading
+        n_plots = len(self.index_shimmed)
+        max_diff_field = 0
+        for i_plot in range(n_plots):
+            diff_field = curated_unshimmed_trace[i_plot].max() - curated_unshimmed_trace[i_plot].min()
+            if abs(max_diff_field) < abs(diff_field):
+                max_diff_field = diff_field
+                min_field = curated_unshimmed_trace[i_plot].min()
+                max_field = curated_unshimmed_trace[i_plot].max()
```
4rnaudB: I think you can simply get the list with the max and min value like this:
```
max_diff_field _list= max(curated_unshimmed_trace, key=lambda x: abs(x.max() - x.min()))
min_field = max_diff_field_list.min()
max_field = max_diff_field_list.max()
max_diff_field = max_field - min_field
```

4rnaudB: To avoid unnecessary for loops

shimmingtoolbox/shim/sequencer.py
```diff
@@ -1371,44 +1387,102 @@ def plot_static_riro(self, masked_unshimmed, masked_shim_static, masked_shim_sta
...
+
+        diff_pressure = pmu_pressures_curated.max() - pmu_pressures_curated.min()
+        scaling = max_diff_field / diff_pressure
+        avg_pressure = np.mean(pmu_pressures_curated)
+
+        curated_unshimmed_trace_scaled = np.zeros_like(curated_unshimmed_trace)
+        for i_plot in range(n_plots):
+            avg_b0field = np.mean(curated_unshimmed_trace[i_plot])
+            curated_unshimmed_trace_scaled[i_plot] = (curated_unshimmed_trace[
+                                                          i_plot] - avg_b0field) / scaling + avg_pressure
```
4rnaudB:
```suggestion
        curated_unshimmed_trace_scaled = np.array([(x - np.max(x)) / scaling + avg_pressure for x in curated_unshimmed_trace])
```

4rnaudB: depends if n_plot is the same length as curated_unshimmed_trace

shimmingtoolbox/shim/sequencer.py
```diff
@@ -1564,20 +1638,21 @@ def plot_full_time_std(self, unshimmed, masked_shim_static_riro, mask_fmap_cs, m
...
 
         # Transpose is used to cater to numpy division order
         # (3, 2, 4) / (3, 2) Does not work
         # (4, 2, 3) / (2, 3) Does work
-        #* Using out parameter in np.divide() prevents inconsistent results
+        # * Using out parameter in np.divide() prevents inconsistent results
         shimmed_masked = np.zeros(mask_extended.shape)
-        np.divide(np.sum(masked_shim_static_riro, axis=-1).T, sum_mask_fmap_cs.T, where=mask.T.astype(bool), out=shimmed_masked.T)
+        np.divide(np.sum(masked_shim_static_riro, axis=-1).T, sum_mask_fmap_cs.T, where=mask.T.astype(bool),
+                  out=shimmed_masked.T)
```
4rnaudB, line 1648: Note: very good to get the output of the divide in its `out` parameter. I've had problems before doing `res = np.divide()`.

4rnaudB: (it is unreliable)

4rnaudB: Actually, this is my code ^^. This exact line would give different results from time to time by doing `res = np.divide()`

shimmingtoolbox/pmu.py
po09i: `raster` gives the amount of time between 2 time points, if there is 10ms between timepoints, stop - start would give 90ms for 10 timepoints, so we need the -1.

shimmingtoolbox/shim/sequencer.py
po09i: It is, I used np.mean but otherwise changed it to your suggestion in: [b8f20da](https://github.com/shimming-toolbox/shimming-toolbox/pull/462/commits/b8f20da1744efa41af3bf649551277518322d6a8)

shimmingtoolbox/shim/sequencer.py
po09i: Changed it here: [b8f20da](https://github.com/shimming-toolbox/shimming-toolbox/pull/462/commits/b8f20da1744efa41af3bf649551277518322d6a8)

4rnaudB (approved):

PR #463, opened 2023-06-28
Shimmed field map plotting for real time shimming by averaging over time dimension

4rnaudB:
## Checklist

## Description

- Correction of an error for `shimmingtoolbox/shim/sequencer.py at line 931` where the return order was opposite to what was expected in the code

- Adaptation of ShimSequencer.plot_full_mask() for RealTimeSequencer. The shimmed field map was averaged over the fourth dimension (time) and fifth (not sure what it is @po09i). Here is the current state result from a subject using coil + first order real time shimming (results seem too good):

image

### DIFF
--- a/shimmingtoolbox/shim/sequencer.py
+++ b/shimmingtoolbox/shim/sequencer.py
@@ -1523,7 +1523,6 @@ class RealTimeSequencer(Sequencer):
         fig.subplots_adjust(top=0.85)
 
         # Save
-        print("---------------------------------------------------------------- \n SAVING")
         fname_figure = os.path.join(self.path_output, 'fig_shimmed_vs_unshimmed.png')
         fig.savefig(fname_figure, bbox_inches='tight')
 
--- a/shimmingtoolbox/shim/sequencer.py
+++ b/shimmingtoolbox/shim/sequencer.py
@@ -1479,8 +1479,7 @@ class RealTimeSequencer(Sequencer):
         mt_unshimmed = montage(unshimmed_avg)
         mt_unshimmed_masked = montage(unshimmed_avg * mask)
         mt_shimmed_masked = montage(shimmed_masked_avg)
-        print("----------------------------------------------------------------")
-        print(mt_unshimmed.shape, mt_shimmed_masked.shape, mt_shimmed_masked.shape)
+        
         metric_unshimmed_std = calculate_metric_within_mask(unshimmed_avg, mask, metric='std')
         metric_shimmed_std = calculate_metric_within_mask(shimmed_masked_avg, mask, metric='std')
         metric_unshimmed_mean = calculate_metric_within_mask(unshimmed_avg, mask, metric='mean')
--- a/shimmingtoolbox/shim/sequencer.py
+++ b/shimmingtoolbox/shim/sequencer.py
@@ -1248,10 +1248,11 @@ class RealTimeSequencer(Sequencer):
             i_shim += 1
             if i_shim >= n_shim - 1:
                 break
-
+        
         if logger.level <= getattr(logging, 'DEBUG') and self.path_output is not None:
             # Plot before vs after shimming averaged on PMU signal and time
-            self.plot_full_mask(unshimmed, masked_shim_static_riro, mask_full_binary)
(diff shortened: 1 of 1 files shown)


4rnaudB:
The plotting function had to be moved to DEBUG only mode since some tests for real time shimming don't provide the output path required to save the figure. The problem is DEBUG mode is not accessible from the GUI. If we don't want to generate the figure automatically, we can keep the tests like this, but if we want to generate it everytime real time shimming is run (like in dynamic shimming) we should maybe modify the test to give an output path or make the default output path different than None?

po09i:
If the figure is a figure that should be output to the user every time, we can move it outside of the `DEBUG` condition. We should still wrap it within a `path_output` condition. The reasoning is that if sequencer is used as an API, we do not necessarily want to output figures, but specifying a path would write it to a file. When using the CLI, the default path exists and we should therefore have a oath to write to.

po09i:
> The shimmed field map was averaged over the fourth dimension (time) and fifth (not sure what it is @po09i).

The 4th dimension is indeed time while the 5th dimension is each shim group (for slice-wise shimming, there would be one shim group per slice). Therefore the 5 dimension array tells you the 3D corrected volume for each time points and for each shim group.

4rnaudB:
The equivalent figure is plotted by default when using dynamic shimming

4rnaudB:
> The 4th dimension is indeed time while the 5th dimension is each shim group (for slice-wise shimming, there would be one shim group per slice). Therefore the 5 dimension array tells you the 3D corrected volume for each time points and for each shim group.

So does it make sense to average the 5th dimension to get an average corrected field map?

4rnaudB:
I summed the fifth dimension considering the weighed mask and kept the average for the time dimension. New results for first order real time shimming:
image

po09i:
There is an issue with averaging the 5th dimension. It does not tell the full story. We are acquiring a fieldmap that we expect to vary by a small amount between timepoints. The role of the realtime correction is to reduce the variation between different timepoints. By averaging each timepoints, we lose this information. I think we also need a map that describes by how much each points change over the time dimension (std for example).

4rnaudB:
Yes, Julien asked to plot a heatmap of the STD over time to show the time correction effect. If we want, to compare to dynamic shimming, should we have an option to plot this for dynamic shimming as well if the field map is 4D?

po09i:
> should we have an option to plot this for dynamic shimming as well if the field map is 4D?

I don't think we should have this for non real-time shimming.

4rnaudB:
@po09i The inconsistency issue is fixed, we just need to decide when and where we should plot these figures.

po09i:
I think if they do not take too long, we should always output them as they are what tells how good the shim is. As for where, they should be in "path_output" and ideally be calculated after the text file is written.

4rnaudB:
> The plotting function had to be moved to DEBUG only mode since some tests for real-time shimming don't provide the output path required to save the figure

So I should make change the sequencer tests to have them set an output path?

po09i:
The code looks good!, how long does it take to generate and plot the figures that you added?

po09i:
> So I should make change the sequencer tests to have them set an output path?

I would do as you did and output them if an `output_path` is provided

4rnaudB:
> how long does it take to generate and plot the figures that you added?
image
image

shimmingtoolbox/shim/sequencer.py
```diff
@@ -1454,6 +1466,138 @@ def print_rt_metrics(self, unshimmed, shimmed_static, shimmed_static_riro, shimm
...
+        ax.get_yaxis().set_visible(False)
+        divider = make_axes_locatable(ax)
+        cax = divider.append_axes('right', size='5%', pad=0.05)
+        fig.colorbar(im, cax=cax)
+
+        # Lower suptitle
+        fig.subplots_adjust(top=0.85)
+
+        # Save
+        print("---------------------------------------------------------------- \n SAVING")
```
```suggestion
```

If you want to print something, use `logger.info("")`

po09i: However in this case I would not output anything since there could be multiple plots that could output "saving".

po09i (changes requested):

po09i (approved):

PR #465, opened 2023-07-07
Fix an issue where real-time shimming would not work using pseudo-inverse

po09i:
## Description
Since PR #451, a bug was introduced that made real-time shimming crash when using the pseudo-inverse optimizer. This PR fixes that bug and adds a test to make sure pseudo-inverse does not crash real-time shimming.

## Linked issues
#451


Aurelienpujolm (approved):
It's all good for me!

4rnaudB (approved):

PR #466, opened 2023-07-10
Add mp2rage, and more field maps to dcm2bids config file and update to version >= 2.1.7

po09i:
## Description
Add MP2RAGE to the automatically detected acquisitions in the dcm2bids config file. Also support more field maps. This PR also updates the dcm2bids version to be >= 2.1.7. This was required since the patch allowing to use dcm2bids on LAN was introduced in this [PR](https://github.com/UNFmontreal/Dcm2Bids/pull/154) which is on version 2.1.7.

### DIFF
--- a/config/dcm2bids.json
+++ b/config/dcm2bids.json
@@ -101,6 +101,35 @@
             }
         },
         {
+            "dataType": "anat",
+            "modalityLabel": "mp2rage_inv1",
+            "criteria": {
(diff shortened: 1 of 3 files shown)


4rnaudB (commented):
After `make install`, I get this 
image

Was woking before `make install`

4rnaudB (approved):

PR #469, opened 2023-08-08
Add windows compatibility

po09i:
## Description
This PR add windows compatibility to the toolbox

In more details:

- Adds a .bat script to install Shimming Toolbox
- Make modifications to ensure "all" features work (All tests are passing except prelude and sct tests since I have not installed them) 
- Solve unwrapper issue (We only natively support prelude which seems to only be installable on linux/Macs, #473 implemented skimage so we have an unwrapper installed)
- Add windows installation section

A future PR should add windows installation and testing in the CI


4rnaudB:
st_b0shim dynamic works well. Create field maps doesn't work and I couldn't do the create coil profiles tutorial because of the bash commands

po09i:
Thanks for testing it out!

> st_b0shim dynamic works well. Create field maps doesn't work and I couldn't do the create coil profiles tutorial because of the bash commands

I believe prelude does is not available on windows (FSL recommends installing it through WSL). Therefore I would not expect it to work, is that the error message you get? Good point about the tutorial, they will need to be updated.

docs/source/user_section/installation/install_on_windows.rst
```diff
@@ -0,0 +1,140 @@
+***********************
```
4rnaudB, line 1: I would mention possible problems with conda installation if the path has spaces in it.

shimmingtoolbox/cli/check_env.py
```diff
@@ -83,7 +83,10 @@ def check_prelude_installation():
         bool: True if prelude is installed, False if not.
     """
     try:
-        subprocess.check_call(['which', 'prelude'], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+        if sys.platform != 'win32':
+            subprocess.check_call(['which', 'prelude'], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+        else:
+            subprocess.check_call(['where', 'prelude'], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
```
4rnaudB: Would switch the if to == 'win32', else

shimmingtoolbox/cli/check_env.py
```diff
@@ -103,9 +106,12 @@ def check_dcm2niix_installation():
         bool: True if dcm2niix is installed, False if not.
     """
     try:
-        subprocess.check_call(['which', 'dcm2niix'], stdout=subprocess.DEVNULL,
-                              stderr=subprocess.DEVNULL)
+        if sys.platform != 'win32':
+            subprocess.check_call(['which', 'dcm2niix'], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+        else:
+            subprocess.check_call(['where', 'dcm2niix'], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
```
4rnaudB: idem

shimmingtoolbox/unwrap/skimage_unwrap.py
```diff
@@ -57,10 +57,22 @@ def skimage_unwrap(nii_wrapped_phase, mag=None, mask=None, threshold=None, fname
...
-    # Unwrap the phase
-    ma_wrapped_phase = ma.array(wrapped_phase, mask=~unwrap_mask)
+    # Mask the wrapped phase
+    ma_wrapped_phase = ma.array(wrapped_phase, mask=~unwrap_mask.astype(bool))
+
+    singleton = False
+    if ma_wrapped_phase.ndim == 3 and ma_wrapped_phase.shape[2] == 1:
+        # If the phase is 3D with a single slice, remove the last dimension
+        ma_wrapped_phase = ma_wrapped_phase[:, :, 0]
+        singleton = True
```
4rnaudB: Why not add the line from if singleton here instead?

shimmingtoolbox/unwrap/skimage_unwrap.py
```diff
@@ -57,10 +57,22 @@ def skimage_unwrap(nii_wrapped_phase, mag=None, mask=None, threshold=None, fname
...
+
+    # Unwrap the masked phase
     ma_unwrapped_phase = unwrap_phase(ma_wrapped_phase, rng=0)
 
+    if singleton:
+        # If the phase was 3D with a single slice, add the last dimension back
+        ma_unwrapped_phase = ma_unwrapped_phase[:, :, np.newaxis]
```
4rnaudB: Previous comment

docs/source/user_section/installation/install_on_windows.rst
po09i: Done here: 18c7d78d82f45626ff350eda5b758557ffb63187

shimmingtoolbox/cli/check_env.py
po09i: Done here: 18c7d78d82f45626ff350eda5b758557ffb63187

shimmingtoolbox/cli/check_env.py
po09i: Done here: 18c7d78d82f45626ff350eda5b758557ffb63187

shimmingtoolbox/unwrap/skimage_unwrap.py
po09i: I did an if else

po09i: Done here: 18c7d78d82f45626ff350eda5b758557ffb63187

shimmingtoolbox/unwrap/skimage_unwrap.py
po09i: Done here: 18c7d78d82f45626ff350eda5b758557ffb63187

4rnaudB (approved):

PR #473, opened 2023-08-15
Implement skimage's phase unwrapper

po09i:
## Description
This PR adds a phase unwrapper that is bundled with shimming toolbox (skimage).

- Adds a wrapper to call skimage's [unwrap_phase](https://scikit-image.org/docs/stable/api/skimage.restoration.html#skimage.restoration.unwrap_phase) function.
- Adds the ability to call it from the command line
- Add skimage_unwrap_phase specific tests
- Changed most prelude calls this new unwrapper to minimize the tests that required prelude as a dependency

## Linked issues
Related to #45


po09i:
Example in the brain of prelude vs skimage:
skimage_vs_prelude

More tests should be done but initially it seems promising on good quality phase maps

shimmingtoolbox/prepare_fieldmap.py
```diff
@@ -29,7 +29,7 @@ def prepare_fieldmap(list_nii_phase, echo_times, mag, unwrapper='prelude', mask=
                                The values must range from [-pi to pi].
         echo_times (list): List of echo times in seconds for each echo. The number of echotimes must match the number of
                            echoes. It input is a phasediff (1 phase), input 2 echotimes.
-        unwrapper (str): Unwrapper to use for phase unwrapping. Supported: prelude.
+        unwrapper (str): Unwrapper to use for phase unwrapping. Supported: prelude, skimage.
```
4rnaudB: In unwrap_phase.py it's written `Supported: ``prelude``, ``skimage`` `

test/cli/test_cli_prepare_fieldmap.py
```diff
@@ -110,23 +114,25 @@ def test_cli_prepare_fieldmap_default_output():
     os.remove(os.path.join(os.curdir, 'fieldmap.json'))
 
 
-@pytest.mark.prelude
 def test_cli_prepare_fieldmap_gaussian():
```
4rnaudB, line 117: Is there no test using prelude? If yes I would specify the unwrapper in the test name

4rnaudB: This applies to all tests below

test/test_prepare_fieldmap.py
```diff
@@ -28,12 +28,22 @@ def setup(self):
 
     def test_prepare_fieldmap_1_echo(self):
```
4rnaudB, line 29: Same here, the test below has prelude in its name but skimage doesn't appear here.

4rnaudB (commented):
`skimage` vs `prelude`:
1- Output messages (`skimage` gets warning, not `prelude`)
image
2- Get significantly different field maps (using -100 to 100):
`prelude`:
image
`skimage`:
image
Note the right side and the extra black pixel (bottom right) in `skimage`. 

On a seconde test, it was less notable:
image
`Prelude`:
image
`skimage`:
image

test/cli/test_cli_prepare_fieldmap.py
po09i: The idea is to test the field mapping pipeline with as little external dependency as possible. So the specific unwrapper is not important for the test, but I have to specify skimage since prelude remains the default one.

test/test_prepare_fieldmap.py
po09i: Similar answer to my comment above, I specified prelude in the test below because I'm testing for prelude specifically in that test (it's the only test using prelude in `test_prepare_fieldmap`

shimmingtoolbox/prepare_fieldmap.py
po09i: Fixed [4b37af6](https://github.com/shimming-toolbox/shimming-toolbox/pull/473/commits/4b37af6aa7352f7dc179dcb14712734ccfe2dfe6)

po09i:
> 2- Get significantly different field maps (using -100 to 100):

Indeed skimage's unwrapper seems to struggle a bit more with noise and with separated regions.

> 1- Output messages (skimage gets warning, not prelude)

This is a sign that skimage performed poorly on that field map

4rnaudB (approved):

PR #477, opened 2023-08-28
Update nibabel to latest version and allow to input a scaled threshold

po09i:
## Description
- Update nibabel to the latest version
- Add the ability to scale the threshold when masking

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,7 +41,7 @@ setup(
         'importlib-metadata ~= 4.0 ; python_version < "3.8"',
         "numpy>=1.21",
         "phantominator~=0.6.4",
-        "nibabel~=3.2.1",
+        "nibabel>=3.2.1",
         "requests",
         "scipy>=1.7",
         "tqdm",
--- a/shimmingtoolbox/cli/mask.py
+++ b/shimmingtoolbox/cli/mask.py
@@ -1,13 +1,13 @@
 #!/usr/bin/env python3
 
 import click
+import logging
 import nibabel as nib
 import numpy as np
 import os
-import logging
 
-import shimmingtoolbox.masking.threshold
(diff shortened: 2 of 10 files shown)


4rnaudB (commented):
When using --scaled-thr, there is no verification to make sure the threshold is between 0 and 1 or between 0 and 100 and scaling it between 0 and 1. Instead, no error is thrown and a empty mask is created. 

I tried to mask using a anatomical ranging from 0 to 1163.52 using the following commands:
image

With the following results:
half the range with scaled flag:
image

Half the range no scale flag (581.76)
image

There are some small differences, I'm not sure if it only comes from a rounding difference while using the threshold flag.

po09i:
> When using --scaled-thr, there is no verification to make sure the threshold is between 0 and 1 or between 0 and 100 and scaling it between 0 and 1.

Good catch, there was something in place to catch that, but it was set up incorrectly. I fixed it in [dfe3bfb](https://github.com/shimming-toolbox/shimming-toolbox/pull/477/commits/dfe3bfbe4aed2a61ba3686c614ee53d965f42ab4).

> I tried to mask using a anatomical ranging from 0 to 1163.52 using the following commands...

I tried it on my hand and found out that for some reason the maximum value is not what is reported in fsleyes. This is what explains the differences.

Screenshot 2023-08-29 at 3 02 42 PM
Screenshot 2023-08-29 at 3 04 32 PM

4rnaudB:
Fsleyes seems to be in the wrong, as seen in the image below, the data type of the nifti file is INT16, which makes a maximum value of 1163.2 questionable.
image

Using numpy's half of maximum value gives the same result as the flag

4rnaudB (approved):

PR #480, opened 2023-08-31
Resolve quadrog issue when the sum of the channels constraint was infinite

### LINKED ISSUE #478: Bug for the quadprog optimizer
### Checklist

- [x] I've given this issue a self-descriptive, meaningful title (e.g. *Installation failure: problem creating launchers*)
- [x] I've described the bug

In scenarios involving multiple coils for shimming, a potential bug can arise if some coils have a coef_sum_max set to np.inf, while others do not. One possible dirty  approach to address this issue, could be to remove the if statement within the for loop where we check whether coil.coef_sum_max differs from np.inf. Following this, you could keep the subsequent lines within the loop. Additionally, it might be advisable to omit the case involving np.all(sum_constraints == 0) and maintain the version containing the else clause for both real-time and dynamic shimming.

Another way to tackle this could involve removing the lines from sum_constraints where coef.sum_max = np.inf, to not have a line of sum_constraints with zeros. However, careful consideration is required to ensure that sum_constraints does not end up being empty (particularly if all the coils have no restrictions on the sum of currents).

po09i:
## Description
This PR fixes an issue where quadprog would set the sum of the channels constraints to 0 if they were infinite.

## Linked issues
Fixes #478

### DIFF
--- a/shimmingtoolbox/cli/b0shim.py
+++ b/shimmingtoolbox/cli/b0shim.py
@@ -494,9 +494,8 @@ def _save_to_text_file_static(coil, coefs, list_slices, path_output, o_format, o
 @click.option('--scanner-coil-order', type=click.Choice(['-1', '0', '1', '2']), default='-1', show_default=True,
               help="Maximum order of the shim system. Note that specifying 1 will return "
                    "orders 0 and 1. The 0th order is the f0 frequency.")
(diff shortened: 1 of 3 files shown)


Aurelienpujolm (approved):
All good!

Issue #471, opened 2023-08-09
Division of shim optimization between static/dynamic and real-time 


4rnaudB:
### Checklist

- [x] I've given this issue a self-descriptive, meaningful title (e.g. *Document the use of config/dcm2bids.json*)
- [x] I've described the enhancement for an existing feature
- [x] (optional) I've added some ideas about how to implement my suggestion
- [x] I've given this issue the correct labels

### Description of Problem / Inefficiency:

Currently, the optimization parameters (coils, scanner order and sequence) are common to both static/dynamic and real-time. This limits experiment possibilities. For example, you cannot obtain coefficients for a static 2nd-order shim with a 0th-order real-time z-shim. Also, you cannot calculate coefficients for a static/dynamic coil shim and a real-time scanner shim. 

### Implementation Suggestion:

In sequencer.py line 352, the following function selects the optimizer:
```
    if self.method in supported_optimizers:
          if self.method == 'least_squares':
              self.optimizer = supported_optimizers[self.method](self.coils, unshimmed, affine, self.opt_criteria,
                                                                 reg_factor=self.reg_factor)

          elif self.method == 'least_squares_rt':
              # Make sure pmu is defined
              if pmu is None:
                  raise ValueError(f"pmu parameter is required if using the optimization method: {self.method}")

              # Add pmu to the realtime optimizer(s)
              self.optimizer_riro = supported_optimizers[self.method](self.coils, unshimmed, affine,
                                                                      self.opt_criteria, pmu,
                                                                      reg_factor=self.reg_factor)
```
The first parameter of the optimizer (`self.coil`) is the same for both `optimizer `and `optimizer_riro`.  This parameter comes from b0shim.py (`list_coils = _load_coils(coils, scanner_coil_order, fname_sph_constr, nii_fmap, initial_coefs, json_fm_data['Manufacturer']))`). A possible implementation would be to duplicate every parameter making up `list_coils` : coils, scanner_coil_order to have 2 `list_coils`, one for real-time and one for static/dynamic. A duplicate of `list_slices` could also be made to allow different sequences. Using [click group](https://click-option-group.readthedocs.io/en/latest/tutorial.html#a-simple-example) or [cloup](https://github.com/janLuke/cloup) option would make it easier for the user to understand.

PR #484, opened 2023-09-18
Split coils and scanner order for optimization + allow specific scanner orders for each

4rnaudB:
## Checklist

- [x] I've updated the documentation and/or added the correct docstrings

## Description
PR to split the coefficients' optimization between static and RIRO to allow users to optimize both coefficient types with different coils and scanner order. This PR will also implement the possibility of selecting specific scanner orders to use instead of using every scanner order below the selected one. 

## Linked issues
#471 

## Missing elements

- [x] Tests 
- [x] Coil profile creation to follow dictionary rules (from CAD and field maps)
- [x] Change test files 
- [x] Change txt files output
- [x] Adapt GE


shimmingtoolbox/cli/b0shim.py
```diff
@@ -31,13 +31,24 @@
...
              help="Shim according to the specified algorithm as an argument e.g. st_b0shim xxxxx")
 def b0shim_cli():
     pass
 
+def parse_orders(orders:str):
+    orders = orders.split(',')
+    try:
+        orders = [int(order) for order in orders]
+        orders.sort()
+        if any(order<-1 or order>2 for order in orders):
```
po09i: Use `AVAILABLE_ORDERS` to check so that if we add order 3, we only have to change that that list rather than the parse_order function ad well "`order in AVAILABLE_ORDER`"

shimmingtoolbox/cli/b0shim.py
```diff
@@ -31,13 +31,24 @@
 CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
-
+AVAILABLE_ORDERS = [-1, 0, 1, 2]
```
po09i, line 35:
```suggestion


```

shimmingtoolbox/cli/b0shim.py
```diff
@@ -685,111 +721,182 @@ def realtime_dynamic(fname_fmap, fname_anat, fname_mask_anat_static, fname_mask_
...
-    end_channel = 0
-    for i_coil, coil in enumerate(list_coils):
+    #TODO change so that every coil can be visualized
+    # Get common coils between static and riro
+    coil_static_only = [coil for coil in list_coils_static if coil not in list_coils_riro]
+    coil_riro_only = [coil for coil in list_coils_riro if coil not in list_coils_static]
+    list_coils_common = [coil for coil in list_coils_static if coil in list_coils_riro]
+    all_coils = list_coils_common + coil_static_only + coil_riro_only
+
+    index = 0
```
po09i, line 726: Add comments mentioning what you are doing

shimmingtoolbox/cli/b0shim.py
```diff
@@ -685,111 +721,182 @@ def realtime_dynamic(fname_fmap, fname_anat, fname_mask_anat_static, fname_mask_
...
         # If it's a scanner
         if type(coil) == ScannerCoil:
-            manufacturer = json_anat_data['Manufacturer']
-            # If outputting in the gradient CS, it must be the 1st order and must be in the delta CS and Siemens
-            # The check has already been done earlier in the program to avoid processing and throw an error afterwards.
-            # Therefore, we can only check for the o_format_sph.
-            if o_format_sph == 'gradient':
-                logger.debug("Converting scanner coil from Shim CS to Gradient CS")
+            for key in [str(order) for order in AVAILABLE_ORDERS
+                        if (order != -1 and (str(order) in coil_indexes_riro[coil.name]
```
po09i, line 758: Is there a problem if the name of the coil is the same for 2 different coils? Maybe we should add a check that all coil names must be different.

shimmingtoolbox/coils/siemens_basis.py
po09i, line 1: this file should not exist anymore, `siemens_basis` is now part of `sphere_harm_basis`

shimmingtoolbox/coils/spher_harm_basis.py
```diff
@@ -52,14 +52,20 @@ def siemens_basis(x, y, z, orders=(1, 2), shim_cs='LAI'):
...
     # Reorder according to siemens convention: X, Y, Z, Z2, ZX, ZY, X2-Y2, XY
-    reordered_spher = _reorder_to_siemens(spher_harm)
+    reordered_spher = _reorder_to_siemens(spher_harm, orders)
 
     # Select order
-    range_per_order = {1: list(range(3)), 2: list(range(3, 8))}
+    range_per_order = {}
+    index = 0
+    for order in orders:
+        range_per_order[order] = list(range(index, index+(order*2 +1)))
```
po09i: Siemens' order 3 does not contain all 7 channels (I think they did not implement 2 of them). Therefore I'm a bit reluctant to modifying everything based on `order * 2 + 1`since it will break with order 3.

shimmingtoolbox/coils/spher_harm_basis.py
```diff
@@ -330,7 +349,7 @@ def _check_basis_inputs(x, y, z, orders):
         raise NotImplementedError("Spherical harmonics not implemented for order 3 and up")
 
 
-def get_flip_matrix(shim_cs='ras', manufacturer=None, xyz=False):
+def get_flip_matrix(shim_cs='ras', orders=(1,2), manufacturer=None, xyz=False):
```
po09i: Since you added orders as an option, `xyz` does not make sense anymore (equivalent to orders=(1,))

shimmingtoolbox/shim/shim_utils.py
```diff
@@ -175,10 +176,10 @@ def phys_to_shim_cs(coefs, manufacturer):
         flip_mat = np.ones(len(coefs))
         # Order 1
         if len(coefs) == 3:
-            flip_mat[:3] = get_flip_matrix(shim_cs[manufacturer], manufacturer=manufacturer, xyz=True)
+            flip_mat[:3] = get_flip_matrix(shim_cs[manufacturer], orders, manufacturer=manufacturer, xyz=True)
         # Order 2
         elif len(coefs) >= 8:
-            flip_mat[:8] = get_flip_matrix(shim_cs[manufacturer], manufacturer=manufacturer)
+            flip_mat[:8] = get_flip_matrix(shim_cs[manufacturer], orders, manufacturer=manufacturer)
```
po09i, line 182: same comment as above, if orders is (1,), it is redundant to `xyz`

po09i (commented):
In terms of style, try to have spaces between operations (`2 * 2 `rather than `2*2`).

For GE, you can do something very similar to what you have done for Siemens.

shimmingtoolbox/cli/b0shim.py
4rnaudB: There would be a problem yes, but the verification should probably be done at the input level and not once al the shimming is done right?

4rnaudB: fixed in [7f7c64f](https://github.com/shimming-toolbox/shimming-toolbox/pull/484/commits/7f7c64f6620861b8e7a78da30c8c6353ae88b4bb)

po09i: I see you made a check based on if the coil has the same path. However, if the "name" of the coil is the same, there would still be an error (`coil.name` would refer to the same coil though they are different coils)

4rnaudB: not sure what you mean because my test verifies if some coils have the same name

4rnaudB:
```
if len(coils) != len(set(coils)):
        raise ValueError("Coils must be unique. Make sure different coils have different names.")
```
Is a verification based on coil names

po09i (commented):
We should have an example config file for custom coils as well now that they are defined differently.

We should have the same behaviour for the --scanner-coil-order-riro (something as --scanner-coil-order when not defined) and for --coil-riro (same thing as --coil when not defined). I do not believe this is the case.. This has me thinking, what do we do if we do not want to shim static, or if we do not want to shim riro? What would be the inputs?

shimmingtoolbox/cli/b0shim.py
po09i: This is a verification based on the file name/path. However, there is a `name` key in the .json file that could still be the same even if they have different paths.
Screenshot 2023-10-11 at 5 06 39 PM

4rnaudB: True, don't know why I put it there, now fixed

4rnaudB:
>what do we do if we do not want to shim static, or if we do not want to shim riro? What would be the inputs?

If you don't want to shim riro, you use dynamic shim. If you don't want to shim static, it's not a possibility right now, but when would you not want to shim static?

po09i:
The was a problem for coils other than the Prisma, I modified it and it should work now. However, I suggest testing for them by modifying the `manufacturer_model_name` tag of the fmap sidecar to something other than the Prisma. You can also test for GE or Philips using this method and making sure the coefficients are the same as what you would get on master. We should probably test that by modifying the sidecar and simulate data from other manufacturers.

shimmingtoolbox/coils/coil.py
```diff
@@ -19,6 +20,7 @@
 
 
 class Coil(object):
+    # ! Modify description if everything works
```
po09i: Description was modified?

shimmingtoolbox/coils/coil.py
```diff
@@ -134,45 +143,58 @@ def __init__(self, dim_volume, affine, constraints, order, manufacturer=""):
...
+    def __hash__(self):
+        return hash(self.name)
+
+    def __eq__(self, __value: object) -> bool:
+        return super().__eq__(__value)
 
-def get_scanner_constraints(manufacturers_model_name, order=2):
+
+def get_scanner_constraints(manufacturers_model_name, orders):
+    # ! Modified
```
po09i: Remove comment

shimmingtoolbox/coils/coil.py
```diff
@@ -181,80 +203,82 @@ def get_scanner_constraints(manufacturers_model_name, order=2):
...
+            constraints["coef_channel_minmax"]["1"] = [[None, None] for _ in range(3)]
+        if 2 in orders:
+            constraints["coef_channel_minmax"]["2"] = [[None, None] for _ in range(5)]
 
     return constraints
 
 
-def restrict_sph_constraints(bounds, order):
+def restrict_sph_constraints(bounds: dict, orders):
+    # ! Modify description if everything works
```
po09i, line 258: Remove comment

shimmingtoolbox/coils/spher_harm_basis.py
```diff
@@ -52,14 +52,20 @@ def siemens_basis(x, y, z, orders=(1, 2), shim_cs='LAI'):
...
 
     # Select order
-    range_per_order = {1: list(range(3)), 2: list(range(3, 8))}
+    range_per_order = {}
+    index = 0
+    for order in orders:
+        range_per_order[order] = list(range(index, index + ORDER_INDEXES[order]))
+        index += ORDER_INDEXES[order]
+
+    # range_per_order = {1: list(range(3)), 2: list(range(3, 8))}
```
po09i, line 68: Remove comment

shimmingtoolbox/optimizer/basic_optimizer.py
```diff
@@ -167,6 +167,7 @@ def merge_coils(self, unshimmed, affine):
         return coil_profiles, bounds
 
     def merge_bounds(self):
+        #! Modified
```
po09i: Remove comment

shimmingtoolbox/coils/spher_harm_basis.py
```diff
@@ -214,15 +243,25 @@ def _reorder_to_ge(spher_harm):
...
+    if orders == (1,):
+        if spher_harm.shape[-1] != 3:
+            raise RuntimeError(f"Input arrays should have 4th dimension's shape equal to 3 {spher_harm.shape}")
+        reordered = spher_harm[..., [2, 0, 1]]
 
     return reordered
 
 
-def scaled_spher_harm(x, y, z, orders=(1, 2), shim_cs='ras'):
+def scaled_spher_harm(x, y, z, all_orders, shim_cs='ras'):
```
po09i, line 264: If the input spherical harmonic order is 2, the scaling factor that gets applied is wrong. (scaling_factors always has 8 coefficients)

po09i (changes requested):

po09i (approved):
LGTM!

PR #485, opened 2023-09-26
Add ability to simulate x, y and z gradients

po09i:
## Description
This PR adds the ability to simulate x, y and z gradients


shimmingtoolbox/simulate/numerical_model.py
```diff
@@ -107,30 +106,69 @@ def generate_deltaB0(self, field_type, params):
...
-                field (Hz).
+            field_type (str): Type of field to be generated. Available implementations are: ``'x'``, ``'y'``, ``'z'``.
+            params (list): List of parameters defining the field for the selected
+                           field type. If ``field_type = 'x' or 'y' or 'z'``, then ``params`` are
+                           ``[m b]`` where m (Hz/pixel) is the slope and b is the floor
+                           field (Hz).
         """
 
-        if field_type == "linear":
+        if field_type == "x":
```
4rnaudB, line 131: Would do:
if len(dims) == 2:
        [X, Y] = np.meshgrid(
         np.linspace(-dims[0], dims[0], dims[0]),
         np.linspace(-dims[1], dims[1], dims[1]))
        
if len(dims) == 3:
         [X, Y, Z] = np.meshgrid(
                        np.linspace(-dims[0], dims[0], dims[0]),
                        np.linspace(-dims[1], dims[1], dims[1]),
                        np.linspace(-dims[2], dims[2], dims[2]))

and then check for which gradient to set self.deltaB0 = m * Z + b

po09i: Good idea

shimmingtoolbox/simulate/numerical_model.py
```diff
@@ -112,60 +112,30 @@ def generate_deltaB0(self, field_type, params):
...
-            if len(dims) == 2:
-                Z = 0
-
-            if len(dims) == 3:
-                [_, _, Z] = np.meshgrid(
-                    np.linspace(-dims[0], dims[0], dims[0]),
-                    np.linspace(-dims[1], dims[1], dims[1]),
-                    np.linspace(-dims[2], dims[2], dims[2]))
+        elif field_type == "z":
             self.deltaB0 = m * Z + b
```
4rnaudB: Do we want to return b if len(dim) == 2 and field_type =='z'?

po09i: Yes, this is currently the case because Z is getting set to 0 if it is 2D

4rnaudB (approved):

PR #498, opened 2023-11-06
Implement Philips+GE related spherical harmonics conventions

### LINKED ISSUE #486: Validate GE shim coordinate system
We have implemented GE spherical harmonics in #476, but we still need to validate it on the scanner.

po09i:
## Description
This PR adds Philips as a manufacturer for B0 shimming with spherical harmonics. 

More specifically
- Uses the coordinate system of the scanner when calculating the spherical harmonics rather than changing the coordinates after calculating the harmonics
- Windows installs in release mode by default (`pip install .` rather than `pip install -e .`)
- Adds Philips and modifies GE output for the spherical harmonics scanner basis fields
- Fixes a bug where a plot would make exactly 0 values transparent rather than showing a 0 (fig_shimmed_vs_unshimmed)
- Add robust tests for the basis functions for each manufacturer

- Investigate in subsequent PR: #505 

Fixes #486


4rnaudB (approved):
Cannot test more than you but the code looks good. Seems like you simply divided the specific coefs that were causing problems for GE and Philips.

PR #410, opened 2022-09-15
Field mapping pipeline improvement

po09i:
## Description
This PR improves the field mapping pipeline. More precisely, it:

- Resolves and issue where the scaling of phase data was erroneously subtracted instead of "wrapped". (This issue was only present in data ranging from [0, 4095])

PR #499, opened 2023-11-16
Change phase range in field mapping calculations

4rnaudB:
## Checklist

- [x] I've updated the documentation and/or added the correct docstrings

## Description
Siemens outputs phase value in different formats ([0, 4095], [-4096, 4095]). We previously thought that [0, 4095] would map to [0, 2pi]. With experiments, we realized that it is mapped to [-pi, pi] for at least one sequence (gre_fm). This created a shift of f0 when calculating the field maps. This PR fixes this problem by subtracting pi when the phase values are between [0, 4095].

### DIFF
--- a/shimmingtoolbox/load_nifti.py
+++ b/shimmingtoolbox/load_nifti.py
@@ -182,11 +182,12 @@ def read_nii(fname_nifti, auto_scale=True):
 
             if np.amin(image) < 0 and (0.95 * 2 * PHASE_SCALING_SIEMENS < extent < 2 * PHASE_SCALING_SIEMENS * 1.05):
                 # Siemens' scaling: [-4096, 4095] --> [-pi, pi)
-                image = image * math.pi / PHASE_SCALING_SIEMENS
+                image = (image * math.pi / PHASE_SCALING_SIEMENS)
+                image = np.angle(np.exp(1j * image))
             elif np.amin(image) >= 0 and (0.95 * PHASE_SCALING_SIEMENS < extent < PHASE_SCALING_SIEMENS * 1.05):
                 # Siemens' scaling [0, 4095] --> [0, 2pi)
                 # We want: [-pi, pi]
-                image = image * 2 * math.pi / PHASE_SCALING_SIEMENS
+                image = (image * 2 * math.pi / PHASE_SCALING_SIEMENS) - math.pi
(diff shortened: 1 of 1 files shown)


po09i:
Related to: https://github.com/shimming-toolbox/shimming-toolbox/pull/410

4rnaudB:
Update with 5 different sequences tested:
Notes:
- Sequence 1: gre_fl3d2 (Eva eddy currents sequence)
     - Pairing found
     - 2 echoes – 2 phases
- Sequence 2: gre_reverse_fl3d2 
     - No pairing found 
     - 2 echoes – 2 phases 
- Sequence 3: gre_TTL_fl2d2 (Invivo sequence for abstract)
     - No pairing found
     - 2 echos – 2 phases
- Sequence 4: gre_field_mapping_fm2d2
     - Pairing found
     - 2 echos – phase diff
     - Pi offset!
- 3’: 3 echos
    - No pairing found
    - 3 echoes - 3 phases

4rnaudB:
Phase diff seems to be the only factor creating this pi offset.

po09i:
I looked into the data from acdc_228 and noticed that the phase difference information from sequence 4 has phase values ranging from ~-4k to ~+4k. That would not make sense with the fix we are proposing right now. I'm a bit confused.

4rnaudB:
I think we can just change the phase range when using a phase difference no?

po09i:
I think we need to make sure that the phase has the appropriate range. The test we have done looks at it after computing the field map. Since we compute the difference for dual-echo images and do a linear fit with multi-echo images, this is not going to find the pi offset.

4rnaudB:
Update 2: 
The difference in phase range ([0, 4095] vs [-4096, 4095]) is caused by the initial input. `Dicoms` files give the [0, 4095] phase range while the `no extension` files give the [-4096, 4095] range (.dcm are output when connected to the VLAN while the extension-less files are downloaded from sequoia). This seems to only cause the phase offset in the field map when creating a phase difference file instead of multiple phase files.

4rnaudB:
@po09i After testing dicoms vs no extension, it is confirmed that dicoms have a pi offset. What would be 0 with ima is -pi for dicoms (in the phase image). The modifications done in this PR fix this issue.

po09i (approved):

PR #504, opened 2023-12-20
Improve reliability of phase diff renaming

po09i:
## Description
This PR specifies the sequence that we should rename to a phase difference. The previous implementation caused issues when the sequenceName had extra characters (i.e.: fm2d2r).


4rnaudB (approved):
My sequence that uses a phase diff is *fm2d2. I don't think the star is important, but just making sure with you.

po09i:
> My sequence that uses a phase diff is *fm2d2. I don't think the star is important, but just making sure with you.

I'm not entirely sure why there is a "*" at the beginning but I encountered a bug with a sequence that had a "r" at the end

PR #509, opened 2024-01-22
Automatic mask creation for MRS shimming based on voxel position 

behrouzvia:
## Checklist

## Description
  A new functionality is added to masking for creating a mask that corresponds to the MRS voxel. The mask is created by reading the MRS voxel position from Siemens' MRS raw data. This mask can be used subsequently to shim the voxel position. The functionality is available through st_mask cli.


shimmingtoolbox/cli/mask.py
```diff
@@ -8,6 +8,7 @@
 
 from shimmingtoolbox.masking.shapes import shape_square, shape_cube, shape_sphere
 import shimmingtoolbox.masking.threshold
+import shimmingtoolbox.masking.mask_mrs
```
po09i:
```suggestion
import shimmingtoolbox.masking.mask_mrs import mask_mrs
```

shimmingtoolbox/cli/mask.py
```diff
@@ -303,6 +304,41 @@ def sct(fname_input, fname_output, contrast, centerline, file_centerline, brain,
...
+def mask_mrs(fname_input, output, raw_data, x_position, y_position, z_position, voxel_size, verbose):
+
+    # Set all loggers
+    set_all_loggers(verbose)
+
+    # Prepare the output
+    create_output_dir(output, is_file=True)
+
+    nii = nib.load(fname_input)
+    mask_mrs = shimmingtoolbox.masking.mask_mrs.mask_mrs(fname_input, raw_data, x_position, y_position, z_position, voxel_size) # creation of the MRS mask
```
po09i:
```suggestion
    mask_mrs = mask_mrs(fname_input, raw_data, x_position, y_position, z_position, voxel_size) # creation of the MRS mask
```

shimmingtoolbox/cli/mask.py
```diff
@@ -303,6 +304,41 @@ def sct(fname_input, fname_output, contrast, centerline, file_centerline, brain,
...
+@click.option('-i', '--input', 'fname_input',  type=click.Path(), required=True,
+               help="Input path of the fieldmap to be shimmed (both nifti and json file should be in this path).")
+@click.option('-raw', '--raw_data', type=str,
+              help="Input path of the of the twix raw-data (supported extention .dat) [optional]")
+@click.option('-o', '--output', type=click.Path(), default=os.path.join(os.curdir, 'mask_mrs.nii.gz'), show_default=True,
+              help="Name of the output mask. Supported extensions are .nii or .nii.gz. (default: "
+                   "(os.curdir, 'mask_mrs.nii.gz'))")
+@click.option('-X', 'x_position', type=click.FLOAT, help="scanner's X position in mm [optional]")
+@click.option('-Y', 'y_position', type=click.FLOAT, help="scanner's Y position in mm [optional]")
+@click.option('-Z', 'z_position', type=click.FLOAT, help="scanner's Z position in mm [optional]")
```
po09i: Maybe harmonize inputs to something similar to `st_mask box` (`--center` option).

shimmingtoolbox/cli/mask.py
```diff
@@ -303,6 +304,41 @@ def sct(fname_input, fname_output, contrast, centerline, file_centerline, brain,
...
+@click.option('-V', 'voxel_size', type=click.FLOAT, help="MRS voxel size in mm [optional]")
```
po09i: You could also potentially add a `--size` option as well to have the user specify the ROI in all dimensions.

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,101 @@
...
+import logging
+import json
+import subprocess
+
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+def mask_mrs(fname_input, raw_data, X, Y, Z, V):
```
po09i: Use lowercase variables

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,101 @@
...
+    """
+    Create a mask to shim single voxel MRS
+
+    Args:
+        fname_input: Input path of the fieldmap to be shimmed (both nii.gz and json file should be in this path)
```
po09i: Add variable types in the docstring

```suggestion
        fname_input (str): Input path of the fieldmap to be shimmed (both nii.gz and json file should be in this path)
```

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,101 @@
...
+    Returns:
+        numpy.ndarray: Cubic mask with same dimensions as MRS voxel.
+    """
+
+    if fname_input is None:
+        raise FileNotFoundError(f"The file '{fname_input}' was not found. See: st_mask mask_mrs -h")
+
+    if X is not None and Y is not None and Z is not None and V is not None:
+        # If X,Y,Z,V arguments were provided, execute this block
+        MRS_voxel_thick = V
```
po09i: Use lowercase

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,101 @@
...
+        Y: scanner's Y position in mm [optional]
+        Z: scanner's Z position in mm [optional]
+        V: MRS voxel size in mm [optional]
+
+    Returns:
+        numpy.ndarray: Cubic mask with same dimensions as MRS voxel.
+    """
+
+    if fname_input is None:
+        raise FileNotFoundError(f"The file '{fname_input}' was not found. See: st_mask mask_mrs -h")
```
po09i: This is not a `FileNotFoundError` if it `fname_input` is `None`, it just means that it was not provided.

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,101 @@
...
+    else:
+        # If X,Y,Z,V arguments were NOT provided, execute this block
+        if raw_data is None:
+            raise FileNotFoundError(f"The file '{raw_data}' was not found. If X,Y,Z and V are not directly given the raw-data must be given to read these info, See: st_mask mask_mrs -h")
+        else:
+            logger.info("Reading the twix raw-data")
+            subprocess.run(f"spec2nii twix {raw_data} -e image", shell=True )
+            name_nii = os.path.splitext(raw_data)[0]
+            header_twix= nib.load(name_nii + '.nii.gz').header
+            affine= nib.load(name_nii + '.nii.gz').affine
```
po09i:
```suggestion
            nii = nib.load(name_nii + '.nii.gz')
            header_twix= nii.header
            affine= nii.affine
```

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,101 @@
...
+            position_sag= header_twix['qoffset_x']
+            position_cor = header_twix['qoffset_y']
+            position_tra = header_twix['qoffset_z']
+            logger.info(f"twix header: {header_twix}")
+            logger.info(f"affine: {affine}")
+            scanner_coordinate= np.array([position_sag, position_cor, position_tra, 1])
+            logger.info(f"Scanner position: {scanner_coordinate}")
```
po09i: Use logger.info for statements that provide value for the user, logger.debug for added verbose, potentially used as debugging

shimmingtoolbox/masking/mask_mrs.py
po09i: Also, this variable is not used

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,101 @@
...
+
+    else:
+        # If X,Y,Z,V arguments were NOT provided, execute this block
+        if raw_data is None:
+            raise FileNotFoundError(f"The file '{raw_data}' was not found. If X,Y,Z and V are not directly given the raw-data must be given to read these info, See: st_mask mask_mrs -h")
+        else:
+            logger.info("Reading the twix raw-data")
+            subprocess.run(f"spec2nii twix {raw_data} -e image", shell=True )
+            name_nii = os.path.splitext(raw_data)[0]
+            header_twix= nib.load(name_nii + '.nii.gz').header
```
po09i: Fix all equal signs
```suggestion
            header_twix = nib.load(name_nii + '.nii.gz').header
```

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,101 @@
...
+            position_sag= header_twix['qoffset_x']
+            position_cor = header_twix['qoffset_y']
+            position_tra = header_twix['qoffset_z']
+            logger.info(f"twix header: {header_twix}")
+            logger.info(f"affine: {affine}")
+            scanner_coordinate= np.array([position_sag, position_cor, position_tra, 1])
+            logger.info(f"Scanner position: {scanner_coordinate}")
+
+
+    MRS_voxel_thick = header_twix['pixdim'][1]
```
po09i: header_twix will not be assigned if the user specifies X, Y and Z. This would result in a crash.

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,101 @@
...
+        V: MRS voxel size in mm [optional]
+
+    Returns:
+        numpy.ndarray: Cubic mask with same dimensions as MRS voxel.
+    """
+
+    if fname_input is None:
+        raise FileNotFoundError(f"The file '{fname_input}' was not found. See: st_mask mask_mrs -h")
+
+    if X is not None and Y is not None and Z is not None and V is not None:
```
po09i: What if only X is input by the user

shimmingtoolbox/cli/mask.py
```diff
@@ -303,6 +304,41 @@ def sct(fname_input, fname_output, contrast, centerline, file_centerline, brain,
...
 
+@mask_cli.command(context_settings=CONTEXT_SETTINGS,
+                  help="Create a mask to shim single voxel MRS."
+                       "Scanner's XYZ coordiante and MRS Voxel size can be directly given or these info can be read "
+                       "from the twix raw-data."
+                       "The mask is stored by default under the name 'mask_mrs.nii.gz' in the output "
+                       "folder. Return an output nifti file to be used as a mask for MRS shimming.")
+@click.option('-i', '--input', 'fname_input',  type=click.Path(), required=True,
+               help="Input path of the fieldmap to be shimmed (both nifti and json file should be in this path).")
+@click.option('-raw', '--raw_data', type=str,
```
po09i: Use type: `click.Path()` if it is a path

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,101 @@
...
+        scanner_coordinate = np.array([X, Y, Z, 1])
+        logger.info(f"Scanner position: {scanner_coordinate}")
+
+    else:
+        # If X,Y,Z,V arguments were NOT provided, execute this block
+        if raw_data is None:
+            raise FileNotFoundError(f"The file '{raw_data}' was not found. If X,Y,Z and V are not directly given the raw-data must be given to read these info, See: st_mask mask_mrs -h")
+        else:
+            logger.info("Reading the twix raw-data")
+            subprocess.run(f"spec2nii twix {raw_data} -e image", shell=True )
```
po09i: Use Shimming Toolbox's `shimmingtoolbox/utils/run_subprocess()` function.

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,101 @@
...
+            name_nii = os.path.splitext(raw_data)[0]
```
po09i: I suggest using `shimming toolbox/utils:splitext()` for all extension splitting.

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,101 @@
...
+    data_dir = os.path.dirname(fname_input)
+
+    # Get the file name of the input without extension
+    json_name = os.path.splitext(os.path.basename(fname_input))[0][:-4]
+
+    # Combine the directory and file name without extension
+    json_dir = os.path.join(data_dir, json_name)
+
+    # Open the JSON file
+    json_file = json_dir+'.json'
```
po09i: I usually do something like this:
```suggestion
    _, ext = splitext(fname_input)
    json_file = fname_anat.rsplit('.' + ext, 1)[0] + '.json'
```

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,101 @@
...
+
+    with open(json_file, 'r') as info:
+        # Load the data from the json file
+        header_dcm = json.load(info)
```
po09i: This is the BIDS metadata, not the dicom header. I suggest renaming the variable name

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,101 @@
...
+
+    # Read the slice thickness of the reference GRE image.
+    slice_thickness = header_dcm['SliceThickness'] # Isotropic fieldmap (mm)
+
+    # The given XYZ is the voxel's center position.
+    # To create a mask the distance from center of MRS voxel to its edge should be calculated based on number of fieldmap voxels.
+    SD = int(np.ceil(MRS_voxel_thick/(2*slice_thickness)))
```
po09i: Is the MRS voxel thickness always a square? This is only using pixdim[1] which I believe is only in the x direction.
What do you think about using --size and letting the user choose? This could be the default option though for the case where it is not provided as it seems useful to have.

po09i (changes requested):

shimmingtoolbox/cli/mask.py
po09i: -V currently does nothing (it is assigned to a unused variable)

shimmingtoolbox/cli/mask.py
```diff
@@ -303,6 +304,41 @@ def sct(fname_input, fname_output, contrast, centerline, file_centerline, brain,
...
+    # Set all loggers
+    set_all_loggers(verbose)
+
+    # Prepare the output
+    create_output_dir(output, is_file=True)
+
+    nii = nib.load(fname_input)
+    mask_mrs = shimmingtoolbox.masking.mask_mrs.mask_mrs(fname_input, raw_data, x_position, y_position, z_position, voxel_size) # creation of the MRS mask
+    mask_mrs = mask_mrs.astype(np.int32)
+    nii_img = nib.Nifti1Image(mask_mrs, nii.affine)
```
po09i:
```suggestion
    nii_img = nib.Nifti1Image(mask_mrs, nii.affine, header=nii.header)
```

shimmingtoolbox/masking/mask_mrs.py
behrouzvia: For most of the cases, the voxel is square but you are right it's better to have a 'size' variable to include all the voxel's dimensions. However, in that case, I need to include the in-plane fmap's resolution and I could not find this info in the json file. Do you know how can I get the size of the FOV to calculate the resolution in x and y direction?

po09i: That is included in the NIfTI header (`nii.header['pixdim']`) I believe. You could also extract it from the affine. Nibabel also has a helper function to output that (`nii.header.get_zooms()`). I would probably go with my last suggestions (`get_zooms()`)

shimmingtoolbox/cli/mask.py
```diff
@@ -303,6 +304,39 @@ def sct(fname_input, fname_output, contrast, centerline, file_centerline, brain,
     return fname_output
 
 
+@mask_cli.command(context_settings=CONTEXT_SETTINGS,
+                  help="Create a mask to shim single voxel MRS."
+                       "voxel position and size can be directly given or these info can be read "
```
po09i: Make sure to leave a space between two sentences and capitalize the "V". This makes sure the help text is formatted properly.
Currently
Screenshot 2024-01-25 at 2 52 47 PM

shimmingtoolbox/cli/mask.py
```diff
@@ -303,6 +304,39 @@ def sct(fname_input, fname_output, contrast, centerline, file_centerline, brain,
...
+               help="Input path of the fieldmap to be shimmed.")
+@click.option('-r', '--raw_data', type=click.Path(),
+              help="Input path of the of the twix raw-data (supported extention .dat) [optional]")
+@click.option('-o', '--output', type=click.Path(), default=os.path.join(os.curdir, 'mask_mrs.nii.gz'), show_default=True,
+              help="Name of the output mask. Supported extensions are .nii or .nii.gz. (default: "
+                   "(os.curdir, 'mask_mrs.nii.gz'))")
+@click.option('-c', '--center', nargs=3, type=click.FLOAT, help="voxel's center position in mm of the x, y and z of the scanner's coordinate [optional]")
+@click.option('-s', '--size', nargs=3, type=click.FLOAT, help="voxel size in mm of the x, y and z of the scanner's coordinate")
+@click.option('--verbose', type=click.Choice(['info', 'debug']), default='info', help="Be more verbose")
+def mrs_mask(fname_input, output, raw_data, center, size, verbose):
```
po09i: The name of the function is the name of the CLI argument. Currently you need to write `st_mask mrs-mask` which is redundant. I suggest renaming to "mask" so that the user inputs `st_mask mrs`
```suggestion
def mrs(fname_input, output, raw_data, center, size, verbose):
```

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,85 @@
...
+logger = logging.getLogger(__name__)
+
+
+def mask_mrs(fname_input, raw_data, center, size):
+    """
+    Create a mask to shim single voxel MRS
+
+    Args:
+        fname_input (str): Input path of the fieldmap to be shimmed (supported extention .nii and .nii.gz) [required]
+        raw_data (str): Input path of the of the twix raw-data (supported extention .dat) [optional]
```
po09i:
```suggestion
        raw_data (str): Input path of the twix raw-data (supported extention .dat)
```

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,85 @@
...
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+def mask_mrs(fname_input, raw_data, center, size):
+    """
+    Create a mask to shim single voxel MRS
+
+    Args:
+        fname_input (str): Input path of the fieldmap to be shimmed (supported extention .nii and .nii.gz) [required]
```
po09i:
```suggestion
        fname_input (str): Input path of the fieldmap to be shimmed (supported extention .nii and .nii.gz)
```

shimmingtoolbox/masking/mask_mrs.py
po09i: Since there is no default option, this argument is required.

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,85 @@
...
+        raw_data (str): Input path of the of the twix raw-data (supported extention .dat) [optional]
+        center (float): voxel's center position in mm of the x, y and z of the scanner's coordinate [optional]
+        size (float): voxel size in mm of the x, y and z of the scanner's coordinate [optional]
```
po09i: Same comment
```suggestion
        center (float): voxel's center position in mm of the x, y and z of the scanner's coordinate
        size (float): voxel size in mm of the x, y and z of the scanner's coordinate
```

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,85 @@
...
+
+    Returns:
+        numpy.ndarray: Cubic mask with same dimensions as MRS voxel.
+    """
+
+    if fname_input is None:
+        raise TypeError(f"The fname_input is cannot be None. See: st_mask mrs-mask -h")
```
po09i: This is the API, you are referencing something that is outside its scope (the CLI) by saying `st_mask mrs-mask -h`.  You should do this in the CLI file if you want tot write this error message. The API should contain this type of error checking, but with a different comment that makes sense to its scope.

shimmingtoolbox/masking/mask_mrs.py
po09i: Are these rally floats? They seem to be lists in your code

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,85 @@
...
+
+    if raw_data is None:
+        logger.info(" The raw_data is not provided, creating the mask with the given voxel position and size info")
+        if center is None or size is None:
+             raise TypeError('As the raw_data is not provided, the voxel position and size must be given to create the mask (See: st_mask mrs-mask -h)')
```
po09i: Same comment

po09i: Also, it is over indented (5 spaces rather than 4)

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,85 @@
...
+        else:
+            mrs_voxel_size = size
+            logger.debug('center:',center )
```
po09i:
```suggestion
            logger.debug('center:', center)
```

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,85 @@
...
+            header_twix = nii.header
+            affine = nii.affine
+            position_sag = header_twix['qoffset_x']
+            position_cor = header_twix['qoffset_y']
+            position_tra = header_twix['qoffset_z']
+            logger.debug(f"twix header: {header_twix}")
+            logger.debug(f"affine: {affine}")
+            scanner_coordinate = np.array([position_sag, position_cor, position_tra, 1])
+            logger.debug(f"Scanner position is: {scanner_coordinate}")
+            mrs_voxel_size = header_twix['pixdim'][1:4]
```
po09i: This whole block is over indented

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,85 @@
...
+
```
po09i:
```suggestion
```

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,85 @@
...
+    fmap_array = fmap_nii.get_fdata()
+    fmap_affine = fmap_nii.affine
+    fmap_header = fmap_nii.header
+    logger.debug('reference_affine:', fmap_affine)
+    logger.debug('reference_affine shape:', np.shape(fmap_affine))
+    voxel_position = npl.inv(fmap_affine).dot(scanner_coordinate)
+    voxel_position = np.round(voxel_position)
+    logger.debug('voxel_position:', voxel_position)
+
+      # The given coordinate (I, J, K) is the voxel's center position.
```
po09i: Over indented

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,85 @@
...
+    I, J, K = map(int, voxel_position[:3])
```
po09i: Capital letters

po09i (changes requested):
We have tried keeping the number of characters for each line below 120 characters. Some of the lines are too long.

Nice job on the last changes!

behrouzvia:
> We have tried keeping the number of characters for each line below 120 characters. Some of the lines are too long.
> 
> Nice job on the last changes!

Thanks for the review and comments! I applied the suggestions.

shimmingtoolbox/cli/mask.py
```diff
@@ -303,6 +304,41 @@ def sct(fname_input, fname_output, contrast, centerline, file_centerline, brain,
...
 
 
+@mask_cli.command(context_settings=CONTEXT_SETTINGS,
+                  help="Create a mask to shim single voxel MRS. "
+                       "Voxel position and size can be directly given or these info can be read "
+                       "from the twix raw-data. "
+                       "The mask is stored by default under the name 'mask_mrs.nii.gz' in the output "
+                       "folder. Return an output nifti file to be used as a mask for MRS shimming.")
+@click.option('-i', '--input', 'fname_input', type=click.Path(), required=True,
+               help="Input path of the fieldmap to be shimmed.")
```
po09i: Indentation

shimmingtoolbox/cli/mask.py
```diff
@@ -303,6 +304,41 @@ def sct(fname_input, fname_output, contrast, centerline, file_centerline, brain,
...
+@click.option('-r', '--raw_data', type=click.Path(),
+              help="Input path of the of the twix raw-data (supported extention .dat)")
+@click.option('-o', '--output', type=click.Path(), default=os.path.join(os.curdir, 'mask_mrs.nii.gz'),
+              show_default=True, help="Name of the output mask. Supported extensions are .nii or .nii.gz. (default: "
+                   "(os.curdir, 'mask_mrs.nii.gz'))")
```
po09i: Indentation

shimmingtoolbox/cli/mask.py
```diff
@@ -303,6 +304,41 @@ def sct(fname_input, fname_output, contrast, centerline, file_centerline, brain,
...
+def mrs(fname_input, output, raw_data, center, size, verbose):
+
+    # Set all loggers
+    set_all_loggers(verbose)
+
+    # Prepare the output
+    create_output_dir(output, is_file=True)
+
+    nii = nib.load(fname_input)
+    output_mask = mask_mrs(fname_input, raw_data, center, size) # creation of the MRS mask
```
po09i: Inline comments should have 2 spaces between the end of the line of code and the comment

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,83 @@
...
+logger = logging.getLogger(__name__)
+
+
+def mask_mrs(fname_input, raw_data, center, size):
+    """
+    Create a mask to shim single voxel MRS
+
+    Args:
+        fname_input (str): Input path of the fieldmap to be shimmed (supported extention .nii and .nii.gz)
+        raw_data (str): Input path of the of the twix raw-data (supported extention .dat)
```
po09i:
```suggestion
        raw_data (str): Input path of the twix raw-data (supported extension .dat)
```

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,83 @@
...
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+def mask_mrs(fname_input, raw_data, center, size):
+    """
+    Create a mask to shim single voxel MRS
+
+    Args:
+        fname_input (str): Input path of the fieldmap to be shimmed (supported extention .nii and .nii.gz)
```
po09i:
```suggestion
        fname_input (str): Input path of the fieldmap to be shimmed (supported extension .nii and .nii.gz)
```

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,83 @@
...
+        raw_data (str): Input path of the of the twix raw-data (supported extention .dat)
+        center (list): voxel's center position in mm of the x, y and z of the scanner's coordinate
```
po09i:
```suggestion
        center (list): Voxel's center position in mm of the x, y and z scanner coordinates
```

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,83 @@
...
+        size (list): voxel size in mm of the x, y and z of the scanner's coordinate
```
po09i:
```suggestion
        size (list): Voxel size in mm of the x, y and z scanner coordinates
```

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,83 @@
...
+    Returns:
+        numpy.ndarray: Cubic mask with same dimensions as MRS voxel.
```
po09i:
```suggestion
        numpy.ndarray: Cubic mask with the same dimensions as the MRS voxel.
```

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,83 @@
...
+    """
+
+    if fname_input is None:
+        raise TypeError(f"The fname_input cannot be None.")
```
po09i:
```suggestion
        raise TypeError(f"The input field map needs to be specified")
```

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,83 @@
...
+
+    if raw_data is None:
+        logger.info(" The raw_data is not provided, creating the mask with the given voxel position and size info")
```
po09i:
```suggestion
        logger.info("MRS raw data not provided, creating the mask with the given voxel position and size info")
```

shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,83 @@
...
+        if center is None or size is None:
+            raise TypeError('The raw_data is not provided; the voxel position and size are required to proceed')
+        else:
+            mrs_voxel_size = size
+            logger.debug('center:', center)
+            scanner_coordinate = np.array(center + (1,))
```
po09i: Add space between comma and following character
```suggestion
            scanner_coordinate = np.array(center + (1, ))
```

po09i (changes requested):
- Add tests for your new function
- I'm not sure if you are using an editor, but most of the comments I am making are simply underlined in my editor to respect PEP conventions. If you are not using an editor, I suggests you do, if not, I suggest familiarizing yourself wit PEP conventions.
- You can [add my suggestions](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request) directly in GitHub rather than copying what I propose.

po09i:
Once a test has been implemented to automatically figure out the center location of MRS data, we can merge this PR

shimmingtoolbox/cli/mask.py
```diff
@@ -303,6 +304,41 @@ def sct(fname_input, fname_output, contrast, centerline, file_centerline, brain,
...
+@mask_cli.command(context_settings=CONTEXT_SETTINGS,
+                  help="Create a mask to shim single voxel MRS. "
+                       "Voxel position and size can be directly given or these info can be read "
+                       "from the siemens raw-data. "
+                       "The mask is stored by default under the name 'mask_mrs.nii.gz' in the output "
+                       "folder. Return an output nifti file to be used as a mask for MRS shimming.")
+@click.option('-i', '--input', 'fname_input', type=click.Path(), required=True,
+              help="Input path of the fieldmap to be shimmed.")
+@click.option('-r', '--raw_data', type=click.Path(),
+              help="Input path of the of the siemens raw-data (supported extention .rda)")
```
po09i: Why remove the twix extension? We should find a way to test for it but we should have support for both

behrouzvia: The standard Siemens raw data for MRS is rda file. It's more accessible to every user without the need to enable the advanced user to save it. Also as discussed before the twix files are larger (with file sizes reaching 300 MB compared to the much smaller 20 KB of RDA files) and cannot be stored on the Github testing repo.

po09i: Indeed, but what if someone is not using Siemens.. I think we should keep the possibility of using twix files and if we can't test for it in the CI because it is too big, we can test locally.

po09i:
Tests are failing, I'll review when all tests pass :)

shimmingtoolbox/cli/mask.py
po09i: @behrouzvia We should add back support for twix data

behrouzvia: > @behrouzvia We should add back support for Twix data

Well, the twix format is also specific to Siemens, and each vendor has their own specific format for MRS. For example, Philips' format is sdat and GE uses a P-file format. So adding the possibility of using the twix format would not bring a huge gain.

po09i:
```suggestion
              help="Input path of the siemens raw-data (supported extention .rda)")
```

shimmingtoolbox/cli/mask.py
```diff
@@ -303,6 +304,41 @@ def sct(fname_input, fname_output, contrast, centerline, file_centerline, brain,
...
+@click.option('-o', '--output', type=click.Path(), default=os.path.join(os.curdir, 'mask_mrs.nii.gz'),
+              show_default=True, help="Name of the output mask. Supported extensions are .nii or .nii.gz. (default: "
+              "(os.curdir, 'mask_mrs.nii.gz'))")
```
po09i: No need to tell what the default is in the description. You already see it with the option "show_default"

```suggestion
              show_default=True, help="Name of the output mask. Supported extensions are .nii or .nii.gz.")
```

test/cli/test_cli_mask.py
```diff
@@ -180,3 +181,39 @@ def test_cli_mask_sct_4d():
...
+        nii = nib.load(nii_path)
+        header_raw_data = nii.header
+        position_sag = header_raw_data['qoffset_x']
+        position_cor = header_raw_data['qoffset_y']
+        position_tra = header_raw_data['qoffset_z']
+
+        mrs_center = np.array([position_sag, position_cor, position_tra])
+        mrs_voxel_size = header_raw_data['pixdim'][1:4]
+        mrs_voxel_size_str = ' '.join(map(str, mrs_voxel_size))
+        mrs_center_str = ' '.join(map(str, mrs_center))
```
po09i: Do not reimplement mask_mrs in the test. Just use what `mrs_center_str` and what `mrs_voxel_size_str` contain. The idea is giving a know input and verify that the output is the same across versions of the code. If you include all or some of the implementation of the API, we might not catch these differences.

test/cli/test_cli_mask.py
po09i, line 1: Add a test where you give the .rda file as an input

shimming-toolbox/shimmingtoolbox/masking/mask_mrs.py
```diff
@@ -0,0 +1,89 @@
...
+
+    # The given coordinate (i, j, k) is the voxel's center position.
+    i, j, k = map(int, voxel_position[:3])
+    fmap_resolution = fmap_header['pixdim'][1:4]
+
+    # The distance from the center of the MRS voxel to its edges is calculated based on number of fieldmap voxels.
+    sd = np.ceil(mrs_voxel_size / (2 * fmap_resolution)).astype(int)
+
+    # create a zero mask with the same size as the input fieldmap to be shimmed.
+    mask = np.zeros(fmap_array.shape)
```
4rnaudB, line 85: no header for the mask?

po09i: I'm not sure what you mean? What header?

4rnaudB: Right now it seems the mask is only a numpy array and it allows the mask creation without a nifti as input. Wouldn't it be a problem if there is no affine transform?

4rnaudB (approved):

shimming-toolbox/shimmingtoolbox/masking/mask_mrs.py
po09i: The conversion happens in the CLI

po09i (approved):

PR #482, opened 2023-09-06
Improve realtime shimming

po09i:
## Description
This PR improves the following

- More precise timepoints (1 per slice when possible rather than only one per volume)
- Mean pressure is taken from the average pressure in the acquisition window rather than the average pressure at the specific volume timepoints)
- New plot to visualize pressure vs field
- Modified the eval function of the Realtime sequencer to calculate the field on the original fmap (if it was changed due to having insufficient slices)
- New CLI option to change the relative time offset between the time reported in the acquisitions and the PMU

Useful ref:
- https://github.com/andrewrosss/fmri-physio-log
- https://www.cfn.upenn.edu/aguirre/wiki/doku.php?id=public:pulse-oximetry_during_fmri_scanning
- https://wiki.humanconnectome.org/display/PublicData/Understanding+Timing+Information+in+HCP+Physiological+Monitoring+Files
- https://en.wikibooks.org/wiki/SPM/Slice_Timing

PR #515, opened 2024-02-08
Adding Gradient-based optimizer (BFGS) and pseudo huber loss function 

behrouzvia:
## Checklist

## Description

This PR adds a gradient-based optimizer called "L-BFGS-B" which belongs to the Quasi-Newton family and is available from the Scipy library.  Also, I added a pseudo huber loss function to both the least-square optimizer and the gradient optimizer.


shimmingtoolbox/cli/b0shim.py
```diff
@@ -70,18 +70,18 @@ def b0shim_cli():
...
               help="Regularization factor for the current when optimizing. A higher coefficient will penalize higher "
                    "current values while 0 provides no regularization. Not relevant for 'pseudo-inverse' "
                    "optimizer_method.")
-@click.option('--optimizer-criteria', 'opt_criteria', type=click.Choice(['mse', 'mae']), required=False,
+@click.option('--optimizer-criteria', 'opt_criteria', type=click.Choice(['mse', 'mae', 'ps_huber']), required=False,
               default='mse', show_default=True,
-              help="Criteria of optimization for the optimizer 'least_squares'."
-                   " mse: Mean Squared Error, mae: Mean Absolute Error")
+              help="Criteria of optimization for the optimizer 'least_squares' and 'gradient'."
+                   " mse: Mean Squared Error, mae: Mean Absolute Error, ps_huber: pseudo huber cost function ")
```
po09i:
```suggestion
                   " mse: Mean Squared Error, mae: Mean Absolute Error, ps-huber: pseudo huber cost function ")
```

shimmingtoolbox/optimizer/gradient_optimizer.py
po09i, line 1: This looks like you copied the lsq_optimizer file. Instead of doing so, you could inherit from lsq_optimizer and redefine what needs to be different.

shimmingtoolbox/cli/b0shim.py
```diff
@@ -70,18 +70,18 @@ def b0shim_cli():
                    "'--slice-factor' value is '3', then with the 'sequential' mode, shimming will be performed "
                    "independently on the following groups: {0,1,2}, {3,4,5}, etc. With the mode 'interleaved', "
                    "it will be: {0,2,4}, {1,3,5}, etc.")
-@click.option('--optimizer-method', 'method', type=click.Choice(['least_squares', 'pseudo_inverse', 'quad_prog']),
-              required=False, default='quad_prog', show_default=True,
-              help="Method used by the optimizer. LS, and QP will respect the constraints,"
-                   "PS will not respect the constraints")
+@click.option('--optimizer-method', 'method', required=False, default='quad_prog', show_default=True,
+              type=click.Choice(['least_squares', 'pseudo_inverse', 'quad_prog', 'gradient']),
+              help="Method used by the optimizer. LS, QP and gradient will respect the constraints,"
```
po09i: I'm not sure if gradient is is the right name to call this.. L-BFGS is not easy to remember though...

behrouzvia: Yes, that's why I thought a more general name was better for that.

shimmingtoolbox/optimizer/gradient_optimizer.py
behrouzvia: Thanks, I updated the code and learned something new.

shimmingtoolbox/optimizer/gradient_optimizer.py
```diff
@@ -0,0 +1,48 @@
...
+            coils (ListCoil): List of Coil objects containing the coil profiles and related constraints
+            unshimmed (np.ndarray): 3d array of unshimmed volume
+            affine (np.ndarray): 4x4 array containing the affine transformation for the unshimmed array
+            opt_criteria (str): Criteria for the optimizer 'least_squares'. Supported: 'mse': mean squared error,
+            'mae': mean absolute error, 'std': standard deviation, 'ps_huber': pseudo huber cost function.
+            reg_factor (float): Regularization factor for the current when optimizing. A higher coefficient will
+                                penalize higher current values while a lower factor will lower the effect of the
+                                regularization. A negative value will favour high currents (not preferred).
+        """
+        super().__init__(coils, unshimmed, affine, opt_criteria, initial_guess_method, reg_factor)
```
po09i: Is __init__ the same as the lsq class? if so, I don't believe you need to redefine __init__.

po09i:
I just read that the L-BFGS-B method is a mound constrained optimization (it supports "bounds" but not "constraints". It Shimming Toolbox, that would mean that it could respect individual channel bounds, but not the constraint on the sum of the currents. For info, pseudo-inverse does not support bounds or constraints, SLSQP (least-squares) and quadprog do.

behrouzvia:
> I just read that the L-BFGS-B method is a mound constrained optimization (it supports "bounds" but not "constraints". It Shimming Toolbox, that would mean that it could respect individual channel bounds, but not the constraint on the sum of the currents. For info, pseudo-inverse does not support bounds or constraints, SLSQP (least-squares) and quadprog do.

Yes, I also read in the docs that it's a bound-constraint optimizer but as you can see here on this website (https://docs.scipy.org/doc/scipy/reference/optimize.minimize-lbfgsb.html#optimize-minimize-lbfgsb) the command syntax has both constraints the bound keyword argument. Because of that, I let the constraint remain in the code. But, actually now I tested that and the result is the same regardless of inputting constraint or not. So, I'll remove that.

po09i:
Can you update the help text in the CLI to reflect that.

shimmingtoolbox/cli/b0shim.py
```diff
@@ -70,18 +70,19 @@ def b0shim_cli():
...
                    "it will be: {0,2,4}, {1,3,5}, etc.")
-@click.option('--optimizer-method', 'method', type=click.Choice(['least_squares', 'pseudo_inverse', 'quad_prog']),
-              required=False, default='quad_prog', show_default=True,
-              help="Method used by the optimizer. LS, and QP will respect the constraints,"
-                   "PS will not respect the constraints")
+@click.option('--optimizer-method', 'method', required=False, default='quad_prog', show_default=True,
+              type=click.Choice(['least_squares', 'pseudo_inverse', 'quad_prog', 'gradient']),
+              help="Method used by the optimizer. LS, QP will respect the constraints, "
+              "gradient method only accepts bounds for each channel "
+              "PS will not respect the constraints")
```
po09i:
```suggestion
              help="Method used by the optimizer. LS and QP will respect the constraints, "
              "the gradient method only accepts constraints for each channel (not constraints on the total current), "
              "PS will not respect any constraints")
```

shimmingtoolbox/optimizer/gradient_optimizer.py
po09i: I believe you can remove it entirely rather than calling super()

shimmingtoolbox/optimizer/gradient_optimizer.py
```diff
@@ -0,0 +1,78 @@
...
+            method = 'L-BFGS-B'
+
+        return opt.minimize(self._criteria_func, currents_0,
+                            args=(a, b, c) if self.opt_criteria == 'mse' else (unshimmed_vec, coil_mat, factor),
+                            method=method,
+                            bounds=self.merged_bounds,
+                            jac=self._jacobian_func if self.opt_criteria == 'mse' else None,
+                            options={'maxiter': 1000})
+
+    def _get_currents(self, unshimmed_vec, coil_mat, currents_0):
```
po09i: Is this the same function as the one in LsqOptimizer? If so you do not need to redefine it

shimmingtoolbox/optimizer/lsq_optimizer.py
```diff
@@ -190,6 +215,14 @@ def _scipy_minimize(self, currents_0, unshimmed_vec, coil_mat, scipy_constraints
                                        jac=self._jacobian_func,
                                        options={'maxiter': 1000})
 
+        elif self.opt_criteria == 'ps_huber':
+            currents_sp = opt.minimize(self._criteria_func, currents_0,
+                                       args=(unshimmed_vec, coil_mat, factor),
+                                       method='SLSQP',
+                                       bounds=self.merged_bounds,
+                                       constraints=tuple(scipy_constraints),
+                                       options={'maxiter': 1000})
```
po09i: Why do you need this? Isn't this the same code that would be run is the "else" statement?

shimmingtoolbox/shim/sequencer.py
```diff
@@ -375,6 +377,10 @@ def select_optimizer(self):
                 optimizer = supported_optimizers[self.method](self.coils, self.nii_fieldmap.get_fdata(),
                                                               self.nii_fieldmap.affine, self.opt_criteria,
                                                               reg_factor=self.reg_factor)
+            elif self.method == 'gradient':
+                optimizer = supported_optimizers[self.method](self.coils, self.nii_fieldmap.get_fdata(),
+                                                              self.nii_fieldmap.affine, self.opt_criteria,
+                                                              reg_factor=self.reg_factor)
```
po09i: This is the same code called for the "least_squares" method. You can simply do:
if self.method in ['least_squares', 'gradient'] :

po09i:
Merge Master in your branch then I can review :)

behrouzvia:
> Merge Master in your branch then I can review :)

Done.

shimmingtoolbox/optimizer/gradient_optimizer.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+
+    def _scipy_minimize(self, currents_0, unshimmed_vec, coil_mat, factor):
+        if self.opt_criteria == 'mse':
+            a, b, c = self.get_quadratic_term(unshimmed_vec, coil_mat, factor)
+            method = 'L-BFGS-B'
+        elif self.opt_criteria == 'ps_huber':
+            method = 'L-BFGS-B'
+        else:
+            method = 'L-BFGS-B'
```
po09i: If all conditions contain `method = 'L-BFGS-B'`, you can move I out of the if statement

shimmingtoolbox/cli/b0shim.py
```diff
@@ -71,18 +71,19 @@ def b0shim_cli():
...
+@click.option('--optimizer-method', 'method', required=False, default='quad_prog', show_default=True,
+              type=click.Choice(['least_squares', 'pseudo_inverse', 'quad_prog', 'gradient']),
+              help="Method used by the optimizer. LS and QP will respect the constraints, "
+              "the gradient method only accepts constraints for each channel (not constraints on the total current), "
+              "PS will not respect any constraints")
 @click.option('--regularization-factor', 'reg_factor', type=click.FLOAT, required=False, default=0.0, show_default=True,
               help="Regularization factor for the current when optimizing. A higher coefficient will penalize higher "
                    "current values while 0 provides no regularization. Not relevant for 'pseudo-inverse' "
                    "optimizer_method.")
-@click.option('--optimizer-criteria', 'opt_criteria', type=click.Choice(['mse', 'mae', 'rmse']), required=False,
```
po09i: Don't remove rmse

po09i (commented):
Merge master, also make sure that rmse is still in the CLI options

shimming-toolbox/shimmingtoolbox/optimizer/lsq_optimizer.py
```diff
@@ -438,13 +468,56 @@ def __init__(self, coils, unshimmed, affine, opt_criteria, pmu: PmuResp, reg_fac
         super().__init__(coils, unshimmed, affine, opt_criteria, initial_guess_method='zeros', reg_factor=reg_factor)
         self.pressure_min = pmu.min
         self.pressure_max = pmu.max
+        # TODO: Implement mean pressure calculation. This should be taken from the PMU object between different times.
+        # I implemented something in the rt shim PR
```
po09i, line 473: Todo in #482

4rnaudB (changes requested):
bgfgs mse performs the exact same the lsq mse with the gradients. The update with mse_signal_recovery has not been fully done, adding w paramter doesn't change the optimzation. Also, using the opt-criteria ps_huber I get this error:
RuntimeError: Optimization failed due to: Inequality constraints incompatible

po09i:
I tested the different things you mentioned and could not replicate your findings, maybe the branch was not updated?

4rnaudB:
After updating the branch, the signal recovery problem was fixed, and the result in the spinal cord showed bfgs_rmse > lsq_rmse > lsq_mse > bfgs_mse. I did get an issue when using a custom coil in combination with ps_huber criteria:
```
/Users/arnaud/shimming-toolbox/python/lib/python3.10/site-packages/shimmingtoolbox/optimizer/lsq_optimizer.py:176: RuntimeWarning: invalid value encountered in scalar divide
  return np.sum(pseudo_huber(self._delta, np.abs(residuals))) / factor + np.abs(coef).dot(self.reg_vector)
Traceback (most recent call last):
  File "/Users/arnaud/shimming-toolbox/bin/st_b0shim", line 8, in <module>
    sys.exit(b0shim_cli())
  File "/Users/arnaud/shimming-toolbox/python/lib/python3.10/site-packages/click/core.py", line 1161, in __call__
    return self.main(*args, **kwargs)
  File "/Users/arnaud/shimming-toolbox/python/lib/python3.10/site-packages/click/core.py", line 1082, in main
    rv = self.invoke(ctx)
  File "/Users/arnaud/shimming-toolbox/python/lib/python3.10/site-packages/click/core.py", line 1697, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/arnaud/shimming-toolbox/python/lib/python3.10/site-packages/click/core.py", line 1443, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/arnaud/shimming-toolbox/python/lib/python3.10/site-packages/click/core.py", line 788, in invoke
    return __callback(*args, **kwargs)
  File "/Users/arnaud/shimming-toolbox/python/lib/python3.10/site-packages/shimmingtoolbox/utils.py", line 245, in timed
    result = func(*args, **kw)
  File "/Users/arnaud/shimming-toolbox/python/lib/python3.10/site-packages/shimmingtoolbox/cli/b0shim.py", line 332, in dynamic
    coefs = sequencer.shim()
  File "/Users/arnaud/shimming-toolbox/python/lib/python3.10/site-packages/shimmingtoolbox/shim/sequencer.py", line 380, in shim
    coefs = self.optimize(masks_fmap_dilated)
  File "/Users/arnaud/shimming-toolbox/python/lib/python3.10/site-packages/shimmingtoolbox/shim/sequencer.py", line 124, in optimize
    coefs.append(self.optimizer.optimize(masks_fmap[..., i]))
  File "/Users/arnaud/shimming-toolbox/python/lib/python3.10/site-packages/shimmingtoolbox/optimizer/lsq_optimizer.py", line 135, in optimize
    currents = self._get_currents(unshimmed_vec, coil_mat, currents_0)
  File "/Users/arnaud/shimming-toolbox/python/lib/python3.10/site-packages/shimmingtoolbox/optimizer/lsq_optimizer.py", line 402, in _get_currents
    raise RuntimeError(f"Optimization failed due to: {currents_sp.message}")
RuntimeError: Optimization failed due to: Inequality constraints incompatible
```

po09i:
> I did get an issue when using a custom coil in combination with ps_huber criteria:

There was an error in the ps_huber loss calculation. As seen in the [Scipy documentation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.huber.html), the huber/pseudo_huber calculation is done using a threshold where the loss follows the MSE when lower than the threshold and follows the MAE when higher than the threshold. The goal is to have a loss function that does not penalize extraneous values as much as the MSE does. There is however a third regime where if the threshold is negative, the loss becomes infinite. In the dataset you provided, the threshold was set to a negative value (which should never happen) and was always returning an infinite loss. I added an absolute value to fix the issue.

Moreover, the code was set up to use a threshold that was the highest value of the unshimmed field map. That seems like if there was an extraneous value, we would not want to set the threshold below that, therefore I selected the threshold to be at the 90th percentile of the absolute value of the unshimmed field map. 

[71586b4](https://github.com/shimming-toolbox/shimming-toolbox/pull/515/commits/71586b43ce639d532dc473e793ff9bac72f5ed55) should have resolved the issue.

4rnaudB (approved):

PR #520, opened 2024-03-21
Add unwrap CLI

po09i:
## Description
This PR adds a CLI to unwrap NIfTIs. The motivation behind this CLI is that some manufacturers directly output field maps from the scanner. These field maps can have wraps which require unwrapping. This CLI aims at unwrapping data (not just phase data).

Prelude requires data to be in the -pi to pi range. The data is scaled to that range, phase unwrapping is performed then scaled back to the initial range. 

Option 1: If possible, provide the unit of the data (Hz, T, G, etc) and the echo time. This will allow to unambiguously scale the data and allow unwrapping.
Option 2: Providing the --range option
Providing the --range of the data is all that is necessary to unambiguously unwrap the data. For datasets that are dimensionless (Siemens phase difference [0, 4095]), this is the only option possible.

Similar features to the prepare_fieldmap pipeline are provided (mask resampling, possibility to input a threshold to calculate the mask, 2npi offset fix, 4d unwrapping)

## How to use example
```
st_unwrap -i /path/to/wrapped.nii.gz --mag /path/to/input.nii.gz -o /path/to/output.nii.gz --unit hz --dte 0.001
```
## Note
There could be a 3rd option that would not require additional inputs from the user. The "range" is calculated from the max() and min() of the data. However, this results in compounding errors if data.max() - data.min() != range. I opted to not allow this option.

## Linked issues
Related to #519

### DIFF
--- a/shimmingtoolbox/cli/unwrap.py
+++ b/shimmingtoolbox/cli/unwrap.py
@@ -1,0 +1,123 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*
+
+import click
+import json
+import logging
+import math
+import nibabel as nib
+import numpy as np
+import os
+
+from shimmingtoolbox.unwrap.unwrap_phase import unwrap_phase
+from shimmingtoolbox.prepare_fieldmap import get_mask, correct_2pi_offset, VALIDITY_THRESHOLD
+from shimmingtoolbox.utils import create_fname_from_path, set_all_loggers, create_output_dir, save_nii_json
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+FILE_OUTPUT_DEFAULT = 'unwrapped.nii.gz'
+MASK_OUTPUT_DEFAULT = 'mask_unwrap.nii.gz'
+
+CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
+
+
+@click.command(
+    context_settings=CONTEXT_SETTINGS,
+)
+@click.option('-i', '--input', 'fname_data', type=click.Path(exists=True), required=True,
+              help="Input path of data nifti file")
+@click.option('--mag', 'fname_mag', type=click.Path(exists=True), required=True,
+              help="Input path of mag nifti file")
+@click.option('--unwrapper', type=click.Choice(['prelude', 'skimage']), default='prelude',
+              show_default=True,
+              help="Algorithm for unwrapping. skimage is installed by default, prelude requires FSL to be installed.")
+@click.option('-o', '--output', 'fname_output', type=click.Path(),
+              default=os.path.join(os.curdir, FILE_OUTPUT_DEFAULT),
+              show_default=True, help="Output filename for the unwrapped data, supported types : '.nii', '.nii.gz'")
+@click.option('--mask', 'fname_mask', type=click.Path(exists=True),
+              help="Input path for a mask.")
+@click.option('--threshold', 'threshold', type=float, show_default=True, default=0.05,
+              help="Threshold for masking if no mask is provided. Allowed range: [0, 1] where all scaled values lower "
+                   "than the threshold are set to 0.")
+@click.option('--savemask', 'fname_save_mask', type=click.Path(),
+              help="Filename of the mask calculated by the unwrapper")
+@click.option('-v', '--verbose', type=click.Choice(['info', 'debug']), default='info',
+              help="Be more verbose")
+def unwrap_cli(fname_data, fname_mag, unwrapper, fname_output, fname_mask, threshold, fname_save_mask, verbose):
+    """
+    Unwraps images. This algorithm expects the input to have wraps. Edge cases might occur if no wraps are present.
+    The unwrapper tries to correct 2pi ambiguity when unwrapping by bringing the mean closest to 0 in increments of 2pi
+    jumps.
+    """
+    # Set logger level
+    set_all_loggers(verbose)
+
+    # Create filename for the output if it's a path
+    fname_output_v2 = create_fname_from_path(fname_output, FILE_OUTPUT_DEFAULT)
+
+    # Create output directory if it doesn't exist
+    create_output_dir(fname_output_v2, is_file=True)
+
+    # Save mask
+    if fname_save_mask is not None:
+        # If it is a path, add the default filename and create output directory
+        fname_save_mask = create_fname_from_path(fname_save_mask, MASK_OUTPUT_DEFAULT)
+        create_output_dir(fname_save_mask, is_file=True)
+
+    # Scale to radians
+    nii_data = nib.load(fname_data)
+    data = nii_data.get_fdata()
+
+    # Scale the input from -pi to pi
+    # If the input is wrapped, the unwrapper will unwrap normally, if not, no unwrapping will be done
+    scalar = get_scalar_to_fit_2pi(data)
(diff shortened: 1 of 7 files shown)


shimmingtoolbox/cli/unwrap.py
```diff
@@ -0,0 +1,188 @@
...
+
+    # Load data
+    nii_data = nib.load(fname_data)
+    data = nii_data.get_fdata()
+
+    # Convert 'unit' to upper()
+    if unit is not None:
+        unit_up = unit.upper()
+    else:
+        unit_up = None
```
4rnaudB: else = unit is None no?

shimmingtoolbox/cli/unwrap.py
```diff
@@ -0,0 +1,188 @@
...
+    units_upper = [a_unit.upper() for a_unit in ALLOWED_UNITS]
+
+    # Scale the input from -pi to pi
+    if unit_up in units_upper and dte is not None:
```
4rnaudB: Could be merged to the if statement above since we know unit_up will be in units_upper if it is not None (cli Choice)

shimmingtoolbox/cli/unwrap.py
```diff
@@ -0,0 +1,188 @@
...
+            # G
+            scalar = 2 * math.pi * dte * H_GYROMAGNETIC_RATIO * 1e-4
+
+    elif extent is not None:
+        logger.info(f"Scaling the input data to radians using the --range: {extent}.")
+        scalar = _get_scalar_to_fit_2pi(data, extent)
+
+    else:
+        raise ValueError("Neither --unit and --dte nor --range were provided."
+                         "Use whichever is more convenient for you.")
```
4rnaudB, line 112: Already checked from the cli

shimmingtoolbox/cli/unwrap.py
```diff
@@ -0,0 +1,188 @@
...
+            scalar = 1 * dte
+        elif unit_up == units_upper[2]:
+            # T
+            scalar = 2 * math.pi * dte * H_GYROMAGNETIC_RATIO
+        elif unit_up == units_upper[3]:
+            # mT
+            scalar = 2 * math.pi * dte * H_GYROMAGNETIC_RATIO * 1e-3
+        elif unit_up == units_upper[4]:
+            # G
+            scalar = 2 * math.pi * dte * H_GYROMAGNETIC_RATIO * 1e-4
```
4rnaudB: Suggestion: Create a separate script called conversion.py with all the conversion functions and change ALLOWED_UNITS to a dictionary where each unit has its conversion function.

shimmingtoolbox/cli/unwrap.py
```diff
@@ -0,0 +1,188 @@
...
+    # Log output file
+    logger.info(f"Filename of the unwrapped data is located: {fname_output_v2}")
+
+
+def _get_scalar_to_fit_2pi(data, extent=None):
+    """Return the scalar that scales the data to a range of 2pi"""
+
+    if extent is None:
+        extent = np.max(data) - np.min(data)
+        logger.debug(f"--range not provided, using max() - min() to calculate the range: {extent}")
```
4rnaudB, line 182: Is it possible to not provide the range or the TEs?

shimmingtoolbox/cli/unwrap.py
4rnaudB: else = unit_up is None

shimmingtoolbox/cli/unwrap.py
4rnaudB: Should be combined with the if statement above since the cli handles the choices already

shimmingtoolbox/cli/unwrap.py
```diff
@@ -0,0 +1,188 @@
...
+            scalar = 2 * math.pi * dte * H_GYROMAGNETIC_RATIO * 1e-3
+        elif unit_up == units_upper[4]:
+            # G
+            scalar = 2 * math.pi * dte * H_GYROMAGNETIC_RATIO * 1e-4
+
+    elif extent is not None:
+        logger.info(f"Scaling the input data to radians using the --range: {extent}.")
+        scalar = _get_scalar_to_fit_2pi(data, extent)
+
+    else:
```
4rnaudB, line 110: Shouldn't be possible

shimmingtoolbox/cli/unwrap.py
4rnaudB: Suggestions: create a new Python file (e.g. conversions.py) with all the conversion functions and change ALLOWED_UNITS to a dictionary with each unit having a conversion function.

shimmingtoolbox/cli/unwrap.py
4rnaudB, line 182: Shouldn't be possible to end up in this if

shimmingtoolbox/prepare_fieldmap.py
```diff
@@ -235,7 +252,15 @@ def correct_2pi_offset(unwrapped, mag, mask, validity_threshold):
     validity_masks = np.logical_and(mask, validity_masks)
 
     if unwrapped.ndim == 4:
-        for i_time in range(1, unwrapped_cp.shape[3]):
+        for i_time in range(0, unwrapped_cp.shape[3]):
+            # Correct the first time point and bring closest to the mean
+            if i_time == 0:
+                unwrapped_cp[..., i_time] = correct_2pi_offset(unwrapped_cp[..., i_time],
```
4rnaudB, line 258: correct_2pi offset already handles 4D data

4rnaudB (commented):
Tested with a brain field map and correctly got the same output as the input

shimmingtoolbox/cli/unwrap.py
po09i: I'd rather keep these to make it explicit that it should not happen. As a bonus, if we use this as an API then it will already be set up.

shimmingtoolbox/cli/unwrap.py
po09i: same comment as above

shimmingtoolbox/cli/unwrap.py
po09i: Done in [8082d7d](https://github.com/shimming-toolbox/shimming-toolbox/pull/520/commits/8082d7d79ab10aa54c8c5e4e7fbabd77db53ac43) and [598b980](https://github.com/shimming-toolbox/shimming-toolbox/pull/520/commits/598b980e9b3a151b3a232aeafd0c211cef323e71)

shimmingtoolbox/cli/unwrap.py
4rnaudB: Makes sense if we want to use it as an API

shimmingtoolbox/prepare_fieldmap.py
po09i: Indeed, 4d data will all have an offset between timepoints that is less than 2npi (this was already the case). However, this makes sure that the first timepoints has "n" that results in the first volume closest to the mean as possible.

shimmingtoolbox/cli/unwrap.py
po09i: Resolved

4rnaudB (approved):

po09i:
Here are the results when used on wrapped data:


Unwrapped:

po09i:
The PR was tested by @gxu32 without issue

PR #523, opened 2024-04-25
Solve dcm2niix not in PATH issue and improve st_prepare_fieldmap documentation

### LINKED ISSUE #518: dcm2niix not found if conda environment is not activated
### Environment details

- [ ] I followed the [installation instructions](https://shimming-toolbox.org/en/latest/2_getting_started/1_installation.html)

### Bug Description
dcm2niix is not found if the conda environment is not activated. The error persists even after using source `$HOME/shimming-toolbox/python/etc/profile.d/conda.sh`. Only running `conda activate ~/shimming-toolbox/python` before using any scripts using dcm2niix prevents the error from happening.

po09i:
## Description
`Dcm2bids` calls `dcm2niix` through a subprocess call. When calling subprocess, the environment (mainly the `PATH` shell variable) used is the one that was set when calling the shell function (`st_dicom_to_nifti`), not the environment where the shimming toolbox is installed. That creates an issue where `dcm2niix` is installed in the Shimming Toolbox environment but the `PATH` is not set up correctly when called by `dcm2bids`. To remedy this, I added `dcm2niix` to the launchers. As a side effect, this makes Shimming Toolbox's `dcm2niix` install exposed to the shell, which might be unwanted for some that already have it installed and require specific versions. (I could call it `st_dcm2niix`, but I would need to change some `dcm2bids` source code to make it work making the process of updating `dcm2bids` a hassle). Let me know what you think.

- Improved `st_prepare_fieldmap` documentation

Fixes #518


shimmingtoolbox/dicom_to_nifti.py
```diff
@@ -58,6 +58,13 @@ def dicom_to_nifti(path_dicom, path_nifti, subject_id='sub-01', fname_config_dcm
     # Copy original dicom files into nifti_path/sourcedata
     copy_tree(path_dicom, os.path.join(path_nifti, 'sourcedata'))
 
+    # Update the PATH environment variable to include the dcm2niix executable
+    # TODO: Try this out on Windows, path could be Scripts instead of bin
+    if 'ST_DIR' in os.environ and os.path.exists(os.environ['ST_DIR']):
+        os.environ['PATH'] = os.path.join(os.environ['ST_DIR'], 'python', 'bin') + os.pathsep + os.environ['PATH']
+    else:
+        logger.warning("Environment variable ST_DIR not found. Using default path.")
```
joshuacwnewton, line 66: >  To remedy this, I added dcm2niix to the launchers. As a side effect, this makes Shimming Toolbox's dcm2niix install exposed to the shell, which might be unwanted for some that already have it installed and require specific versions.

Does it expose `dcm2niix` in the shell? From what I understand, changes to `os.environ['PATH']` are only relevant within the Python process. The underlying `PATH` variable doesn't actually get modified, I think!

Here's a quick demo:

image

po09i: This is also my experience. Running `st_dicom_to_nifti` did not affect the PATH in the shell. Thanks!

4rnaudB (approved):
Problem fixed!
image
image

PR #530, opened 2024-06-28
Update how constraints are handled

po09i:
## Description
- Raise an error if the frequency is not within bounds for scanners with constraints
- Add validity for constraints for each order

### DIFF
--- a/shimmingtoolbox/shim/shim_utils.py
+++ b/shimmingtoolbox/shim/shim_utils.py
@@ -341,13 +341,12 @@ class ScannerShimSettings:
 
     def concatenate_shim_settings(self, orders=[2]):
         coefs = []
-        if not self.shim_settings['has_valid_settings']:
-            logger.warning("Invalid Shim Settings")
-            return coefs
 
         if any(order >= 0 for order in orders):
             for order in sorted(orders):
                 if self.shim_settings.get(str(order)) is not None:
+                    if not self.shim_settings[f'order{order}_is_valid']:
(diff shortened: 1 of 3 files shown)


shimmingtoolbox/shim/shim_utils.py
```diff
@@ -308,6 +318,19 @@ def dac_to_shim_units(manufacturer, manufacturers_model_name, shim_settings):
...
 
 
+def _convert_to_ui_units(shim_settings_coefs, scanner_constraints, scanner_constraints_dac):
+    # Convert to ui units
+    coefs_dac = shim_settings_coefs
+    max_coef_ui = np.array([cst[1] for cst in scanner_constraints])
+    min_coef_ui = np.array([cst[0] for cst in scanner_constraints])
+    coefs_ui = (np.array(coefs_dac) * max_coef_ui / np.array(scanner_constraints_dac))
+    tolerance = 0.001 * max_coef_ui
+    if np.any(coefs_ui > (max_coef_ui + tolerance)) or np.any(coefs_ui < (min_coef_ui - tolerance)):
```
4rnaudB, line 328: Shouldn't it be coefs + tolerance > max_coef and coefs - tolerance < min_coefs?

po09i: I believe it is correct how it is right now. The tolerance is on the bounds so that there is no unnecessary errors.

4rnaudB (approved):

PR #541, opened 2024-07-31
Add versions and a changelog

### LINKED ISSUE #528: Versioning strategy for Shimming Toolbox
### Context
Shimming Toolbox is set up between 2 repos
- [Shimming Toolbox](https://github.com/shimming-toolbox/shimming-toolbox)
- [FSLeyes plugin](https://github.com/shimming-toolbox/fsleyes-plugin-shimming-toolbox)

Currently, a user downloads the main branch of the plugin and runs the plugin installer. The installer downloads and installs a [specific commit](https://github.com/shimming-toolbox/fsleyes-plugin-shimming-toolbox/blob/c58065807653a5e98cacc756d7fc041815fb7875/installer/install_shimming_toolbox.sh#L17) of Shimming Toolbox.

We want to be able to version Shimming Toolbox. It would also make sense to have a version of the plugin as well.

po09i:
## Description
This PR adds package versions to both Shimming Toolbox and the plugin. It updates both versions to 1.0. It also adds a git version in the CLI (st_check_dependencies and st_dump_env_info). CHANGES.md files were also added to keep track of the changes.

Todo:
- [x] Update the PR template to assign labels

## Testing
You can look at read the docs previewing the website in the CI links to see what it will look like.

## Linked issues
Fixes #528


CHANGES.md
```diff
@@ -0,0 +1,373 @@
...
+ - **st_prepare_fieldmap**: Resolve prelude error if --savemask directory is not already created when field mapping. [View pull request](https://github.com/shimming-toolbox/shimming-toolbox/pull/448)
+ - **st_prepare_fieldmap**: Change phase range in field mapping calculations. [View pull request](https://github.com/shimming-toolbox/shimming-toolbox/pull/499)
+ - **st_realtime_shim**: Fix realtime shimming mask resampling modes. [View pull request](https://github.com/shimming-toolbox/shimming-toolbox/pull/289)
+ - **st_realtime_shim**: gradient realtime shim output coordinate system. [View pull request](https://github.com/shimming-toolbox/shimming-toolbox/pull/310)
+ - **st_realtime_shim**: Gradient realtime_shim orientation and nan problems. [View pull request](https://github.com/shimming-toolbox/shimming-toolbox/pull/323)
+ - Solve launchers needing to be launched from the ST venv. [View pull request](https://github.com/shimming-toolbox/shimming-toolbox/pull/346)
+ - Restrict numpy version < 2.x. [View pull request](https://github.com/shimming-toolbox/shimming-toolbox/pull/527)
```
4rnaudB: Numpy is not restricted anymore (fixed in signal recovery PR)

fsleyes-plugin-shimming-toolbox/CHANGES.md
```diff
@@ -0,0 +1,67 @@
+# Changelog
```
4rnaudB, line 1: necessary to have this + changes.md?

po09i: No, but  since it's a different package, it makes sense for it to have its own CHANGES.md file. And setting it up now makes it easier in the future.

CHANGES.md
po09i: The signal recovery PR was not included in the changelog because it was missing a label. I added it back. It appears in the "feature" section though. 

Also, the changes are not exactly in chronological order, they are in order of CLIs and PR numbers so it can be confusing when looking at PRs that occurred over 3 years. If we have releases that occur more frequently, it should be less confusing.

4rnaudB: But shouldn't we remove changes that are not applied in the current state of the code?

po09i: Since it got removed and added back in the same version, I think it makes sense. I removed it in [ba97c7e](https://github.com/shimming-toolbox/shimming-toolbox/pull/541/commits/ba97c7e18aba3f16c276d9e739b23dd073e82bae)

4rnaudB (approved):

PR #550, opened 2024-09-06
Add CLI to combine text files

po09i:
## Description
We sometimes perform different optimizations one after another. This leaves us with different text files that need to be combined. In a perfect world, we would optimize all at once, but for now a CLI that adds the same format text files and one that combines the different text files is useful.

add_shim_coefs:
- Adds coefficients
- Must have the same number of channels and shim lines

convert_shim_coefs_format:
- Convert between 'volume', 'slicewise', 'chronological'
- Adds a output conversion of `custom-cl` text file specifically created to output in a required format and scaling coefficients
- Ability to reverse the slice encoding direction
- Ability to add channels in the text file where the coefficients are set to 0s

## Todo in a subsequent PR
A conversion between fat sat and no fat sat format could be done in the future


shimming-toolbox/shimmingtoolbox/cli/b0shim.py
```diff
@@ -1460,7 +1460,152 @@ def max_intensity(fname_input, fname_mask, fname_output, verbose):
...
+@click.option('-i', '--input', 'fname_input', nargs=1, type=click.Path(exists=True), required=True,
+              help="Text file containing the shim coefficients. Supported formats: .txt")
+@click.option('-i2', '--input2', 'fname_input2', nargs=1, type=click.Path(exists=True), required=True,
+              help="Text file containing the shim coefficients. Supported formats: .txt")
+@click.option('--input-file-format', 'i_format',
+              type=click.Choice(['slicewise', 'chronological', 'custom-cl']), required=True,
+              help="Syntax used to describe the sequence of shim events for a coil or coil channel. "
+                   "Use 'slicewise' to output in row 1, 2, 3, etc. the shim coefficients for slice "
+                   "1, 2, 3, etc. Use 'chronological' to output in row 1, 2, 3, etc. the shim value "
+                   "for trigger 1, 2, 3, etc. The trigger is an event sent by the scanner and "
```
4rnaudB: Is chronological still one text file with one line being all coefficients for the coil? if there is one trigger per slice, is chronological  the same as slicewise

4rnaudB: Also, what is custom-cl

shimming-toolbox/shimmingtoolbox/cli/b0shim.py
```diff
@@ -1460,7 +1460,152 @@ def max_intensity(fname_input, fname_mask, fname_output, verbose):
...
+    with open(fname_input, 'r') as f:
+        coefs1 = []
+        for i_line, line in enumerate(f):
+            list_line = line.strip('\n').split(',')
+            temp = []
+            for i, value in enumerate(list_line):
+                if value.strip(' ') != '':
+                    temp.append(float(value.strip()))
+            coefs1.append(temp)
+        coefs1 = np.array(coefs1)
```
4rnaudB: I would consider this function to read data files:
https://github.com/shimming-toolbox/Teensy-communication/blob/0f5b7d11673ba0ddfdfe3a04baf6a361a2b2d740/Python_communication/src/shimming_toolbox_teensy/teensyComLib.py#L188-L225

shimming-toolbox/shimmingtoolbox/cli/b0shim.py
```diff
@@ -1460,7 +1460,152 @@ def max_intensity(fname_input, fname_mask, fname_output, verbose):
...
+                        else:
+                            coefs[i_shim][i_channel] = coefs1[i_shim][i_channel]
+                else:
+                    for i_channel in range(n_channels2):
+                        if i_channel < n_channels1:
+                            coefs[i_shim][i_channel] = coefs1[i_shim][i_channel] + coefs2[i_shim][i_channel]
+                        else:
+                            coefs[i_shim][i_channel] = coefs2[i_shim][i_channel]
+        else:
+            raise ValueError("The number of shim events is not the same in both text files")
```
4rnaudB: I would avoid this altogether. Basically, remove the file format input and force them to give two identical formatted file. You could do a second function that converts formatted files instead.

shimming-toolbox/shimmingtoolbox/cli/b0shim.py
po09i: Chronological is in the order that the slices are acquired on the scanner chronologically. This means that if there is one trigger per slice, but a slice order that is interleaved, it will not be the same as slice-wise but will have the same number of lines.

Custom-cl is the custom format used by our GE collaborator.

shimming-toolbox/shimmingtoolbox/cli/b0shim.py
po09i: It's essentially a convenience for adding volume shim and slice-wise shim. I could convert this into its own API function, but I don't see a benefit in breaking it in 2 CLIs if that is what you suggest?

4rnaudB: I understand it is more practical for this specific use, but I feel that in a broader view of how it will be used, it is better to make one function convert text file formats and one function combine those text files

shimming-toolbox/shimmingtoolbox/cli/b0shim.py
4rnaudB: I would add it in the description in the cli inputs

shimming-toolbox/shimmingtoolbox/cli/b0shim.py
po09i: I kept the code I had but converted it into a function inL [8d65241](https://github.com/shimming-toolbox/shimming-toolbox/pull/550/commits/8d652415419d811fd9a2ed737a886453b9212875)

shimming-toolbox/shimmingtoolbox/cli/b0shim.py
po09i: I made 2 CLIs, one to add coefficients and one to convert between formats ([8d65241](https://github.com/shimming-toolbox/shimming-toolbox/pull/550/commits/8d652415419d811fd9a2ed737a886453b9212875))

shimming-toolbox/test/cli/test_cli_b0shim.py
```diff
@@ -1622,22 +1613,194 @@ def test_combine_shim_coefs_sl(self):
                 f.write("1,2,3, 4,\n")
                 f.write("5,6,7, 4\n")
```
4rnaudB: Shouldn't it be counted as the same number of channels?

shimming-toolbox/shimmingtoolbox/cli/b0shim.py
```diff
@@ -1461,112 +1461,207 @@ def max_intensity(fname_input, fname_mask, fname_output, verbose):
...
+
+    if coefs1.shape == coefs2.shape:
+        coefs = coefs1 + coefs2
+    else:
+        raise ValueError("The number of shim events and/or the number of channels is not the same in both text files")
+
+    logger.debug(coefs1)
+    logger.debug(coefs2)
+    logger.debug(coefs)
+    write_coefs_to_text_file(coefs, fname_output, 'slicewise')
```
4rnaudB, line 1492: Always slice-wise?

shimming-toolbox/test/cli/test_cli_b0shim.py
po09i: Not anymore, the format needs to be the same now.

shimming-toolbox/shimmingtoolbox/cli/b0shim.py
po09i: Yes, this could be either 'slicewise' or 'chronological' since the function has no difference when writing them, the difference is in the ordering of the input data. Since the format needs to be the same, we just rewrite in the same order

shimming-toolbox/test/cli/test_cli_b0shim.py
4rnaudB: There is a format with an extra comma and one without?

po09i: Yes, it's not necessary, it's testing that it does not throw an error based on that. It should be able to parse the input whether there are spaces and commas at the end (or no comma)

4rnaudB (changes requested):
when changing from slice-wise to chronological, the coefs values got changed from regular notation (0.00005) to scientific notation (5e-05) which would be problematic. Same problem with add_coefs function.

po09i:
> when changing from slice-wise to chronological, the coefs values got changed from regular notation (0.00005) to scientific notation (5e-05) which would be problematic. Same problem with add_coefs function.

Should be fixed in the latest commit [8d66e11](https://github.com/shimming-toolbox/shimming-toolbox/pull/550/commits/8d66e111c1d56eeb24e77f0b354cb7274d1f89ab)

4rnaudB (approved):

PR #552, opened 2024-09-30
Update documentation configuration file to support ReadTheDocs addons migration

po09i:
## Description
ReadTheDocs is moving its configuration options in an "addons" tab on RTD rather than asking the users to define their options in the `conf.py` file. They provided a code snippet to add that should provide a smooth transition. The changes will be enabled by default on October 7th 2024. 

[Here](https://about.readthedocs.com/blog/2024/07/addons-by-default/) is more information


4rnaudB:
Maybe add https://github.com/readthedocs/sphinx-build-compatibility

4rnaudB:
Fixes the build issue we encountered previously #554

4rnaudB (approved):

PR #555, opened 2024-10-21
Update signal recovery calculation

4rnaudB:
## Description
Previous signal loss maps were computed as follows:
phi = gradient * epi_TE, with the gradient in Hz

But phi should be
phi = 2pi * gradient / B0_map_thickness * epi_slice_thickness * epi_TE, with gradient in Hz

Note, in the code a  /pi was added because np.sinc() multiplies input by pi.

### DIFF
--- a/shimming-toolbox/shimmingtoolbox/shim/sequencer.py
+++ b/shimming-toolbox/shimmingtoolbox/shim/sequencer.py
@@ -41,6 +41,7 @@ supported_optimizers = {
     'pseudo_inverse': Optimizer
 }
 
+GAMMA = 42.576E6  # in Hz/Tesla
 
 class Sequencer(object):
     """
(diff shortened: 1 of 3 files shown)


shimming-toolbox/shimmingtoolbox/shim/sequencer.py
```diff
@@ -820,29 +821,29 @@ def calc_shimmed_anat_orient(self, coefs, list_shim_slice):
         nib.save(nii_shimmed_anat_orient, fname_shimmed_anat_orient)
 
     def _plot_static_signal_recovery_mask(self, unshimmed, shimmed_Gz, mask):
-    # Plot signal loss maps
-        def calculate_signal_loss(B0_map):
-            G = np.gradient(B0_map, axis = 2)
-            signal_map = abs(np.sinc(self.epi_te * G))
+        # Plot signal loss maps
+        def calculate_signal_loss(gradient):
+            slice_thickness = self.nii_anat.header['pixdim'][3]
```
po09i: Does this include slice gap?

4rnaudB: no, you're right I looked at an acquisition of 1mm with a slice gap of 1mm and pixdim[3] is 2

4rnaudB: I don't see any parameter that gives the correct value though

po09i: The JSON file should include a parameter that has slice gap information.

po09i (approved):

PR #571, opened 2025-01-20
Add the ability to set the current shim coefficient values as inputs in the constraint file

po09i:
## Description
This PR adds the ability to set the current values of the shim coefficients so that the constraints can be respected when optimizing. In the case of scanner coils, the resulting coefficients are also updated in the bids JSON sidecar of the computed calculated field map.

Verify that Philips ImagingFrequency tag is expected


shimming-toolbox/shimmingtoolbox/coils/coil.py
```diff
@@ -138,32 +185,98 @@ def profile(self, profile):
...
+                    skipping_coef_sum_max_error_check = True
+                    break
+            if not skipping_coef_sum_max_error_check:
+                # Log if the current coefs are outside the total sum constraint
+                total_current = 0
+                for key in constraints["coefs_used"]:
+                    total_current += sum(np.abs(constraints["coefs_used"][key]))
+                if total_current > constraints["coef_sum_max"]:
+                    logger.warning(f"Sum of the current coefficients is greater than the maximum allowed value: "
+                                   f"{constraints['coef_sum_max']}")
```
4rnaudB: Not sure I see the use of this section before the optimization

po09i: Good point, fixed in [c113eaf](https://github.com/shimming-toolbox/shimming-toolbox/pull/571/commits/c113eaffd1f00a5f308bc52b615f24e2658a7973)

4rnaudB (approved):

4rnaudB (approved):

PR #578, opened 2025-02-07
Prerelease fixes

4rnaudB:
## Description
Fixes two issues found while testing before 1.2 release

1.  When using w = 0, the optimizer would not do signal recovery optimization. This is now fixed.
2.  The json data was outputted for every shim. THis output has been removed.


po09i (commented):
I don't see the fix where the JSON does not get outputted

4rnaudB:
The second change didn't push for some reason

po09i (approved):

PR #582, opened 2025-02-11
Update BIDS config file

po09i:
## Description
- Add multi-echo MPRAGE sequence to BIDS config file
- Add "a_gre_dyn_realtime_shimming_TTL" as a field map
- Use `shutil` instead of `distutil` when using `copy_tree` in st_dicom_to_nifti


4rnaudB (approved):
worked on my data

PR #585, opened 2025-02-18
add requirement.txt to setup files in preparation for release automa…

4rnaudB:
## Description
Add the possibility to use a requirements.txt file for the download. This will allow the automatization of creating a new release. 

## Related issues
Todo in a subsequent PR: #588


fsleyes-plugin-shimming-toolbox/setup.py
```diff
@@ -8,13 +8,20 @@
 with open(path_version) as f:
     version = f.read().strip()
 
+requirements_path = path.join(path.dirname(here), "requirements_fsleyes.txt")
```
po09i:
```suggestion
requirements_path = path.join(path.dirname(here), "requirements_stplugin.txt")
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
...
+        with:
+          python-version: "3.10"
+
+      - name: Install Only Specified Dependencies
+        run: |
+          python -m venv venv
+          source venv/bin/activate
+          pip install --upgrade pip
+          pip install -r ./shimming-toolbox/requirements_st.txt
+          pip install -r ./fsleyes-plugin-shimming-toolbox/requirements_stplugin.txt
```
po09i: I think that installing Shimming Toolbox in a different way than the provided way of installing it (make install) is bound to create issues. As an example right now, fsleyes and its requirements are not in the pinned dependency list. Unless there was another reason I'm not thinking, we should probably use "make install".

I also just noticed that fsleyes seems to be on [Pypi](https://pypi.org/project/fsleyes/). We should include it in the list of dependencies and remove it from the conda install of the installer. Maybe that's another PR though.

4rnaudB: Fixed with [900b0cc](https://github.com/shimming-toolbox/shimming-toolbox/pull/585/commits/900b0cc221fd14c2e42ed02ed38828a9f565fa8f) and [e08ba2f](https://github.com/shimming-toolbox/shimming-toolbox/pull/585/commits/e08ba2fe4edef2d44afdba1eb5bc097f5752b2f4)

4rnaudB: I think we should do a separate PR for fsleyes. I can make it after this PR is closed.

po09i: I opened an issue (#588)

po09i (approved):

PR #595, opened 2025-03-13
Add fix when dicom files have the same name

po09i:
## Description
If the input DICOMS are in different folders (using -r option) and that they have the same name, they will be copied in the same directory and will create a name conflict (only one of the two files will be copied). To fix that, we check that the output file does not already exist, if it does, we append a suffix.


4rnaudB (approved):
I though the os doesn't allwo to have two folders with the same name?

po09i:
They could have the same name if they are not in the same folder.

PR #607, opened 2025-06-04
Dockerize ST for scanner

### LINKED ISSUE #597: Dockerize ST for scanner

AntoineGuenette:
## Checklist

- [x] I've applied the relevant labels to this PR. (You need to choose: the scope of the PR, 1 or more (pale brown), the main reason for the PR, only 1 (dark purple) and if applicable, the CLI affected (dark cyan))

## Description
This PR introduces a Dockerfile that enables the installation of Shimming Toolbox (ST) and all its dependencies directly on the MRI console. The goal is to eliminate the need for an external laptop to run ST, facilitating a self-contained and portable deployment.

## Performances
As of [931110d](https://github.com/shimming-toolbox/shimming-toolbox/pull/607/commits/931110d7085c582935c82b584a17874bfba287e2), building the image takes approximately 9 minutes. Here are the details on the space needed after each step  : 

1. After building the image from the Dockerfile
    - Compressed size of the image : 3.32 GB
        - Includes only the image itself
        - It corresponds to the size shown by Docker Desktop and the command `docker image ls` 
    - Uncompressed size of the image: 8.352 GB
        - Includes all extracted files from all layers, as actually stored on disk by Docker
        - It corresponds to the size shown with the command `docker system df`
        - Using the `docker history <image_id>` command, we can see the actual space taken by each layer of the image on the disk : 
            - Dependecies (from `apt-get install`): 866 MB
            - SCT : 5.26 GB
            - ST : 2.11 GB
            - FSL prelude : 30.7 MB
2. After running a container from the image
    - Compressed size of the image : 11.67 GB
    - Uncompressed size of the image : 16.62 GB
3. After exporting the filesystem of the container in a tarball (.tar file)
    - Size of the tarball : 7.99 GB
4. After creating the chroot image (.img file)
    - Size of the chroot image : 9.54 GB
    - Size of the zip file containing the chroot image : 3.3 GB
    - It's the zip file that is exported to the MARS system

> [!WARNING]
> All documents have been moved to the [siemens-fire](https://github.com/shimming-toolbox/siemens-fire) repository. Please refer to that repository for the most up-to-date versions of the files.


shimming-toolbox/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+# Install system dependencies
+RUN apt-get update && \
+    apt-get install -y git build-essential
+
+# Install ST
+RUN mkdir /ST_code && \
+    git clone https://github.com/shimming-toolbox/shimming-toolbox.git /ST_code
+
+# Install python dependencies
+RUN pip3 install -r ST_code/shimming-toolbox/requirements_st.txt
```
po09i: That's not the "proper" way to install ST, we usually use "make install". There are a few things (such as copying the config files and creating launchers) that are done in that script.

AntoineGuenette: Changed in commit [fa0f999](https://github.com/shimming-toolbox/shimming-toolbox/pull/607/commits/fa0f9994c53abc06a1e328e5d74a7999f796bcdb).

shimming-toolbox/Dockerfile
```diff
@@ -0,0 +1,35 @@
+FROM ubuntu:latest
```
4rnaudB: Bad idea to get the latest version as an update could create issues without even changing the code. A fixed version would make the Docker image more stable.

shimming-toolbox/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+                    python3 python3-pip libglib2.0-0 libgl1 libxrender1 libxkbcommon-x11-0 libdbus-1-3 && \
+    apt-get clean
+
+# Install FSL
+RUN curl -Ls https://fsl.fmrib.ox.ac.uk/fsldownloads/fslconda/releases/getfsl.sh | sh -s
+
+# Install SCT
+RUN mkdir -p code/SCT && \
+    git clone https://github.com/spinalcordtoolbox/spinalcordtoolbox.git /code/SCT && \
+    cd /code/SCT && \
```
4rnaudB: same here, pin a SCT version with -b

shimming-toolbox/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+    ./install_sct -y
+
+# Install ST
+RUN cd .. && \
+    git clone https://github.com/shimming-toolbox/shimming-toolbox.git /code/ST  && \
+    cd /code/ST && \
```
4rnaudB: This should also point to the latest release

4rnaudB (changes requested):

shimming-toolbox/Dockerfile
AntoineGuenette: Changed to ubuntu 24.04 (supported until april 2036) in commit [23af0cb](https://github.com/shimming-toolbox/shimming-toolbox/pull/607/commits/23af0cbe874300c8574e6e8581aa32f1a8cfe970).

shimming-toolbox/Dockerfile
AntoineGuenette: Changed to version 7.0 in commit [23af0cb](https://github.com/shimming-toolbox/shimming-toolbox/pull/607/commits/23af0cbe874300c8574e6e8581aa32f1a8cfe970).

shimming-toolbox/Dockerfile
AntoineGuenette: Changed to version v1.2 in commit [23af0cb](https://github.com/shimming-toolbox/shimming-toolbox/pull/607/commits/23af0cbe874300c8574e6e8581aa32f1a8cfe970). 

TO BE FIXED : The installation crashes because of an update to dcm2niix.

AntoineGuenette: Fixed in [2a4fc0d](https://github.com/shimming-toolbox/shimming-toolbox/pull/607/commits/2a4fc0dfb9bfd308ff0f4c05df288d7d31a5e6b6).

jcohenadad:
> 28 GB of disk space, as it includes FSL, SCT, and ST.

wow! what's the approximate distribution of sizes? Can we propose a lighter install for some software? eg do we need all of FSL? If we only need prelude, you could just download it from our ST organization

jcohenadad:
@AntoineGuenette I suggest you try to look for lighter Linux distros in your Docker image. @SomeoneInParticular might have ideas

SomeoneInParticular:
I doubt its the OS thats the problem here; even modern Ubuntu's Docker image only takes up [~150MB of storage](https://hub.docker.com/_/alpine) (scroll down for the comparison). Likewise, SCT and ST are not that large; the former is 2GB in an Apptainer (which has a comparable compression regime as Docker), and I can't imagine ST is any worse.

That leaves more than 20GB laid solely on FSL. As @jcohenadad suggested, if there is a way to trim it down to include only its needed parts, that would be ideal. Alternatively, you could have two different docker images; ST-standalone (for us command-line users), and a "full" install w/ FSLeyes for use with the GUI.

In theory, we could also make a "wrapper" for the former Docker image, which would allow the user to use it with an existing FSLeyes install (saving us the hassle of shipping it ourselves). Thats just my intuition though; I haven't had the chance to look at FSLeyes plugin handling yet to confirm how feasible this would be.

jcohenadad:
> I doubt its the OS thats the problem here; even modern Ubuntu's Docker image only takes up [~150MB of storage](https://hub.docker.com/_/alpine)

I was referring to conversation with @AntoineGuenette who mentioned about 3GB for the OS-- hence my comment

AntoineGuenette:
> > 28 GB of disk space, as it includes FSL, SCT, and ST.
> 
> wow! what's the approximate distribution of sizes? Can we propose a lighter install for some software? eg do we need all of FSL? If we only need prelude, you could just download it from our ST organization

The installation time and space taken by the image has been largely reduced with commit [6cf950c](https://github.com/shimming-toolbox/shimming-toolbox/pull/607/commits/6cf950caebcee99bca81f15d9694c190594f7066), installing prelude with ST binaries instead of downloading the whole of FSL. Now, building the image takes approximately 9 minutes and takes 3.85 GB. When the image is runned, it then takes around 14 GB.

AntoineGuenette:
> I doubt its the OS thats the problem here; even modern Ubuntu's Docker image only takes up [~150MB of storage](https://hub.docker.com/_/alpine) (scroll down for the comparison). Likewise, SCT and ST are not that large; the former is 2GB in an Apptainer (which has a comparable compression regime as Docker), and I can't imagine ST is any worse.
> 
> That leaves more than 20GB laid solely on FSL. As @jcohenadad suggested, if there is a way to trim it down to include only its needed parts, that would be ideal. Alternatively, you could have two different docker images; ST-standalone (for us command-line users), and a "full" install w/ FSLeyes for use with the GUI.
> 
> In theory, we could also make a "wrapper" for the former Docker image, which would allow the user to use it with an existing FSLeyes install (saving us the hassle of shipping it ourselves). Thats just my intuition though; I haven't had the chance to look at FSLeyes plugin handling yet to confirm how feasible this would be.

Thanks a lot for the suggestions! I'll look into that!

SomeoneInParticular:
> When the image is runned, it then takes around 14 GB.

Any reason that might be? What is it doing post-install to more than double its filesize? Just worries me, as (knowing containerization) it might try to do that every time, potentially wasting a lot of the users time (or, worse, storage)

po09i:
> Alternatively, you could have two different docker images; ST-standalone (for us command-line users), and a "full" install w/ FSLeyes for use with the GUI.

Good point! For our use case, we do not need to package the plugin within the docker image. @AntoineGuenette you can look into `make install PLUGIN=false`.

AntoineGuenette:
> Any reason that might be? What is it doing post-install to more than double its filesize? Just worries me, as (knowing containerization) it might try to do that every time, potentially wasting a lot of the users time (or, worse, storage)

This jump in size happens when I run an interactive Bash shell inside the new container created from the image, with the following command :  `docker run -it --rm st_test /bin/bash `. I need to make further testing to investigate this!

AntoineGuenette:
> Here are the details on the space needed after each step :
> 
> 1. After building the image from the Dockerfile
>    
>    * Compressed size of the image : 3.32 GB
>      
>      * Includes only the image itself
>      * It corresponds to the size shown by Docker Desktop and the command `docker image ls`
>    * Uncompressed size of the image: 8.352 GB
>      
>      * Includes all extracted files from all layers, as actually stored on disk by Docker
>      * It corresponds to the size shown with the command `docker system df`
> 2. After running a container from the image
>    
>    * Compressed size of the image : 11.67 GB
>    * Uncompressed size of the image : 16.62 GB
> 3. After exporting the filesystem of the container in a tarball (.tar file)
>    
>    * Size of the tarball : 7.99 GB
>    * It's this file that will need to be exported to the MRI console (i.e. with a USB key)

@jcohenadad @po09i @4rnaudB After investigating the evolution of the size taken by the image after each step (see citation, taken fro the modified PR description), we can indeed see a jump in space taken on the disk when running the container (from 8.4 GB to 16.6 GB). This seems to be a normal phenomenon, since when a container is created, Docker generates additional metadata, cache files and runtime layers to support the container's operations.

PR #611, opened 2025-06-10
Output an updated constraint file after performing a volume shim

po09i:
## Description
This PR adds the functionality to output an updated constraint file after a volume shim. This allows to use this config file in a subsequent acquisition that is based on the already shimmed fmap.

Todo: 
- Verify if this works with Siemens (DAC units might create issues)


shimming-toolbox/shimmingtoolbox/cli/b0shim.py
```diff
@@ -1624,6 +1630,42 @@ def coefs_to_dict(coefs_coil, scanner_coil_order, manufacturer):
...
+        coefs_scanner[f"{order}"] = coefs_coil[:, start_channel_scanner:end_channel_scanner_order][0]
+        start_channel_scanner = end_channel_scanner_order
+
+    coefs_used = {}
+    for order in coil.coefs_used.keys():
+        if coefs_scanner.get(order) is None:
+            coefs_used[order] = coil.coefs_used[order]
+        else:
+            coefs_used[order] = np.nan_to_num(np.array(coil.coefs_used[order], dtype=float), nan=0) + coefs_scanner[
+                order]
```
4rnaudB: Couldn't this be in the same for loop as the one above

shimming-toolbox/shimmingtoolbox/cli/b0shim.py
```diff
@@ -1624,6 +1630,42 @@ def coefs_to_dict(coefs_coil, scanner_coil_order, manufacturer):
...
+
+    coef_channel_minmax = {}
+    for order, coefs in coil.coef_channel_minmax.items():
+        coef_channel_minmax[order] = []
+        coefs_used_order = np.nan_to_num(np.array(coil.coefs_used[order], dtype=float), nan=0)
+        for i_coef, coefs_channel in enumerate(coefs):
+            coef_channel_minmax[order].append([float(coefs_channel[0]) + coefs_used_order[i_coef],
+                                               float(coefs_channel[1]) + coefs_used_order[i_coef]])
```
4rnaudB: Same here, you can access with coil.coef_channel_minmax[order]

4rnaudB (changes requested):
Tested it and it works, but I would change the code.

shimming-toolbox/shimmingtoolbox/cli/b0shim.py
po09i: Fixed

shimming-toolbox/shimmingtoolbox/cli/b0shim.py
po09i: Fixed

4rnaudB (approved):

PR #612, opened 2025-06-10
Add division and standard deviation calculation to st_maths

po09i:
## Description
Add division and standard deviation calculation to st_maths

### DIFF
--- a/shimming-toolbox/shimmingtoolbox/cli/maths.py
+++ b/shimming-toolbox/shimmingtoolbox/cli/maths.py
@@ -12,7 +12,7 @@ import os
 from shimmingtoolbox.utils import add_suffix, set_all_loggers, splitext, save_nii_json, create_output_dir
 
 CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
-AXES = ['0', '1', '2', '3', '4']
+AXES = ['-1', '0', '1', '2', '3', '4', '5', '6']
 DEFAULT_PATH = os.path.abspath(os.curdir)
 
 
@@ -27,7 +27,7 @@ def maths_cli():
               help="Input filename, supported extensions: .nii, .nii.gz")
 @click.option('-o', '--output', 'fname_output', type=click.Path(), default=None,
               help="Output filename, supported extensions: .nii, .nii.gz. [default: ./input_mean.nii.gz]")
-@click.option('--axis', type=click.Choice(AXES), default=AXES[3], show_default=True,
+@click.option('--axis', type=click.Choice(AXES), default=AXES[0], show_default=True,
               help="Axis of the array to calculate the average")
 @click.option('-v', '--verbose', type=click.Choice(['info', 'debug']), default='info', help="Be more verbose")
 def mean(fname_input, fname_output, axis, verbose):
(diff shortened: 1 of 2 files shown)


shimming-toolbox/shimmingtoolbox/cli/maths.py
```diff
@@ -61,6 +61,78 @@ def mean(fname_input, fname_output, axis, verbose):
...
+    # Create nibabel output
+    nii_output = nib.Nifti1Image(std_data, nii_input.affine, header=header)
+
+    # Save image
+    if fname_output is None:
+        _, filename = os.path.split(fname_input)
+        fname_output = add_suffix(os.path.join(os.curdir, filename), '_std')
+    nib.save(nii_output, fname_output)
+    a=1
+
```
AntoineGuenette, line 104: Does this line have an utility?

4rnaudB: Probably a filler line he added when first creating the function and forgot to remove. You can use `pass` in the future so you don't forget

shimming-toolbox/shimmingtoolbox/cli/maths.py
```diff
@@ -61,6 +61,78 @@ def mean(fname_input, fname_output, axis, verbose):
     nib.save(nii_output, fname_output)
 
 
+@maths_cli.command(context_settings=CONTEXT_SETTINGS)
+@click.option('-i', '--input', 'fname_input', type=click.Path(exists=True), required=True,
+              help="Input filename, supported extensions: .nii, .nii.gz")
+@click.option('-o', '--output', 'fname_output', type=click.Path(), default=None,
+              help="Output filename, supported extensions: .nii, .nii.gz. [default: ./input_std.nii.gz]")
+@click.option('--axis', type=click.Choice(AXES), default=AXES[3], show_default=True,
+              help="Axis of the array to calculate the average")
```
4rnaudB: I don't see the point of limiting the axis. I would simply require an integer between -1 and nb_dims. -1 should be the default value.

shimming-toolbox/shimmingtoolbox/cli/maths.py
```diff
@@ -61,6 +61,78 @@ def mean(fname_input, fname_output, axis, verbose):
...
+
+    # Set logger level
+    set_all_loggers(verbose)
+
+    # Load input
+    nii_input = nib.load(fname_input)
+    nii_input2 = nib.load(fname_input2)
+
+    # Compute division
+    div_data = nii_input.get_fdata() / nii_input2.get_fdata()
```
4rnaudB: need to make sure dim A >= dim B

shimming-toolbox/shimmingtoolbox/cli/maths.py
```diff
@@ -61,6 +61,78 @@ def mean(fname_input, fname_output, axis, verbose):
...
+
+    # Create nibabel output
+    nii_output = nib.Nifti1Image(std_data, nii_input.affine, header=header)
+
+    # Save image
+    if fname_output is None:
+        _, filename = os.path.split(fname_input)
+        fname_output = add_suffix(os.path.join(os.curdir, filename), '_std')
+    nib.save(nii_output, fname_output)
+    a=1
```
po09i, line 102:
```suggestion
```

shimming-toolbox/shimmingtoolbox/cli/maths.py
po09i: The reason is that it is easier for a user to see what the input data can be at a glance using `click.Choice` compared to `click.IntRange`. However, I do agree that having -1 as a default option is better, so I added the option as well.

shimming-toolbox/shimmingtoolbox/cli/maths.py
```diff
@@ -99,29 +101,32 @@ def std(fname_input, fname_output, axis, verbose):
...
         fname_output = add_suffix(os.path.join(os.curdir, filename), '_std')
     nib.save(nii_output, fname_output)
-    a=1
 
 
 @maths_cli.command(context_settings=CONTEXT_SETTINGS)
 @click.option('-i', '--input', 'fname_input', type=click.Path(exists=True), required=True,
               help="Input filename, supported extensions: .nii, .nii.gz")
-@click.option('-i2', '--input2', 'fname_input2', type=click.Path(exists=True), required=True,
+@click.option('-d', '--denominator', 'fname_denom', type=click.Path(exists=True), required=True,
```
AntoineGuenette, line 109: just to be coherent with the nomenclature, maybe change -i --input to -n --numerator?

4rnaudB (approved):

PR #618, opened 2025-07-15
Add unwrapping and filtering in 2D when field mapping

po09i:
## Description
Add unwrapping and filtering in 2D when field mapping.

### DIFF
--- a/shimming-toolbox/shimmingtoolbox/cli/prepare_fieldmap.py
+++ b/shimming-toolbox/shimmingtoolbox/cli/prepare_fieldmap.py
@@ -41,9 +41,11 @@ CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
               help="Filename of the mask calculated by the unwrapper")
 @click.option('--gaussian-filter', 'gaussian_filter', type=bool, show_default=True, help="Gaussian filter for B0 map")
 @click.option('--sigma', type=float, default=1, help="Standard deviation of gaussian filter. Used for: gaussian_filter")
+@click.option('--2d', 'process_in_2d', type=bool, show_default=True, default=0,
+              help="Unwrap and filter slice by slice. Defaults to False, which unwraps the whole 3D volume at once.")
 @click.option('-v', '--verbose', type=click.Choice(['info', 'debug']), default='info', help="Be more verbose")
 def prepare_fieldmap_cli(phase, fname_mag, unwrapper, fname_output, autoscale, fname_mask, threshold, fname_save_mask,
-                         gaussian_filter, sigma, verbose):
+                         gaussian_filter, sigma, process_in_2d, verbose):
     """Creates fieldmap (in Hz) from phase images.
 
     This function accommodates multiple echoes (2 or more) and phase difference. It also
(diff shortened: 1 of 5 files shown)


shimming-toolbox/shimmingtoolbox/cli/prepare_fieldmap.py
```diff
@@ -63,12 +65,13 @@ def prepare_fieldmap_cli(phase, fname_mag, unwrapper, fname_output, autoscale, f
     set_all_loggers(verbose)
 
     prepare_fieldmap_uncli(phase, fname_mag, unwrapper, fname_output, autoscale, fname_mask, threshold, fname_save_mask,
-                           gaussian_filter, sigma)
+                           gaussian_filter, sigma, process_in_2d)
 
 
 def prepare_fieldmap_uncli(phase, fname_mag, unwrapper='prelude',
```
4rnaudB, line 71: Just realized this exists, but it is confsing to have a prepare_fieldmap.py file in shimmingtoolbox/cli and in shimmingtoolbox/. Also it is weird to have a prepare_fieldmap_uncli() function when we have a prepare_filedmap() function.

shimming-toolbox/shimmingtoolbox/prepare_fieldmap.py
```diff
@@ -153,11 +167,24 @@ def prepare_fieldmap(list_nii_phase, echo_times, mag, unwrapper='prelude', nii_m
             for it in range(fieldmap_hz.shape[-1]):
                 # Fill values
                 filled = fill(fieldmap_hz[..., it], mask[..., it] == False)
-                fieldmap_hz[..., it] = gaussian(filled, sigma, mode='nearest') * mask[..., it]
+
+                if process_in_2d:
+                    for i_slice in range(fieldmap_hz.shape[2]):
+                        # Apply gaussian filter slice by slice
+                        fieldmap_hz[..., i_slice, it] = gaussian(filled[..., i_slice], sigma, mode='nearest') * \
+                                                        mask[..., i_slice, it]
```
4rnaudB: might be worht trying the `channel_axis` paramter of `skimage.filters.gaussian`. It is meant to filter red blue and green seperatly in RGB images but I think it would work for different slices and would avoid a probably long for loop.

shimming-toolbox/shimmingtoolbox/unwrap/skimage_unwrap.py
```diff
@@ -69,6 +72,22 @@ def skimage_unwrap(nii_wrapped_phase, mag=None, mask=None, threshold=None, fname
         # Normal 3D case: Unwrap the phase
         ma_unwrapped_phase = unwrap_phase(ma_wrapped_phase, rng=0)
 
+    if is_unwrapping_in_2d:
+        ma_unwrapped_phase_2d = ma.array(wrapped_phase, mask=~unwrap_mask.astype(bool))
+        for i_slice in range(wrapped_phase.shape[2]):
+            # Unwrap the phase slice by slice
+            ma_unwrapped_phase_2d[:, :, i_slice] = unwrap_phase(ma_unwrapped_phase[:, :, i_slice], rng=0)
```
4rnaudB, line 79: no option for prelude and skimage to unwrap in 2D?

shimming-toolbox/shimmingtoolbox/cli/prepare_fieldmap.py
po09i: > Just realized this exists, but it is confsing to have a prepare_fieldmap.py file in shimmingtoolbox/cli and in shimmingtoolbox/.

Well there is a CLI file and an API file

> Also it is weird to have a prepare_fieldmap_uncli() function when we have a prepare_filedmap() function.

The initial implementation only had the normal cli function, but I had to split it so that the create_coil_profile CLI could use most of the CLI functionalities and not just what the API does. This probably means that some of the CLI code should be moved into the API...

I think it's outside the scope of the PR though...

shimming-toolbox/shimmingtoolbox/unwrap/skimage_unwrap.py
po09i: Prelude does have the option, skimage does not. This is why I added it myself in skimage but not in Prelude.

shimming-toolbox/shimmingtoolbox/prepare_fieldmap.py
po09i: Nice find, fixed!

shimming-toolbox/shimmingtoolbox/cli/prepare_fieldmap.py
4rnaudB: Agreed, but might need to be rethink

4rnaudB (approved):

PR #643, opened 2026-02-20
Add env executables to the PATH when necessary

po09i:
## Description
ST runs into an error when running from git bash on Windows if sub commands are launched (dcm2niix, spec2nii). 

Adding the path of these subcommands fixes the problem.

To make sure this does not happen in the future. I tested that dcm2niix and spec2nii are correctly installed in `st_check_dependencies`

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -66,6 +66,7 @@ jobs:
       - name: Set path for Shimming Toolbox
         run: |
           echo "$HOME/shimming-toolbox/bin" >> $GITHUB_PATH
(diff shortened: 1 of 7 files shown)


Annaelle8 (approved):

po09i:
I added a few more things since the last approval. Just make sure things are still ok

Annaelle8 (approved):
All pytest tests passed (macOS). I did not test the Windows configuration.

po09i:
Coverall is having issues and making the PR fail. Waiting for it to get resolved before merging

Issue #600, opened 2025-04-08
Improve visualization of the shimmed vs unshimmed figure


po09i:
### Description of Problem / Inefficiency:
We can improve the visualization of the fig_shimmed_vs_unshimmed.png figure. In particular:

- [ ] When there are many slices, the individual slices become small (remove non shimmed slices)
- [ ] When the mask is small, we should "zoom in/crop" on the important part of the figure
- [ ] Add units (technically, the units can be anything, depends on the inputs so it's hard to add something)

AntoineGuenette:
Also, when shimming with the scanner's coils for order 0 (f0) and orders 1-3, there should be a multi-scale y-axis.

PR #644, opened 2026-04-08
Zoom onto the region being shimmed when displaying figures

po09i:
## Description
Zoom onto the region being shimmed when displaying figures. I changed all plots (signal recovery, real-time, etc to have this behaviour)

Example:
Screenshot 2026-04-08 at 6 15 50 PM

Screenshot 2026-04-08 at 6 08 03 PM

Related to #600


Annaelle8 (commented):
All the test passed.

During tests performed with the provided spine dataset (data_dynamic_shimming): 
```
st_b0shim dynamic \
  --fmap fmap/sub-spine_phase2.nii.gz \
  --target anat/sub-spine_unshimmed_e1.nii.gz \
  --scanner-coil-order 1 \
  -v debug
```

The figures appear cropped below and above the area of ​​interest. Is this the intended zoom effect?

Before the change:
Image

After the change:
Image

po09i:
The idea is to focus on the coloured ROI (i.e. the field map values). So yes it is the intended behaviour. In that case, it does not do much..

Annaelle8 (approved):

PR #646, opened 2026-06-05
Allow to disable specific shim channels and to specify their shim values from a text file

po09i:
## Description
This PR adds the ability to turn OFF specific shim channels. Those channels can either be set to 0 (default) or a text file can be provided (using slicewise-coil and delta format) to specify what the shim values should be for each slice.

Todo in another PR:
- Implement for real-time shimming #649


Annaelle8 (approved):

mathisors (approved):
- Ran the pytest, no issues there. 
- Also tested normal real-time shimming functionnality by running example/demo_realtime_shimming.sh, also looked good.
- Read the code and played around in the terminal, everything looked good.
My only comment might be that the off_channel_value variable name is a bit confusing, my initial instinct was that when a channel is an off_channel it's value is zero, so its a bit confusing the an off_channel can have another value. Since this might be more of a personnal preference will approve the pr.

PR #650, opened 2026-06-09
Allow to remove channels (columns) from a text file

po09i:
## Description
In `st_b0shim convert-shim-coefs-format`, allow to delete/remove specific channels. This function can now be used to specify a text file in line with the `--off-channel-values` format in `st_b0shim dynamic` (only specify the channels that are off).

## Linked issues
#646


Annaelle8 (approved):
Every tests passed

PR #651, opened 2026-06-16
Add a constraint file section to the website

po09i:
## Description
Add information about constraint files on the website.

## How to look at the new doc
To look at the [new doc](https://shimming-toolbox-py--651.org.readthedocs.build/en/651/miscellaneous/constraint_file.html).
It can be accessed through the question 6 in the FAQ.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -41,6 +41,7 @@ extensions = ["sphinx.ext.githubpages",
 
 autosectionlabel_prefix_document = True
 autosectionlabel_maxdepth = 2
+myst_heading_anchors = 3
 
 # Autodoc configuration
 autodoc_mock_imports = [
--- a/docs/source/miscellaneous/constraint_file.md
+++ b/docs/source/miscellaneous/constraint_file.md
@@ -3,10 +3,23 @@ orphan: true
 ---
 
 # All about constraint files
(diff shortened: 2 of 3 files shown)


po09i:
Waiting on #652 for tests to pass

mathisors:
Comment on the new doc for the constraint files:
Overall I think it pretty clear and give the right information. There is just a few things I thought could be improved:
- For coef_sum_max, it could be useful to specify when a value other than null should be provided. I feel it may be confusing for a new user to determine if null can safely be used in their cases.
- Regarding the "Do I need to use a scanner constraint file?" section, I don't think it fully answers the question of the heading. After reading it, it is still unclear to me if using a scanner constraint file is obligatory or optional. However, the section is clear on how to add a new scanner constraint file and how to use an already integrated one.
- It could be helpful to include a list of the scanners for which constraint files are already integrated, becasue I don't think its clear if that includes every scanner from Siemens, Philips, and GE, or only specific models?
- Since the page is pretty long, you could consider adding a note or reference near the beginning that mentions that it also covers coil constraints. It was not obvious to me that the page discusses all types of constraint files, not only scanner constraint files.

po09i:
Good points @mathisors ! I updated the doc following your input, let me know if I made things better and if there are still problems!

Annaelle8 (approved):
Everything seems clear to understand

mathisors (approved):
Yes it's looks better!

PR #652, opened 2026-06-16
Fix pytest.mark.parametrize string-based parameter list now handling single element tuples

po09i:
## Description
pytest.mark.parametrize did not correctly handle single parameter tuples and would not unpack the tuple. We needed to unpack it manually: [0]. The bug has been fix in their latest version (9.1.0) and it now correctly unpacks it. The PR removes those unnecessary [0].

## Linked issues
https://github.com/pytest-dev/pytest/issues/719

### DIFF
--- a/shimming-toolbox/requirements_st.txt
+++ b/shimming-toolbox/requirements_st.txt
@@ -12,7 +12,7 @@ phantominator>=0.6.4
 pillow>=9.0.0
 psutil>=5.8.0
 pydicom>=2.4.4
-pytest>=6.2.5
+pytest>=9.1.0
 pytest-cov>=2.5.1
 quadprog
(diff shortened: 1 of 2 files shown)


Annaelle8 (commented):
The test done with pytest failed because of error in dimension

Image

po09i:
This is most likely because you are using a previous version of pytest. I updated the minimum version required. To test it, "pip install" shimming toolbox again in the correct environment.

Annaelle8 (approved):
All the tests passed with this new version of pytest.

mathisors (approved):
All test are passing now, at: =========================================== 494 passed, 7 warnings in 262.35s (0:04:22) ============================================

PR #653, opened 2026-06-18
Make sure an error appears if the constraint file does not exist

po09i:
## Description
Make sure an error appears if the scanner constraint file does not exist when using st_b0shim


Annaelle8 (approved):
All tests passed using pytest

PR #654, opened 2026-06-18
Simplify BET mask output file

po09i:
## Description
I changed the behaviour of the `-o` argument in `st_mask bet`. It was causing unnecessary confusion. Instead of requiring no extension and a mask be created with that path + _mask.nii.gz (behaviour of BET). Now the path entered will be the name of the output mask file (in line with all other st_mask CLI). Under the hood, we create the mask with a standard name in a tmp directory and rename it to the user's input. We need a `.nii.gz` to stay consistent with BET's output.


Annaelle8 (commented):
Using pytest: test_mask_tab.py failled. I think the mask file is not created ?

Image

po09i:
The file should be created (which is why the tests pass online). Are you on the right version of the plugin? (pip install in the right environment of fsleyes-plugin-shimming-toolbox)

Annaelle8 (approved):
You're right I did the pip install for the first test but not for the second one. Everything works !

PR #660, opened 2026-06-26
Remove hard requirement of SoftwareVersion tag the field map JSON sidecar

po09i:
## Description
When using --slices with anything other than auto, there was a hard requirement on the BIDS tag `SoftwareVersion` to be present. That tag is only used for logging and should not be a hard requirement. This removes the error being thrown if that tag is not present.


Annaelle8 (approved):
All tests passed.

PR #663, opened 2026-07-07
Sort non acquisition related DICOM files

po09i:
## Description
Some DICOM files don't have acquisition numbers or series description (e.g.: DICOMDIR). These files would raise error if they were included in the input data. This PR finds these files and copies them into the root of the output directory.

Also add pydicom to the default list of excluded loggers


mathisors (approved):
All test have passed

PR #666, opened 2026-07-14
Support 4d field maps with singleton in the last dimension in st_b0shim dynamic

po09i:
## Description
It is possible to save 4D data in a NIfTI file with a singleton in the last dimension (e.g.: 64x64x20x1). ST flags it as 4D data which is unsupported by st_b0shim dynamic. However, this is effectively 3D data so we should support it. This PR fixes that.

## Linked issues
https://github.com/shimming-toolbox/shimming-toolbox/discussions/664#discussioncomment-17635493


Annaelle8 (approved):
All tests passed

PR #672, opened 2026-07-29
Add ability to optimize in the gradient coordinate system

### LINKED ISSUE #670: Allow to optimize in the Gradient CS
### Feature Description:
We currently optimize using the world coordinates and we can control if we want to turn off specific shim coils. When we shim, we sometimes want to refer to the gradient coordinate system (RO, PE, SL) so that we can only shim in the slice direction for example. This is currently not possible for non pure orientations. For a slightly rotated axial acquisition, most of the slice gradient is the z gradient but also uses some combination of the x and y gradient.

### Proposal
We can't just change the CS of the output coefficients because we won't be able to use the feature to turn off specific axes in the gradient CS.

- Add an option (--opt-cs: [shim-cs, gradient-cs])
- Modify all inputs (mask, coil profiles to be in the new coordinate system)
- The output could be in either CS. For now, we only need it in the shim CS so convert it if necessary
- What happens to the second order coefficients, order 0?

po09i:
## Description
This PR adds the ability to optimize in the gradient coordinate system. In conjunction with `--off-channels` this allow to turn off the frequency encoding direction for example whereas before we could only turn off specific world coordinate axes (x, y, z)

Todo:
- [x] Verify bounds are correct

## Linked issues
Fixes #670

### DIFF
--- a/shimming-toolbox/shimmingtoolbox/coils/scanner_shim_settings.py
+++ b/shimming-toolbox/shimmingtoolbox/coils/scanner_shim_settings.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 """
-This file includes functions a classes useful to extract scanner's shim settings
+This file includes functions and classes useful to extract scanner's shim settings
 """
 import copy
 import logging
--- a/shimming-toolbox/shimmingtoolbox/coils/scanner_shim_settings.py
+++ b/shimming-toolbox/shimmingtoolbox/coils/scanner_shim_settings.py
@@ -1,0 +1,132 @@
+# -*- coding: utf-8 -*-
+"""
+This file includes functions a classes useful to extract scanner's shim settings
+"""
+import copy
+import logging
+import numpy as np
+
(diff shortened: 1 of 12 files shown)


shimming-toolbox/shimmingtoolbox/coils/scanner_shim_settings.py
```diff
@@ -0,0 +1,132 @@
+# -*- coding: utf-8 -*-
+"""
+This file includes functions a classes useful to extract scanner's shim settings
```
Annaelle8:
```suggestion
This file includes functions and classes useful to extract scanner's shim settings
```

Annaelle8 (approved):
The bounds seem correct for me. All tests passed.

shimming-toolbox/shimmingtoolbox/coils/scanner_shim_settings.py
po09i: Fixed in [464e237](https://github.com/shimming-toolbox/shimming-toolbox/pull/672/commits/464e237675521702bbf2f68b29d074aa7cbf5f8c)

rhiannonbutler (approved):
All tests passed for me!

PR #673, opened 2026-07-30
Don't display fig_currents if it's volume shim (only 1 set of currents)

po09i:
## Description
Skip generating the fig_currents figure if we do volume shim as there is only one to show.


sriosq:
Checked. Approved.

sriosq (approved):
Good

PR #676, opened 2026-08-07
Use threading backend with joblib

po09i:
## Description
I have noticed that when running manual tests, execution would hang after the test seemed to have executed. Then a few minutes later, I would get an error about leaked folders/semaphores. The issue is related to joblib's loky backend. This PR changes the backend to use "threading" which seems to be preferred when using numpy arrays. Loky seems to be better for cpu tasks though, so we should monitor how speed is impacted.

## Traceback
```
.../miniconda3/envs/shim_venv/lib/python3.11/site-packages/joblib/externals/loky/backend/resource_tracker.py:359: UserWarning: resource_tracker: There appear to be 3 leaked folder objects to clean up at shutdown
```


rhiannonbutler (approved):
looks good to me! and tests passed.

Annaelle8 (approved):
All tests passed.

PR #680, opened 2026-09-04
Add missing default option from gaussian-filter option

mathisors:
## Checklist

- [x] I've applied the relevant labels to this PR. (You need to choose: the scope of the PR, 1 or more (pale brown), the main reason for the PR, only 1 (dark purple) and if applicable, the CLI affected (dark cyan))

## Description

Adding the missing default value from the gaussian-filter option of the st_prepare_fieldmap command, so it shows up in the docs instead of Sentinel.UNSET.

### DIFF
--- a/shimming-toolbox/shimmingtoolbox/cli/prepare_fieldmap.py
+++ b/shimming-toolbox/shimmingtoolbox/cli/prepare_fieldmap.py
@@ -39,7 +39,7 @@ CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
                    "than the threshold are set to 0.")
 @click.option('--savemask', 'fname_save_mask', type=click.Path(),
               help="Filename of the mask calculated by the unwrapper")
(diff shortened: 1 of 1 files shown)


po09i (changes requested):
Good fix, can you also fix this issue in the `st_create_coil_profiles` documentation. It is the exact same problem where `sentinel.UNSET` is visible.

mathisors:
Question @po09i: For the sigma option there is a defined default value of  '1', but it is not shown using the show_default. I feel like that would be helpful information to show in the doc. What do you think/Should I add it to this PR?

po09i:
> Question @po09i: For the sigma option there is a defined default value of '1', but it is not shown using the show_default. I feel like that would be helpful information to show in the doc. What do you think/Should I add it to this PR?

Sure! We can add the `show_default=True` to that option.

po09i (approved):

PR #681, opened 2026-09-08
Allow changing files from gradient-cs to shim-cs and vice-versa in convert_shim_coefs_format

po09i:
## Description
Allow to change from gradient-cs to shim-cs and vice-versa in `st_b0shim convert_shim_coefs_format`.

This is relevant since the optimizer only outputs in the shim-cs so it allows to change that. It also allows to  manually create a text file in a coordinate system and convert it.


Annaelle8 (approved):
All tests passed.

PR #682, opened 2026-09-08
Return coefficients instead of only the index in max-intensity optimization

po09i:
## Description
This PR allows to input the shim coefficients used for the max-intensity optimization and return the shimmed text file instead of only outputting the index of the max volume.

Indexes are now output with 0 index instead of 1 index.


Annaelle8 (approved):
All tests passed !

